Issue #3, opened 2018-11-24
draw-tools does't work on touch


johnd0e:
There are 2 ways to fix it:

1. As suggested in https://github.com/iitc-project/ingress-intel-total-conversion/issues/1142, leaflet.draw should be updated.
And related fix: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1244 (fix earth radius to same value used in leaflet and S2).
More recents work on both issues is seen here: https://github.com/McBen/ingress-intel-total-conversion

2. There is alternative fix: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1261 (Touch events default passive to false).
You can test it here: https://play.google.com/store/apps/details?id=com.cradle.iitc_mobile

johnd0e:
I see this issue is fixed now, thanks!

But icons at draw panel seem blurry.

johnd0e:
>Leaflet 1.3.0, leaflet.draw 1.0, ...

Related fix: https://github.com/McBen/ingress-intel-total-conversion/commit/35ed971c75030ec0e40344eb003fa34bac9d2f78 (included in #10)

johnd0e:
Perhaps https://github.com/McBen/ingress-intel-total-conversion/commit/1158ade0aed0ef2dd36114a7014e5fa9e75defe6 + https://github.com/McBen/ingress-intel-total-conversion/commit/a3d34cecc01b505051921c9974faf971aaf1f9f6 should be applied too.

PR #8, opened 2018-12-04
fix some plugins to comply with updated leaflet (otherwise it become really hard to click at portal)

### LINKED ISSUE #7: Portal Names plugin problem (after recent update)
After I've enabled the plugin I am unable to click at portal.

Apk from market does not have this issue, so I suppose it is caused by the very recent update, probably related to 68cf49f90b75673380167c59b24dd3cfd2aabee3.

johnd0e:
belongs to #3, fixes #7


johnd0e:
Also I found occurrences of [outdated](https://github.com/Leaflet/Leaflet/pull/2838) property `clickable` in these files:

code\artifact.js (in comment)
code\ornaments.js (in comment)

external\Control.MiniMap.js
external\leaflet.draw-src.js
external\oms.min.js

plugins\layer-count.user.js
plugins\layer-farms-find.user.js
plugins\zaprange.user.js

Perhaps, all these also need to be fixed.

PR #14, opened 2018-12-07
layer-count fix: was accessing a private class property

### LINKED ISSUE #11: layer-count doesn't work after update
I suppose it fails after leaflet update: 68cf49f90b75673380167c59b24dd3cfd2aabee3 (though unchecked).

johnd0e:
This fixes #11.
(from https://github.com/McBen/ingress-intel-total-conversion/commit/8668bc0f335aa377bd5414c2248e738f02c4385c)

### DIFF
--- a/plugins/layer-count.user.js
+++ b/plugins/layer-count.user.js
@@ -33,7 +33,6 @@ plugin.layerCount.onBtnClick = function(ev) {
 		map.off("click", plugin.layerCount.calculate);
 		btn.classList.remove("active");
 	} else {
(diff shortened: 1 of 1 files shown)


johnd0e:
I think the similar fix should also be applied to line 116, just for consistency (though there is no error there currently).

modos189:
> I think the similar fix should also be applied to line 116, just for consistency (though there is no error there currently).

Yes

johnd0e:
> Yes

Just fixed that too

PR #29, opened 2018-12-18
Mobileinfo fix

johnd0e:
Here are two fixes for mobile status line.


modos189:
> Or may be we should tell user that info is requested

If the download takes longer (for example) 500ms

Issue #30, opened 2018-12-18
Portal taphold does not work in canvas mode


johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/df0564f5b97c015446cb7d9bcd1b0b61162e4cfb/code/smartphone.js#L160-L173

In canvas mode `_container` property does not exist.

johnd0e:
I am not sure how exactly it should be fixed, but I suppose that portal marker should support [contextmenu](https://leafletjs.com/reference-1.3.4.html#circlemarker-contextmenu) event.

johnd0e:
This works (replace all the code):
```
  window.addHook('portalAdded', function(data) {
    data.portal.on('contextmenu', function(e) {
      window.renderPortalDetails(e.target.options.guid);
      window.show('info');
    });
  });
```

I am not good in js, but as we do not need to make closure anymore, it could be more efficient to replace anonymous function with defined:

```
  // add event to portals that allows long press to switch to sidebar
  function portal_taphold_handler(e) {
    window.renderPortalDetails(e.target.options.guid);
    window.show('info');
  };

  window.addHook('portalAdded', function(data) {
    data.portal.on('contextmenu', portal_taphold_handler);
  });
```
_(If it is really more efficient then we could make similar improvements in several other parts of code)_

PR #33, opened 2018-12-19
Portal handlers refactoring

johnd0e:
Here is the code to illustrate possible #30 solution, and beyond.
I am not sure enough about all ideas, so please review the code.

### DIFF
--- a/code/boot.js
+++ b/code/boot.js
@@ -437,38 +437,22 @@ window.setupPlayerStat = function() {
   );
 }
 
-window.showSidebar = function(state) {
-  var sidebar = $('#scrollwrapper');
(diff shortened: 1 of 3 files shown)


johnd0e:
As for using "contextmenu" event both on mobile and on desktop I have some doubts, cause it doesn't look like usual behavior of IITC controls.

BTW, it may be good idea to use real context menu's in IITC, and there are ready-to-use leaflet-plugins.
Context menu could be more convenient then current "toolbox" in portal info view.

modos189:
I apologize, I won't be available for the next 3 days. Travel =)

modos189 (approved):
Looks great

PR #35, opened 2018-12-23
show-linked-portals improvements

johnd0e:


johnd0e:
Concerning https://github.com/IITC-CE/ingress-intel-total-conversion/pull/35/commits/6c0f1052e69d08424ce0114f471a4d933e67ea1a: I would prefer to unconditionally auto-request portals' details (like here
https://github.com/johnd0e/ingress-intel-total-conversion/commit/128247e2c66494559046fb7932473b2a78b2e13f), but I am afraid it could be against IITC policy...

modos189:
> I would prefer to unconditionally auto-request portals' details 

It doesn't have to be automatic: there can be a lot of portals for auto-request.
There can be a hybrid mode: if the number of links is less than N, portals are automatically loaded. Otherwise, at the user's request.
But I don't think that's really necessary.

PR #40, opened 2018-12-28
Show linked portals improve

johnd0e:
* portal images fill full width
* on taphold: text instead of portal pic
* last viewed portal is highlighted
* on mobile:
  - single tap to show tooltip
  - double tap to go to portal
* on taphold: improved 'show link on map' function 
* picture in tooltip
* tooltip now updates on new data coming
* style tweaks
* minor bug fixed

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebView.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebView.java
@@ -58,6 +58,7 @@ public class IITC_WebView extends WebView {
         mSettings.setCacheMode(WebSettings.LOAD_DEFAULT);
         mSettings.setAppCachePath(getContext().getCacheDir().getAbsolutePath());
         mSettings.setDatabasePath(getContext().getApplicationInfo().dataDir + "/databases/");
+        mSettings.setTextZoom(100); // otherwise zoom may vary depending on system font settings
 
         // enable mixed content (http on https...needed for some map tiles) mode
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
--- a/plugins/linked-portals-show.css
+++ b/plugins/linked-portals-show.css
@@ -27,6 +27,10 @@
 	left: 50%;
 	margin-left:-25px;
 	cursor: default;
+	border-style: solid;
+	border-color: transparent;
+	line-height: 41px;
+	font-size: 12px;
 }
 
 .showLinkedPortalLink1, .showLinkedPortalLink2, .showLinkedPortalLink3, .showLinkedPortalLink4 {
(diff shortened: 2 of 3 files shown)


johnd0e:
There is minor cosmetic issue left: on click/hover over `outOfRage` div - it gets re-rendered with updated data, but we still see old tooltip, which has `Go to portal` instead of real portal title.
Now investigating if it's easy to reopen tooltip with updated content..

This code works:
```
$('.ui-tooltip-content').html(div.attr('title')); // update tooltip with new content
```
But I feel that it is terribly wrong...

modos189 (approved):

johnd0e:
1 more issue left to fix.
After https://github.com/IITC-CE/ingress-intel-total-conversion/commit/59cc3126bd85f5c9e298e94517384b34c8458d65 some styles had changed, so upper portals' boxes are out of bounds of the main portal's picture.

johnd0e:
Is it possible to merge the PR but still leave it opened?

modos189:
No. We can still leave opened

johnd0e:
@modos189, please check last commit (`portal images fill full width`)
Are images look better now?
Or I should revert it?

modos189:
Good, but just need to add a property
```z-index: 1;```
in
```.showLinkedPortalLink.incoming::before, .showLinkedPortalLink.outgoing::before```

To show icons of link direction

johnd0e:
@modos189, fixed.
And what do you think about border-radius and box-shadow?

modos189:
Then I suggest to add
```background: #00000080;```
to be seen against the background of bright photos.

And all prefixes like `-webkit-border-radius` must be specified before the announcement `border-radius`

johnd0e:
> Then I suggest to add `background: #00000080;`

Well, after some experiments I tend to think that these symbols are hard to see and use as they are ambiguous.
I have another proposition: after all, we care only _outgoing_ links (as they are limited). So wouldn't it better to mark outgoing only?

Please check if you like transparent containers (in latest commit).

P.S.
After that I want to test other concepts:
- Portal info (text) instead of images
- Portal info over the images (which have to be semi-transparent in order to keep text readable)

modos189:
> I have another proposition: after all, we care only outgoing links (as they are limited). So wouldn't it better to mark outgoing only?

It would be handy as a setting, but by default it would be misleading

> if you like transparent containers

like

johnd0e:
> It would be handy as a setting, but by default it would be misleading

Why misleading?
Several days ago we have no symbols at all, tomorrow we'll get some symbol over outgoing.
It couldn't be misleading, as there are also clear text labels incoming/outgoing.

modos189:
I thought you were suggesting not displaying incoming links at all. But since it is only about the icon, I agree not to display it for incoming.

johnd0e:
> show-linked-portals: more bold symbol

Actually I'm not sure that symbol is good.
And we also need to play with it's position (in order to not interfere with text).

PR #45, opened 2019-01-03
portal info: better symbol for close button

johnd0e:
╳ `&#9597;`
✕ `&#10005;`
✖ `&#10006;`
❌ `&#10060;`
❎ `&#10062;` (proposed in the PR)

Sources:
https://en.wikipedia.org/wiki/X_mark
http://xahlee.info/comp/unicode_crosses.html

### DIFF
--- a/code/portal_detail_display.js
+++ b/code/portal_detail_display.js
@@ -98,12 +98,12 @@ window.renderPortalDetails = function(guid) {
       
       $('<h3>').attr({class:'title'}).text(title),
 
-      $('<span>').attr({
-        class: 'close',
-        title: 'Close [w]',
-        onclick:'renderPortalDetails(null); if(isSmartphone()) show("map");',
-        accesskey: 'w'
-      }).text('X'),
+      $('<span>', {class: 'close', title: 'Close [w]', accesskey: 'w' })
+        .click(function() {
+          renderPortalDetails(null);
+          if (isSmartphone()) { show('map'); }
+        })
+        .text('❎'), // .html('&#10062')
 
       // help cursor via ".imgpreview img"
       $('<div>')


johnd0e:
todo: change mobile style (no need for borders).

BTW, may be better to place symbol in css.

modos189:
If remove the border, it will be better, but...
&#10062; in github - red symbol on white background, with border
in mobile - white symbol on red background
in browser - white symbol on green background.

I propose to use the material icon "close"

code/portal_detail_display.js
```diff
@@ -98,12 +98,12 @@ window.renderPortalDetails = function(guid) {
...
       $('<h3>').attr({class:'title'}).text(title),
 
-      $('<span>').attr({
-        class: 'close',
-        title: 'Close [w]',
-        onclick:'renderPortalDetails(null); if(isSmartphone()) show("map");',
-        accesskey: 'w'
-      }).text('X'),
+      $('<span>', {class: 'close', title: 'Close [w]', accesskey: 'w' })
+        .click(function() {
```
modos189, line 102: maybe use .on('click'... ?

johnd0e: [click](https://api.jquery.com/click/) method is just shortcut for `on.('click'`, so it has the same action.

modos189: Ok, I mixed up with the old version of jquery

johnd0e:
Icon 'close' is like ╳✕✖, so I am not sure that we should prefer it over simple symbol. Or there are some special reasons?

(Personally I like boxed cross, but cannot find similar icon in material)

modos189:
There is no box cross, but it is easily implemented using CSS.
The advantage of using only Material icons is that the method of adding icons is then standardized. The same code is used for all icons.

McBen:
since this is a pure 'I Like' question it is hard to decide.

`✕ &#10005;` and `Padding:4px;` (+bold ?)

would look like my desktop  (KDE)  :)
grafik


BTW: `click` is now marked deprecated and `on('click',..` should be used for events.

modos189:
> `✕ &#10005;` and `Padding:4px;` (+bold ?)

looks good

Issue #32, opened 2018-12-19
Scripts updateURL/downloadURL should not be empty


johnd0e:
Scripts that are building for iitc.modos189.ru should have updateURL/downloadURL in order to self-update.

johnd0e:
BTW, #99 also matters, as currently every new build run (on the same code base) will be considered newer than previous, as it's date/time stamp be newer.

PR #53, opened 2019-01-13
deploy: fix update urls (sample)

johnd0e:
!**Not to be merged**
This is sample settings to address #32.

`deploy.sh` should be modified accordingly, as well as travis-ci script.
Actually, I think that we shouldn't really commit this `localbuildsettings.py` into repository.
Better if it be created at runtime, by travis-ci script.

### DIFF
--- a/localbuildsettings.py
+++ b/localbuildsettings.py
@@ -1,0 +1,13 @@
+buildSettings = {
+    # release: using to deploy release builds at official iitc-ce site
+    'release': {
+        'resourceUrlBase': 'https://iitc.modos189.ru/build/release/',
+        'distUrlBase': 'https://iitc.modos189.ru/build/release/',
+    },
+
+    # test: using to deploy test builds at official iitc-ce site
+    'test': {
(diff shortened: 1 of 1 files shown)


johnd0e:
P.S.
Offtopic here, by in my opinion `deploy.sh` also shouldn't be in main repo.

modos189:
>> Offtopic here, by in my opinion deploy.sh also shouldn't be in main repo.

I wouldn't mind, but as far as I know Travis can't add files declared in the control panel

johnd0e:
> I wouldn't mind, but as far as I know Travis can't add files declared in the control panel

But do we really need `deploy.sh` as separate file?
Couldn't it be just included in travis-ci script?

modos189:
I put it in a separate file because it's too individual commands for my site. Unlike the `.travis.yml` file, which can be used in forks.

johnd0e:
> because it's too individual commands for my site

Well, choose what is best for you.
E.g. you could put `deploy.sh` itself at your site, and download it in travis script.

PR #54, opened 2019-01-13
escape macros in code in order to keep js-validity 

### LINKED ISSUE #50: @@-substitutions break validity of plugins' sources
Plugin's sources now are not valid js code, because of build-time substitution blocks like `@@METAINFO@@`, `@@PLUGINSTART@@`, `@@PLUGINEND@@`.

It complicates applying such tools like `eslint`.

It would be great if we could rearrange the code to make the substitutions appear commented, e.g.
```
// @@METAINFO@@
// ...
// @@PLUGINSTART@@
// ...
// @@PLUGINEND@@
```

johnd0e:
Fix #50.

Now all the sources are completely valid js,
that could be processed by external tools, such as eslint.

/ to be discussed before merge /

### DIFF
--- a/build.py
+++ b/build.py
@@ -76,11 +76,14 @@ plugin_info.dateTimeVersion = '@@DATETIMEVERSION@@';
 plugin_info.pluginId = '@@PLUGINNAME@@';
 //END PLUGIN AUTHORS NOTE
 
-"""
+
+// PLUGIN START //"""
 
 pluginWrapperStartUseStrict = pluginWrapperStart.replace("{\n", "{\n\"use strict\";\n", 1)
 
-pluginWrapperEnd = """
+pluginWrapperEnd = """// PLUGIN END //
+
+
 setup.info = plugin_info; //add the script info data to the function as a property
 if(!window.bootPlugins) window.bootPlugins = [];
 window.bootPlugins.push(setup);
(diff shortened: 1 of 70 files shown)


johnd0e:
Well, now I do not like these 'garbage' comments in built files..

johnd0e:
Well, now I'm pretty happy with this PR.

@modos189 
What do you think?

modos189:
Why INCLUDESTRING and INCLUDEIMAGE not updated?

johnd0e:
They do not require escaping as typically applied inside of strings, thus already escaped by quotes.

```
  $('<style>').prop('type', 'text/css').html('@@INCLUDESTRING:plugins/distance-to-portal.css@@').appendTo('head');

```

modos189:
Then all is well. When merge? I don't quite understand the new scheme yet.)

johnd0e:
> When merge?

Well, I've already merged it in my 'develop' branch)
But still wants to work more with this setup to make completely sure than no bugs added.

> I don't quite understand the new scheme yet.)

So let's discuss it.
May be there is even better solution.

johnd0e:
I think that we can completely avoid these macros in plugin sources:
* `@@METAINFO@@` is always placed right before `// ==/UserScript==`
* `@@PLUGINSTART@@` is always right after `// ==/UserScript==`
* `@@PLUGINEND@@` is always near EOF.
* `@@DATETIMEVERSION@@` is near `@version` EOL.
* `@@BUILDNAME@@-@@BUILDDATE@@` is not useful anyway.

PR #60, opened 2019-01-15
Add autonavi.com (China) map layers

magicbug2013:
为IITC增加高德地图，目前修正了portal、link、field及 player tracker plugin的坐标，显示正常，但原来的收藏的portal位置将会有偏移，需要手工重新添加一次。另由于苹果机上的iitc-mobile在新版IITC不能选择图层了，插件在启用后将会将把用户的默认地图切换成高德路网。

### DIFF
--- a/plugins/basemap-gaode-by-gmoogway.user.js
+++ b/plugins/basemap-gaode-by-gmoogway.user.js
@@ -1,351 +1,468 @@
 // ==UserScript==
-// @id             iitc-plugin-basemap-gaode-by-gmoogway
+// @id             iitc-plugin-basemap-gaode@GMOogway
 // @name           IITC plugin: Map layers from GAODE by GMOogway
 // @category       Map Tiles
-// @version        0.2.0.@@DATETIMEVERSION@@
-// @description    [@@BUILDNAME@@-@@BUILDDATE@@] Add autonavi.com (China) map layers
+// @version        0.3.0.@@DATETIMEVERSION@@
+// @description    [@@BUILDNAME@@-@@BUILDDATE@@] Add autonavi.com (China) map layers by GMOogway
 @@METAINFO@@
 // ==/UserScript==
 
 @@PLUGINSTART@@
 
-    // PLUGIN START ////////////////////////////////////////////////////////
+// PLUGIN START ////////////////////////////////////////////////////////
+// use own namespace for plugin
+window.plugin.mapTileGaode = function() {};
 
-    // use own namespace for plugin
-    window.plugin.mapTileGaode = function() {};
+/////////// begin WGS84 to GCJ-02 obfuscator /////////
+var PRCoords = window.plugin.mapTileGaode.PRCoords = (function() {
+  // This set of points roughly illustrates the scope of Google's
+  // distortion. It has nothing to do with national borders etc.
+  // Points around Hong Kong/Shenzhen are mapped with a little more precision,
+  // in hope that it will make agents work a bit more smoothly there.
+  //
+  // Edits around these points are welcome.
+  var POINTS = [
+  // start hkmo
+  114.433722, 22.064310, 114.009458, 22.182105, 113.599275, 22.121763, 113.583463, 22.176002, 113.530900, 22.175318, 113.529542, 22.210608, 113.613377, 22.227435, 113.938514, 22.483714, 114.043449, 22.500274, 114.138506, 22.550640, 114.222984, 22.550960, 114.366803, 22.524255,
+  // end hkmo
+  115.254019, 20.235733, 121.456316, 26.504442, 123.417261, 30.355685, 124.289197, 39.761103, 126.880509, 41.774504, 127.887261, 41.370015, 128.214602, 41.965359, 129.698745, 42.452788, 130.766139, 42.668534, 131.282487, 45.037051, 133.142361, 44.842986, 134.882453, 48.370596, 132.235531, 47.785403, 130.980075, 47.804860, 130.659026, 48.968383, 127.860252, 50.043973, 125.284310, 53.667091, 120.619316, 53.100485, 119.403751, 50.105903, 117.070862, 49.690388, 115.586019, 47.995542, 118.599613, 47.927785, 118.260771, 46.707335, 113.534759, 44.735134, 112.093739, 45.001999, 111.431259, 43.489381, 105.206324, 41.809510, 96.485703, 42.778692, 94.167961, 44.991668, 91.130430, 45.192938, 90.694601, 47.754437, 87.356293, 49.232005, 85.375791, 48.263928, 85.876055, 47.109272, 82.935423, 47.285727, 81.929808, 45.506317, 79.919457, 45.108122, 79.841455, 42.178752, 73.334917, 40.076332, 73.241805, 39.062331, 79.031902, 34.206413, 78.738395, 31.578004, 80.715812, 30.453822, 81.821692, 30.585965, 85.501663, 28.208463, 92.096061, 27.754241, 94.699781, 29.357171, 96.079442, 29.429559, 98.910308, 27.140660, 97.404057, 24.494701, 99.400021, 23.168966, 100.697449, 21.475914, 102.976870, 22.616482, 105.476997, 23.244292, 108.565621, 20.907735, 107.730505, 18.193406, 110.669856, 17.754550, ]
+  var HK_LENGTH = 12 // for future use (e.g. with Baidu)
+  var lats = POINTS.filter(function(_, idx) { return idx % 2 == 1;})
+  var lngs = POINTS.filter(function(_, idx) { return idx % 2 == 0;})
+  POINTS = null
+  // not needed anyway...
+  /// *** pnpoly *** ///
+  // Wm. Franklin's 8-line point-in-polygon C program
+  // Copyright (c) 1970-2003, Wm. Randolph Franklin
+  // Copyright (c) 2017, Mingye Wang (js translation)
+  //
+  // Permission is hereby granted, free of charge, to any person obtaining
+  // a copy of this software and associated documentation files (the
+  // "Software"), to deal in the Software without restriction, including
+  // without limitation the rights to use, copy, modify, merge, publish,
+  // distribute, sublicense, and/or sell copies of the Software, and to
+  // permit persons to whom the Software is furnished to do so, subject to
+  // the following conditions:
+  //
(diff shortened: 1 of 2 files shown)


johnd0e:
>  Because iitc-mobile on iPhone can't select the layer in the new version of IITC

@magicbug2013 Shoudn't you open issue? May be it could be fixed.

> please remove the following comments before adding to user script

Is it possible to detect ios programmatically?

> 但原来的收藏的portal位置将会有偏移，需要手工重新添加一次。

Are you talking about Bookmarks?  (Google translated as 'original collection's portal location')

magicbug2013:
@johnd0e 
> Is it possible to detect ios programmatically?

It seems impossible to debug.
>Are you talking about Bookmarks? (Google translated as 'original collection's portal location')

yes, it means Bookmarks, because Gaode, Google, Tencent and Baidu maps are offset in mainland China.

johnd0e:
> It seems impossible to debug.

Well, I'm sure that it's possible to use something like [this](https://stackoverflow.com/questions/9038625/detect-if-device-is-ios).
(Or even something more advanced, like [mobile-detect.js](http://hgoebl.github.io/mobile-detect.js/))

But better would be just fix iPhone issue (if possible), because it concerns all layers, not just this.

modos189:
About the adjustment of coordinates. Maybe add this to IITC? Then you need a way for the plugin to report the need to enable the adjustment.

johnd0e:
> Then you need a way for the plugin to report the need to enable the adjustment.

IITC could provide general transform function, and any plugin be able to use it.

johnd0e:
And I'm still thinking about Bookmarks etc..
May it be possible to leave coordinates intact, and just offset tiles screen position?
(just brainstorming)

magicbug2013:
> About the adjustment of coordinates. Maybe add this to IITC? Then you need a way for the plugin to report the need to enable the adjustment.

That's great. That the plugins only needs to define an offset to move the map layer.

magicbug2013:
> And I'm still thinking about Bookmarks etc..
> May it be possible to leave coordinates intact, and just offset tiles screen position?
> (just brainstorming)

yes, actually Bookmarks only needs to use portal's guid, it does not need to use latlngs.

magicbug2013:
> > And I'm still thinking about Bookmarks etc..
> > May it be possible to leave coordinates intact, and just offset tiles screen position?
> > (just brainstorming)
> 
> yes, actually Bookmarks only needs to use portal's guid, it does not need to use latlngs.

I'm wrong. We have to consider how to display bookmarks when protal is not displayed.

modos189:
> IITC could provide general transform function, and any plugin be able to use it.

Yes, IITC could provide general transform function, but it should affect other plugins. So, we need to make adjustments not only in autonavi.com, but also Google maps (which I failed to update) and Bookmarks plugin. Instead of making a Monkey patch like this, it's better to set some variable to include the coordinate correction.

By the way, @magicbug2013 , I tried to update the coordinate fix for the new Google maps library, but the changes don't apply. https://github.com/IITC-CE/ingress-intel-total-conversion/tree/fix-google-map-offsets-in-china
Maybe you can fix it

johnd0e:
>  it's better to set some variable to include the coordinate correction.

...so if IITC knows that current map needs correction, than it begins correct coordinates for all plugins.
And it stops correction as soon as we select another map. Right?

modos189:
Yes. The question is, should IITC have a list the maps that require adjustment, or should the plugins themselves call some function to enable

johnd0e:
Or may be it's possible to apply offset to map tiles only? In some way that coordinates stay intact.

modos189:
Um. Wait, why change the coordinates of the Bookmarks? After all, correction is necessary only for map tiles...

magicbug2013:
@modos189 
A few weeks ago, I tried to change the _update and _handleZoomAnim functions, as you did, but it didn't work, so now I'm trying to change the coordinates of protal, link and field. Now the problem is bookmarks. The best result is that don't change the coordinates of collection. just change the display position of the icon. I'll try it.

modos189:
Maybe we can make changes to the Leaflet to support correction?

johnd0e:
> Maybe we can make changes to the Leaflet to support correction?

In form of leaflet extension.
And at first we should check, may be such extension already exists, it's obvious that our project is not unique.

@magicbug2013, could you check if offset problem is solved here: https://github.com/htoooth/Leaflet.ChineseTmsProviders?

modos189:
It can also be useful
https://github.com/hobo0cn/leaflet-gcj02

Perhaps this solution is for autonavi.com, but Google maps we connect in a completely different way

magicbug2013:
> Maybe we can make changes to the Leaflet to support correction?

yes, we fixed the coordinates everywhere. In fact, we only need a function to move the map layer a little bit, but not any other layer.

magicbug2013:
> It can also be useful
> https://github.com/hobo0cn/leaflet-gcj02
> 
> Perhaps this solution is for autonavi.com, but Google maps we connect in a completely different way



> > Maybe we can make changes to the Leaflet to support correction?
> 
> In form of leaflet extension.
> And at first we should check, may be such extension already exists, it's obvious that our project is not unique.
> 
> @magicbug2013, could you check if offset problem is solved here: https://github.com/htoooth/Leaflet.ChineseTmsProviders?

The fixed method can be referred to, but I don't know how to apply it to the project.

magicbug2013:
> It can also be useful
> https://github.com/hobo0cn/leaflet-gcj02
> 
> Perhaps this solution is for autonavi.com, but Google maps we connect in a completely different way

He just added several maps of China and did not fix the coordinates.

modos189:
> He just added several maps of China and did not fix the coordinates.

probably you are confused with
https://github.com/htoooth/Leaflet.ChineseTmsProviders

johnd0e:
Here is one more similar project: https://github.com/sureleo/leaflet-baidu

johnd0e:
@magicbug2013, also consider `Code Style` section of [HACKING.md](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/HACKING.md#code-style).
> indent using two spaces

modos189:
@johnd0e here we have to figure out how to move the map, not the portals. Compared to this, indents are a trifle)

johnd0e:
I propose to separate discussions.

> we have to figure out how to move the map, not the portals

As it is more general question, then let's create dedicated issue. I can do it, if you don't mind.

modos189:
Ok

plugins/basemap-gaode-by-gmoogway.user.js
```diff
@@ -0,0 +1,351 @@
...
+// @name           IITC plugin: Map layers from GAODE by GMOogway
+// @category       Map Tiles
+// @version        0.2.0.@@DATETIMEVERSION@@
+// @description    [@@BUILDNAME@@-@@BUILDDATE@@] Add autonavi.com (China) map layers
+@@METAINFO@@
+// ==/UserScript==
+
+@@PLUGINSTART@@
+
+    // PLUGIN START ////////////////////////////////////////////////////////
```
johnd0e: Here should not be any indent, see other iitc plugins

plugins/basemap-gaode-by-gmoogway.user.js
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        // for iitc-mobile on ios
+        //localStorage['iitc-base-map'] = '高德 路网';
+
+        var newpos;
+
+        var PluginBookmarksExist = false;
+        var PluginPlayerTrackerExist = false;
+        if (window.bootPlugins) {
```
johnd0e: This is wrong approach, because this code may be executed before that plugins loaded.

plugins/basemap-gaode-by-gmoogway.user.js
magicbug2013: fixed

plugins/basemap-gaode-by-gmoogway.user.js
magicbug2013: fixed

magicbug2013:
@modos189 @johnd0e 
now the plugin works fine, maybe it can be merged?

modos189:
@magicbug2013 Let's not override other plugins and change the coordinates of the portals.

[I was able to create](https://github.com/IITC-CE/ingress-intel-total-conversion/commit/64060c4b78564c5d0d7d40f0603b48560cbd01ac) a coordinate correction that only moves the map. So far it does not work perfectly, but it is much better than changing the coordinates of the portals.

I am sure we can bring this decision to perfection. Join the task:
https://github.com/IITC-CE/ingress-intel-total-conversion/issues/64#issuecomment-454941754

magicbug2013:
> @magicbug2013 Let's not override other plugins and change the coordinates of the portals.
> 
> [I was able to create](https://github.com/IITC-CE/ingress-intel-total-conversion/commit/64060c4b78564c5d0d7d40f0603b48560cbd01ac) a coordinate correction that only moves the map. So far it does not work perfectly, but it is much better than changing the coordinates of the portals.
> 
> I am sure we can bring this decision to perfection. Join the task:
> [#64 (comment)](https://github.com/IITC-CE/ingress-intel-total-conversion/issues/64#issuecomment-454941754)

yes, that's better, It will be easier to add maps of China. I close this request, and re-fork the project.

johnd0e:
@magicbug2013 Wellcome to test #72 for Google offset.
And see there comment how to turn on offset correction for any map.

magicbug2013:
@johnd0e 
great works.  there are no problems for maps using gcj02, such as Google map, Gaode map and so on, but there are still many problems for Baidu map (bd09), which is not only a problem of coordinate offset, but also the problem of coordinate system and Title map download path . I think it's possible to integrate gcj02 modifications into boot. js, leaving bd09 alone, and add an options attribute  fixChinaMapOffsetType = 'gcj02'.

johnd0e:
Have you set  `needFixChinaOffset: 'Baidu'` for Baidu?

> which is not only a problem of coordinate offset, but also the problem of coordinate system and Title map download path .

Do you have plugin for Baidu?

johnd0e:
BTW, do you know about these PRs:
https://github.com/iitc-project/ingress-intel-total-conversion/pull/1211
https://github.com/iitc-project/ingress-intel-total-conversion/pull/1262

~Are they usable?~
Have tried to fix them?

**Update**
Here is quick and dirty attempt to update the plugin: https://gist.github.com/johnd0e/4509cd02b71d60510be28ec7ffb8bd38
Please check if it works for you.

PR #61, opened 2019-01-15
build.py: keep newlines in @@INCLUDESTRING

johnd0e:
* compiled sources become more readable
* easy to make changes in-place

### DIFF
--- a/build.py
+++ b/build.py
@@ -112,7 +112,7 @@ def readfile(fn):
 
 def loaderString(var):
     fn = var.group(1)
-    return readfile(fn).replace('\\', '\\\\').replace('\n', '\\n').replace('\'', '\\\'')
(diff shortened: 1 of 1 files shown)


johnd0e:
May be even better to add additional (leading) newline:
```
    return '\\\n' + readfile(fn).replace('\\', '\\\\').replace('\n', '\\\n').replace('\'', '\\\'')
```
I'm not sure...

modos189:
> May be even better to add additional (leading) newline:

I think it's unnecessary

johnd0e:
> I think it's unnecessary

Well, strictly speaking, the whole patch is not necessary, it just lets the resulting source looks better.
Without leading newline we get indentation for 1st line of included file.

But it isn't big issue, I agree.

Issue #56, opened 2019-01-14
Idea: development `test` branch


johnd0e:
I have an idea how we could facilitate our development process, delivering new features faster, while keeping `master` branch clean, neat and stable.

So, what I propose:
1. `master` branch is only for stable builds with well-tested (in other branches) changes.
Push to `master` means (auto)deploying new **stable** build.
2. For every new set of changes we use separate `feature`-branches (PRs), which should be made vs `master`.
PRs are keeping open until final merging into `master` (it's possible to squash/rebase them to make history cleaner).
3. `development` branch is for early-merging of feature-branches, as they go.
'Early' means that we could merge them not well-tested, then fix, remerge, (force)reset, etc, **not bothering** to keep history clean.
   * Cons: could be unstable.
   * Pros: features delivered earlier, which in turn let to test them thoroughly.
4. After we decide that feature is ready - we merge it into `master`, closing PR.

So instead of current `test` builds we get `development` builds (title is not important).

Main difference: there we could play with features more bravely, because we would not care about secondary things (such as clear history).

modos189:
Now with friends I have long argued about the use of git flow. I don't think he's right for us. Also I don't understand you proposed a schema, it also seems unnecessarily confusing.

We can use Releases ( https://github.com/IITC-CE/ingress-intel-total-conversion/releases ). No, we NEED use Releases to build releases, since we need to provide the changelog for Google Play and Telegram channel.
At the same time, really, I hasten to close PR, then not to forget about some of them.

What if we continue to use master for test builds but create a task with a list of changes needed for the new release? When all desired changes are made and merged, Release will be called?

johnd0e:
>  Also I don't understand you proposed a schema, it also seems unnecessarily confusing.

Well, perhaps my explanation was not good. I'll try to explain my motives.
Currently I'm trying in parallel to develop several features. All of them are work-in-progress, and not good enough to be merged into `master`. But the features are still usable.

Obviously, we can merge them right now, and fix later (if needed). But using `master` I feel 'bounded'.
Because I know that at some point I'll have to rewrite that experimental code. Obviously I cannot 'undo' `master`, so there left much 'waste' commits. And I care about it because a year later me, or you, or some other developer will have to dig through dirty unclear history.

[...]

PR #62, opened 2019-01-15
build.py: new macro @@INCLUDESVG to facilitate svg icons usage

johnd0e:
svg includes as text (for sake of readability)

Better with #61.

### DIFF
--- a/build.py
+++ b/build.py
@@ -126,6 +126,10 @@ def loaderImage(var):
         base64.encodestring(open(fn, 'rb').read()).decode('utf8').replace('\n', ''))
 
 
+def loaderSVG(var):
+    return 'data:svg+xml;utf8,' + loaderString(var)
+
+
 def loadCode(ignore):
     return '\n\n;\n\n'.join(map(readfile, sorted(glob.glob('code/*.js'))))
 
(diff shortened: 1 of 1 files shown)


johnd0e:
I was also going to apply this macro to existent code.

Still not sure that we can use this macro in CSS though.

You are merging too fast, please think about separate development branch (#56).

**edit**:
> Still not sure that we can use this macro in CSS though.

Seems to work (need to check different browsers).

johnd0e:
todo:
- [ ] `;utf8,` seems incorrect according [rfc2397]
Perhaps we should just omit that part (leaving `data:svg+xml,`), or use `data:svg+xml;charset=utf8,`
- [ ] Check if it's necessary to escape symbols like `#`.
(we can use [urllib.parse.quote]/[urllib.quote] if needed)
(or see [mini-svg-data-uri](https://github.com/tigt/mini-svg-data-uri/blob/master/index.js))
- [ ] Apply the macro to existing codebase
- [ ] Check if all is working in different browsers (mobile/desktop, webkit/firefox, etc)

[urllib.parse.quote]: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote
[urllib.quote]: https://docs.python.org/2/library/urllib.html#urllib.quote
[rfc2397]: https://tools.ietf.org/html/rfc2397

P.S.
```
import urllib

def loaderString(var, newline = '\\\n'):
    fn = var.group(1)
    return readfile(fn).replace('\\', '\\\\').replace('\n', newline).replace('\'', '\\\'')

#...
        try:
            from urllib import quote  # Python 2.x
        except ImportError:
            from urllib.parse import quote  # Python 3+
        return 'data:svg+xml,' + quote(loaderString(var, newline='\n'), safe='=:/\n').replace('\n', '\\\n')
```

Issue #64, opened 2019-01-16
Support maptiles based on GCJ02 (Chinese)


johnd0e:
**Background info**:
* https://en.wikipedia.org/wiki/Restrictions_on_geographic_data_in_China#GCJ-02
*  https://web.archive.org/web/20140116215242/http://home.wangjianshuo.com/archives/20081109_all_maps_in_china_are_transformed.htm

**Summary**: some map tiles providers use GCJ-02 coordinates, thus map appears with some offset.

_Originally posted by @modos189 in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MjQ0Njc3OTI3/pull_requests/unread_timeline#issuecomment-454773564_

johnd0e:
There are currently two attempts to fix the problem.
* ~PR #60 wip, _trying to change the coordinates of portal, link and field_ (seems not good way).~
* [fix-google-map-offsets-in-china](https://github.com/IITC-CE/ingress-intel-total-conversion/commits/fix-google-map-offsets-in-china) (~still nonfunctional~)

johnd0e:
Some links from former discussion:
* https://github.com/htoooth/Leaflet.ChineseTmsProviders (offset problem)
* https://github.com/hobo0cn/leaflet-gcj02 (Perhaps this solution is for autonavi.com, but Google maps we connect in a completely different way)
* https://github.com/sureleo/leaflet-baidu

johnd0e:
@modos189 could you explain the reason for this change: https://github.com/IITC-CE/ingress-intel-total-conversion/commit/6d3e28493cf5ba1ecc59b8d3432f99f8501b3b53?

Also I'm interesting what is the source of transforming algorithm.
I found similar:
* https://github.com/wandergis/coordtransform
* https://github.com/googollee/eviltransform/blob/master/javascript/transform.js
* https://github.com/Artoria2e5/PRCoords
* https://github.com/hiwanz/wgs2mars.js
* https://github.com/hujiulong/gcoord

And here are also different _'point-in-polygon'_ algorithm implementations: https://github.com/IITC-CE/ingress-intel-total-conversion/pull/65#issuecomment-455151352

modos189:
> @modos189 could you explain the reason for this change

Previously, the fix coordinates included most of India, all of South and North Korea, the Philippines, Vietnam, Mongolia, Thailand, and a number of other countries:
http://programmerz.ru/images/76604/BG9LI.jpg

[...]

PR #65, opened 2019-01-17
fix-googlemap-china-offset: fix style issues

johnd0e:
lint:
* `let` is not in es5
* 'HK_LENGTH' is assigned a value but never used
* 'original' is defined but never used

style:
* == vs ===
* Mixed spaces and tabs
* fix indents, missed/extra/trailing spaces, missed semi


johnd0e:
And there is still error in `L.GridLayer.prototype._setZoomTransform`
```
L.GridLayer.prototype._setZoomTransform = (function (original) {
  return function (level, center, zoom) {
    center = window.plugin.fixChinaOffset.getLatLng(center, this.options.type);
    original.apply(this, arguments);
  };
})(L.GridLayer.prototype._setZoomTransform);
```

* `zoom` is defined but never used
* `center` is assigned a value but never used

So `center` is not propagated back to `arguments`.

johnd0e:
> 'original' is defined but never used

I suppose this also should be fixed adding something like `original.apply(this, arguments);`

modos189:
Please update PR to new commit

modos189:
> So center is not propagated back to arguments.

No, it is propagated back. This is a bug in es linter, or browser?
Maybe `original.apply(this, [level, center, zoom]);`

johnd0e:
> No, it is propagated back. This is a bug in es linter, or browser?

You are right. I suppose bug in `eslint`.
Anyway, your second variant looks cleaner:
> Maybe original.apply(this, [level, center, zoom]);

And please comment about this:
> 'original' is defined but never used

Was it really unused argument?
Or there is missing `original.apply(this, arguments);`

modos189:
> Was it really unused argument?

Here, indeed, unused variable. I tried calling original method as in other cases, but then errors occur when calling class Bounds

johnd0e:
Merged (need to test again).

And I am going to review the code one more time.
E.g. I doubt that we need PRCoords as object. Wouldn't it better just have two distinct functions?

modos189:
Method isInGoogle can be completely removed. It is not used

johnd0e:
> It is not used

As I can see it is used as fast precheck if latlng is in bounds of square, which covering China.
Should be renamed.

modos189:
But why? The plugin still needs to be installed separately for those from China.

johnd0e:
Perhaps for those who are not China, but still want to see real map offset while browsing China.

Could you make some tests comparing performance between these two functions?
(May be we do not need optimization here)

johnd0e:
BTW, here is another implementation of same [pnpoly](https://wrf.ecse.rpi.edu//Research/Short_Notes/pnpoly.html)-algorithm: https://github.com/substack/point-in-polygon / https://github.com/iominh/point-in-polygon-extended

And here is improved version: https://github.com/mikolalysenko/robust-point-in-polygon

And here is different algorithm: https://stackoverflow.com/a/36070256/2520247.
See other answers too, as there are useful references to GeoJSON-libraries. Some of them:
- https://github.com/maxogden/geojson-js-utils#point-in-polygon
- http://turfjs.org/docs/#booleanPointInPolygon
- https://github.com/mapbox/leaflet-pip

And even more: https://libraries.io/search?keywords=points%2Cpolygon&languages=JavaScript&q=Point+polygon&sort=latest_release_published_at

modos189:
I'm confused. Method isInGoogle is used. It as roughly and thoroughly checked the coordinates.
This method calling from gcjObfs.

Another thing is that there is no need to make it public.

johnd0e:
> Another thing is that there is no need to make it public.

Well, it could be useful for someone, who wants to reuse this function.

modos189:
Um. Well, maybe as two distinct functions

johnd0e:
I failed to identify the source of obfuscation algorithm in our code.
As it differs from old (which was referenced as https://on4wp7.codeplex.com/SourceControl/changeset/view/21483#353936).
And it also differs from included with new link: https://github.com/Artoria2e5/PRCoords

So it's sort a mess...

I'm prefer that policy about external sources (and images) should be included unmodified (and inlined only in build-process).

johnd0e:
> Well, maybe as two distinct functions

Agree, as there are 2 distinct action, and there are several methods to do every of them.
So better to divide them in order to make support simpler (e.g. tomorrow we'll prefer to replace one function)

johnd0e:
Parent issue: #64

PR #66, opened 2019-01-17
fix-googlemap-china-offset: refactoring

johnd0e:
divide PRCoords into 2 distinct functions:
* plugin.fixChinaOffset.isInGoogle
* plugin.fixChinaOffset.gcjObfs

### DIFF
--- a/plugins/fix-china-offset.user.js
+++ b/plugins/fix-china-offset.user.js
@@ -1,5 +1,5 @@
 // ==UserScript==
-// @id             iitc-plugin-fix-china-offset@breezewish
+// @id             iitc-plugin-fix-china-offset
 // @name           IITC plugin: Fix maps offsets in China
 // @category       Tweaks
 // @version        0.2.0.@@DATETIMEVERSION@@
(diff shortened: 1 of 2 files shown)


johnd0e:
Oops, I will resubmit the PR..
You've changed filename and id, and this is not that simple, as you should leave stub (meta) for updating old version.

Also, if you change id, than better drop the part after `@`
`iitc-plugin-fix-china-offset@breezewish` --> `iitc-plugin-fix-china-offset`

(So as for me than easier to revert rename)

modos189:
I agree. Will you do?

johnd0e:
You mean I will rename it back in my PR?

modos189:
Changes move to `fix-china-offset`, but create file-dummy `fix-googlemap-china-offset`.

Or I can do it

johnd0e:
Done, but not tested (hurry right now).
I propose rename `fix-china-offset` to `basemap-fix-china-offset` or `fix-china-map-offset`.

johnd0e:
Parent issue: #64

PR #68, opened 2019-01-17
Fix google map offsets in china

modos189:
So, this plugin fixes coordinate correction for some maps in China. #64 

It is also now possible to enable correction not only for Google maps.


johnd0e:
It it possible to extend L.GridLayer via leaflet api (instead of monkey-patching)?

plugins/fix-china-map-offset.js
```diff
@@ -0,0 +1,313 @@
...
+})(L.GridLayer.prototype._getTiledPixelBounds);
+
+L.GridLayer.prototype._setZoomTransform = (function (original) {
+  return function (level, center, zoom) {
+    center = window.plugin.fixChinaOffset.getLatLng(center, this.options);
+    original.apply(this, [level, center, zoom]);
+  };
+})(L.GridLayer.prototype._setZoomTransform);
+
+L.GridLayer.GoogleMutant.prototype._update = (function (original) {
```
johnd0e: `original` is never used

```suggestion
L.GridLayer.GoogleMutant.prototype._update = (function () {
```

plugins/fix-china-map-offset.js
```diff
@@ -0,0 +1,313 @@
...
+  return function () {
+    this.options.needFixChinaOffset = true;
+    // zoom level check needs to happen before super's implementation (tile addition/creation)
+    // otherwise tiles may be missed if maxNativeZoom is not yet correctly determined
+    if (this._mutant) {
+      var center = this._map.getCenter();
+      /// modified here ///
+      center = window.plugin.fixChinaOffset.getLatLng(center, this.options);
+      /////////////////////
```
johnd0e: In order to make eslint happy
```suggestion
      /////////////////////
      /* eslint-disable */
```

plugins/fix-china-map-offset.js
```diff
@@ -0,0 +1,313 @@
...
+      var fractionalLevel = zoom !== Math.round(zoom);
+      var mutantZoom = this._mutant.getZoom();
+
+      //ignore fractional zoom levels
+      if (!fractionalLevel && (zoom != mutantZoom)) {
+        this._mutant.setZoom(zoom);
+
+        if (this._mutantIsReady) this._checkZoomLevels();
+        //else zoom level check will be done later by 'idle' handler
+      }
```
johnd0e:
```suggestion
      }
      /* eslint-enable */
```

johnd0e (changes requested):

plugins/fix-china-map-offset.js
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    L.GridLayer.prototype._update.call(this);
+  };
+})(L.GridLayer.GoogleMutant.prototype._update);
+
+window.plugin.fixChinaOffset.getLatLng = function (pos, options) {
+  // No offsets in satellite and hybrid maps
```
johnd0e: It is true for Google map, but it's not reliable to to extrapolate such assumption for arbitrary map.

So I propose to check real maps.
And may be we need additional configuration option.
```suggestion
  // No offsets in satellite and hybrid maps
  // TODO: it's true for Google maps, need to examine other available Chinese maps
```

modos189:
> It it possible to extend L.GridLayer via leaflet api (instead of monkey-patching)?

I don't know, it's the only way I can do(

plugins/fix-china-map-offset.js
modos189: No, this fix is true for all cards that have a correction.

Another thing is that in exceptional cases, when map is connected not by Leaflet, but by means of its own libraries, it may be necessary to correct loading of tiles in them. Only loading, but not coordinates.

johnd0e:
https://leafletjs.com/examples/extending/extending-1-classes.html
https://leafletjs.com/examples/extending/extending-2-layers.html
It's seems a matter of using of `L.GridLayer.extend(...)` (or sort of).
May be we even could create separate leaflet-plugin, usable in any project, not only iitc).

plugins/fix-china-map-offset.js
johnd0e: > No, this fix is true for all cards that have a correction.

Well, I mean next line:
```
  if (options.needFixChinaOffset === true && options.type !== 'satellite' && options.type !== 'hybrid') {
```
Are you sure that in other maps we also should skip correction for satellite and hybrid?

> when map is connected not by Leaflet

I suppose that it also could be fixed. But let's not run ahead.

plugins/fix-china-map-offset.js
```diff
@@ -0,0 +1,315 @@
...
+// WGS-84. As a result, the Ingress objects on maps would be incomplete.
+//
+// The algorithm of transforming WGS-84 to GCJ-02 comes from:
+// https://on4wp7.codeplex.com/SourceControl/changeset/view/21483#353936
+// There is no official algorithm because it is classified information.
+//
+// Here we use the PRCoords implementation of this algorithm, which contains
+// a more careful yet still rough "China area" check in "insane_is_in_china.js".
+// Comments and code style have been adapted, mainly to remove profanity.
+// https://github.com/Artoria2e5/PRCoords
```
johnd0e: I suggest to add the link as there is more info about source data of polygon.
```suggestion
// https://github.com/Artoria2e5/PRCoords (more info: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1188)
```

johnd0e:
As we named the plugin `fix-china-map-offset` we should include Baidu-related stuff here.
* Additional transformation needed: https://github.com/Artoria2e5/PRCoords/blob/2193bea83ade11726a61543f548e4f0d9d9304b2/js/PRCoords.js#L141-L154
* Boundaries slightly differ: https://github.com/Artoria2e5/PRCoords/blob/2193bea83ade11726a61543f548e4f0d9d9304b2/js/misc/insane_is_in_china.js#L151-L158

~~So our plugin api gets sort of complex..~~

johnd0e:
P.S.
`plugins/fix-china-offset.user.js → plugins/fix-china-map-offset.js`
Need to be ...`user.js`

See #72

plugins/fix-china-map-offset.js
johnd0e: https://en.wikipedia.org/wiki/Restrictions_on_geographic_data_in_China:
> Some map providers, such as Here, choose to also offset their satellite imagery layer to match the GCJ-02 street map.[34]

So we need more flexible options.

johnd0e:
>>It it possible to extend L.GridLayer via leaflet api (instead of monkey-patching)?

>I don't know, it's the only way I can do(

Well, we have a couple examples which do not require leaflet monkey-patching:

1. Look here: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1262
It doesn't work in our current environment, but I should be [fixable](https://gist.github.com/johnd0e/4509cd02b71d60510be28ec7ffb8bd38)
(_my fix is quick and dirty, just to make sure if it ever could be done_).
And I suppose `L.GaodeLayer` could be written in more general way, allowing to be reused by other plugins.

2. And here we see another way https://github.com/hobo0cn/leaflet-gcj02
Haven't tried yet, but it seems to be cleaner method.
**Edit**: CRS is property of map, not just TileLayer, so I'm not sure if we can use it..

May be it couldn't be used on google.com, but see: there is google.cn also.

modos189:
> And here we see another way https://github.com/hobo0cn/leaflet-gcj02
> CRS is property of map, not just TileLayer, so I'm not sure if we can use it..

Yes, this method does not suit us

johnd0e:
New feature idea: instead of hardcoding polygon we could allow import from draw-tools or [geojson](https://gist.github.com/johnd0e/2d61d1fec27da8fc09428fe82b84c3e6#file-china-boundaries-geojson): http://geojson.io/#id=gist:johnd0e/2d61d1fec27da8fc09428fe82b84c3e6&map=4/37.85/104.06

johnd0e:
To be **Squash and merge**

PR #69, opened 2019-01-18
Extracted leaflet monkey patching #3

modos189:
https://github.com/McBen/ingress-intel-total-conversion/commit/1158ade0aed0ef2dd36114a7014e5fa9e75defe6
https://github.com/McBen/ingress-intel-total-conversion/commit/a3d34cecc01b505051921c9974faf971aaf1f9f6

### DIFF
--- a/code/boot.js
+++ b/code/boot.js
@@ -589,6 +589,24 @@ window.setupLayerChooserApi = function() {
   }
 }
 
+function MonkeyPatching_Leaflet() {
+  var iconDefImage = '@@INCLUDEIMAGE:images/marker-icon.png@@';
+  var iconDefRetImage = '@@INCLUDEIMAGE:images/marker-icon-2x.png@@';
+  
+  L.Icon.Default = L.Icon.extend({
+    options: {
+      iconUrl      : iconDefImage,
+      iconRetinaUrl: iconDefRetImage,
+      iconSize     : new L.Point(25, 41),
+      iconAnchor   : new L.Point(12, 41),
+      popupAnchor  : new L.Point(1, -34)
+    },
+    _getIconUrl: function (name) {
+      return L.Icon.prototype._getIconUrl.call(this, name);
+    }
+  });
+}
+
 // BOOTING ///////////////////////////////////////////////////////////
 
 function boot() {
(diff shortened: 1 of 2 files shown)


code/boot.js
```diff
@@ -589,6 +589,24 @@ window.setupLayerChooserApi = function() {
...
+  var iconDefImage = '@@INCLUDEIMAGE:images/marker-icon.png@@';
+  var iconDefRetImage = '@@INCLUDEIMAGE:images/marker-icon-2x.png@@';
+  
+  L.Icon.Default = L.Icon.extend({
+    options: {
+      iconUrl      : iconDefImage,
+      iconRetinaUrl: iconDefRetImage,
+      iconSize     : new L.Point(25, 41),
+      iconAnchor   : new L.Point(12, 41),
+      popupAnchor  : new L.Point(1, -34)
```
johnd0e, line 602: Current leaflet version also has other options here: `tooltipAnchor` and `shadowSize`.

johnd0e:
I am not sure that we should apply this patch right now, as there are questions.

It could wait until we update Leaflet to 1.4 considering #70.

external/leaflet.css
```diff
@@ -464,7 +464,7 @@
 
 /* Default icon URLs */
 .leaflet-default-icon-path {
-    background-image: url(@@INCLUDEIMAGE:images/marker-icon.png@@);
+    background-image: url(""); /* requires Path: not URL! Patched in Boot */
```
johnd0e, line 467: I suppose that we don't need to patch this at all.
Perhaps it is sort of css fallback, for legacy browsers.

code/boot.js
johnd0e: * There is better way to extend `Icon.Default` class.
* And as we anyway [patching][leaflet history] sources, then it's easier just replace url with our @@include macro here
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/d77562b91a5b1e0bf170b4257eb33316a0b9daa3/external/leaflet-src.js#L7072-L7082

[leaflet history]: https://github.com/McBen/ingress-intel-total-conversion/commits/master/external/leaflet-src.js

external/leaflet.css
johnd0e: Actually `.leaflet-default-icon-path` is a dummy class used for `Icon.Default._detectIconPath`.
We shouldn't include inline image here.

PR #71, opened 2019-01-18
Better 'use strict';

johnd0e:
Nowadays there are no reasons why not use [`'use strict';`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) in all our scripts.

- core: could be inserted during build (#234)
- plugins cannot insert 'use strict' statement in random place, currently it must be part of wrapper (inside of wrapper function, but before any other statement).
  We have 0d571f4859c76dc117310eb8ef695c3ebbfec298, but imho that was not good, as we want strict for ALL standard plugins.
So instead of separate macro [`@@PLUGINSTART-USE-STRICT@@`](https://github.com/iitc-project/ingress-intel-total-conversion/pull/1187) it's more proper (and simple) just modify wrapper for all plugins (after #238).

But before we can make this modifications - we have to make sure that code itself is ready for strict mode.

### DIFF
--- a/build.py
+++ b/build.py
@@ -69,6 +69,10 @@ function wrapper(plugin_info) {
 // ensure plugin framework is there, even if iitc is not yet loaded
 if(typeof window.plugin !== 'function') window.plugin = function() {};
 
+"""
+
+pluginWrapperEnd = """
+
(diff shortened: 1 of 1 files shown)


johnd0e:
Wait a minute..
Why not even simpler: `include 'use strict';` into wrapper itself.

PR #72, opened 2019-01-19
update to 'Fix google map offsets in china'

johnd0e:
Misc updates for #68.

fix-china-map-offset: Baidu support

* Sources refactored to be closer to upstream https://github.com/Artoria2e5/PRCoords
* ~Baidu support~
* Clarify comment about usage

...and more:
* applied to googleMutant
* fix filename


johnd0e:
Todo:
1. > applied to googleMutant

   Perhaps it wasn't necessary and should be reverted.

2. > options for HongKong...

   It seems that it needed for Baidu only.
   If we could make plugin for Baidu it would have sense.
   But I'm not sure.
   (If not than this could be removed)

3. > ...and map type

   We need to test different maps.
   Perhaps this required for Google only, so condition need to be enhanced.

4. It also possible that we do not need check _'if in China'_ for every Chinese map.
   Need more maps' plugins to test...
   BTW, leaflet has interesting [function](http://leafletjs.com/reference.html#lineutil-simplify), that allows to simplificate our polygon, like here: http://mourner.github.io/simplify-js/ 

5. > (redefine leaflet methods cleaner)

   I suppose that we can make it even cleaner, not touching `L.GridLayer.prototype`.
  Will try tomorrow.

code/boot.js
```diff
@@ -141,7 +141,9 @@ function createDefaultBaseMapLayers() {
             stylers: [{visibility:"on"}, {hue:"#005eff"}, {invert_lightness:true}] },
           { featureType:"poi", stylers:[{visibility:"off"}]},
           { featureType:"transit", elementType:"all", stylers:[{visibility:"off"}] }
-        ]});
+        ],
+      needFixChinaOffset: true // managed by fix-china-map-offset plugin
+    });
   baseLayers['Google Roads'] = L.gridLayer.googleMutant({type:'roadmap', maxZoom: 21});
```
modos189, line 146: Here and below, too need add
`needFixChinaOffset: true`

johnd0e: Right.
So, we can add this option here, only for that layers that need it (excl. `hybrid` and `satellite`).
(But I am also going to try another way)

johnd0e:
I think that better leave boot.js untouched and patch layers options in plugin setup instead.
All layers are accessible via layerChooser._layers, and it's easy to patch only _mutant layers.
Will try tomorrow.

johnd0e:
@modos189 
I need an advice.

Currently we rewrite `L.TileLayer` class:
```
  L.TileLayer.include(fixChinaOffset);
```

A more proper way would be to extent it with new class:
```
  L.TileLayer.GCG02 = L.TileLayer.extend(fixChinaOffset);
```

And then instead of `needFixChinaOffset` option we would use that new class:
```
var mapLayer = new L.TileLayer.GCG02 (url, options);
```

But I doubt that we should make things more complex here..

johnd0e:
@modos189?

Issue #70, opened 2019-01-18
Manage upstream externals


johnd0e:
Currently we have separate sub-directory with all upstream 'externals' (+ separate `images` subdirectory)

Most of them described in [ATTRIBUTION.md](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/ATTRIBUTION.md).
Some of them have version info inside, but most of them not.
Some of them are even modified for iitc special needs, but in order to know details we have to dig though unclear commit history.

It's a pain to maintain and update these sources.

So I propose:
- [x] Update all 'externals' to current version

Then:
- [x] To document somewhere (for each 'external'):
  * version
  * url to upstream
  * add some comments
     + purpose, and where in iitc it is used
     + why was this component chosen
- [x] When updating write verbose commit message, incl. url / version info / hash.
- [x] If the source was modified, then we should:
  * add original unmodified
  * apply modification(s) as a separate commit(s), with verbose messages explaining purpose of changes
  * in case of heavy modification: keep separate branch, or make a fork  of original repo (could be hosted in https://github.com/IITC-CE/).
- [x] Find a way to avoid source modification
  * extend the component in some official way (if possible)
  * monkey-patch at run-time
  * push changes to upstream (via PR)
And in general:
- [x] Avoid in IITC/plugins own code inline including of 3rd-party code (and images).
Use `@@INCLUDExxx`-macros instead.

Related links:
- Remove customizations of third-party code library [#1165](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1165)
- Which Directory for Reusable Code? [#1185](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1185)

johnd0e:
> If the source was modified, then we should:

May be this workflow could be enough:
* Create branch like `external-leaflet-update`.
* Add unmodified source files.
* Commit needed changes separately, possibly in several commits (in order to make modification as clear as possible)
* Merge into `master` (feature-branch could be keeped)

johnd0e:
Here will be the list of all externals, where we should specify:
- current version (clear checkbox if it's below upstream)
- bold if the source was modified from upstream (specify exact mods)

(WIP)

### jQuery-related components:
- [ ] [**jQuery**](https://jquery.com/) (jquery-3.3.1.min.js) © [_JS Foundation_](https://js.foundation/): [[MIT](https://jquery.org/license/)]
[...]

PR #78, opened 2019-01-27
leaflet update

johnd0e:
**To be merged after #96 and #102.**

This PR is created to track progress of leaflet update to v1.4.0.

Before merging into master I want to figure out if we need these changes:
* [ ] https://github.com/McBen/ingress-intel-total-conversion/commit/b5b33fb1dab8a3e20ff79983196a28743b9f4ed3#diff-66cb4ebaec98b864101ad1f5bcbda09e
  (even if we need this, I suppose that more cleaner way is to extend `L.Popup` in [`oms.js`] 
* [x] #69 (Do we ever need monkey patching here?)
* [x] https://github.com/IITC-CE/ingress-intel-total-conversion/commit/67c17416adc0e82ffffb0148458329d2db882069
  Do we need **replace** default icons? Wouldn't it be better just add our icons with own names?

Even after merge I propose do not delete the branch, in order to easy view our modifications to the external's code (see #70).

[`oms.js`]: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/code/oms.js


johnd0e:
This branch should be rebased on #81.

johnd0e:
> This branch should be rebased on #81.

Done.
So #81 need to be merged first (before #78).

**Edit**: rebased again, now on #96.

johnd0e:
- [ ] Todo: use new `Map.panInside` instead of some current functions

johnd0e:
I suppose we don't need last commit (8953632dddafa2e5e3806b36a8d8a6d4eb361fc2).
Better patch/extend `L.CRS.EPSG3857` / `L.Projection.SphericalMercator`.

johnd0e:
Some thoughts:
1. Currently we do not use `L.Projection.Mercator`, so I'm absolutely sure that no need to modify `R` value there.
2. We use `L.Projection.SphericalMercator` (it's default), but I am not sure that we need modify `R` value here.
Are there any evidence that it leads to more precise linking? I want too see it, as long as the exact part of code (that depends on `R` value).
**We should not modify random values just hoping that it may help**.
3. There is one more value in leaflet: `L.CRS.Earth.R`, which **is not patched** now.
And actually this value used in distance [calculation](https://leafletjs.com/reference-1.4.0.html#latlng-distanceto).
4. Original commit message was `use the earth radius value used by the s2 geometry library`. But I failed to find this constant [there](https://github.com/google/s2geometry).
    **Edit**: [6367](https://github.com/google/s2geometry/search?q=6367), [6371](https://github.com/google/s2geometry/search?q=6371)

**Update**: my way is here: 1ebab824bcf7f1e4522afcd0d8d5a88bc4de7bf8

**Update 2**: in ingress code I was able to found only these constants: `6371`, `6353`, `6384`. <details><summary>https://intel.ingress.com/jsc/gen_dashboard_11c23c6500060287e050e93991e264fed07b0e4e.js:formatted</summary><p>

```
/*

  Latitude/longitude spherical geodesy formulae & scripts           (c) Chris Veness 2002-2014
   - www.movable-type.co.uk/scripts/latlong.html                                   MIT License
*/
function X(a, b, c, d) {
    if (!(this instanceof X))
        return new X(a,b,c,d);
    "undefined" == typeof c && (c = 0);
    "undefined" == typeof d && (d = 6371);
    d = Math.min(Math.max(d, 6353), 6384);
    this.lat = Number(a);
    this.a = Number(b);
    this.height = Number(c);
    this.b = Number(d)
}
```
</p></details>
So I suppose we do not need to modify default leaflet values unless we get any valuable reason.

**P.S.**
Ingress code uses some old revision of [this](http://www.movable-type.co.uk/scripts/latlong.html) lib: https://github.com/chrisveness/geodesy.

**Update 3**: physical meaning of different values (incl. [`6367`][Rectifying_radius]) mentioned in https://en.wikipedia.org/wiki/Earth_radius

[Rectifying_radius]: https://en.wikipedia.org/wiki/Earth_radius#Rectifying_radius

**Update 4**: The origin of the constant is probably here: https://github.com/google/s2-geometry-library-java/blob/c28f287b996c0cedc5516a0426fbd49f6c9611ec/src/com/google/common/geometry/S2LatLng.java#L31
(Java-port of S2 library, seems abandoned).

johnd0e:
And last mod: macros in css ~(474b5b37e3e83d6506a898ffc4a5af3880906a90)~.
I think that we can avoid this, using additional on-the-fly substitutions in `build.py` (#102)

PR #79, opened 2019-01-27
basemap-gaode: new plugin

johnd0e:
Sample China map to illustrate #68.

Currently we can get road map with slightly different look.
It is not straightforward, as there are not much info in official API [docs](https://lbs.amap.com/api/javascript-api/example/).

I've played with some parameters, and you can see result after uncommenting corresponding lines.
Some of the variants allow language change.


johnd0e:
P.S.
Different map themes could be seen here: https://lbs.amap.com/api/javascript-api/example/personalized-map/set-theme-style.
Unfortunately I've failed to figure out if it's possible to implement it in the plugin.

PR #81, opened 2019-01-29
leaflet: extend L.Icon.Default in more clean way

johnd0e:
* put ingress icons into `marker-ingress.png`, `marker-ingress-2x.png`
* restore original leaflet `marker-icon.png`, `marker-icon-2x.png`
* add `marker-shadow.png`
* use `L.Icon.Default.mergeOptions`

This PR is intended to superseed #69.


johnd0e:
So now default icon has shadow, you can see it using player-tracker plugin.

- [ ] But there is still no shadow for svg icons seen in draw-tools plugin.
Perhaps we need style it with `filter: drop-shadow(...)`.
- [ ] bookmarks-by-zaso also could use `shadowUrl` for `marker-star.png`

P.S.
Some general thoughts (for separate issues)
And in general icons usage in the project need to be unified.
- [ ] Use `svg` when possible.
  - [ ] Find more clean way to style icon with different colors (instead of current `marker-icon.svg.template`).
- [ ] For `png`: render hi-res icons for use with `iconRetinaUrl`.
- [ ] Do we really need to place images in dedicated directory? IMHO better place them together with corresponding sources.

johnd0e:
`add-kml.user.js` also need to be adapted

Issue #44, opened 2019-01-03
Portal info: layout breaks down it title is long


johnd0e:
When title is long enough (e.g.: [Ceglie Messapica - Chiesa Di San Rocco](https://intel.ingress.com/intel?ll=40.647331,17.51342&z=17&pll=40.647331,17.51342)) then it wraps on new line.

Title itself has `white-space:nowrap` style, so it wraps at whole, leaving on first line all other objects: location icon `#ic_place_24px`, and bookmark star (if plugin loaded).

The issue becomes more noticeable when `Show linked portals` is active (edit: fixed by https://github.com/IITC-CE/ingress-intel-total-conversion/pull/40/commits/61ec254d337e7c7bf4edc44bf18d9a22684ec1fe).

johnd0e:
Even longer title: [Spomenik kralju Aleksandru I Karađorđeviću](https://intel.ingress.com/intel?ll=43.318132,21.890947&z=17&pll=43.318132,21.890947)

PR #82, opened 2019-01-30
Fix: Portal info: layout breaks down it title is long #44

modos189:


johnd0e:
This solution has limitations.
The problem is that the line contain not only `title` but also 1-2 other elements (and some plugins may require to add even more)...

modos189:
I tested on two elements. Of course, this is not best way, but now plugins do not have method of adding icons, which leads to such situations. We do it right in the future, in a new design

johnd0e:
What if we put the title with all attached icons into additional container like here https://github.com/johnd0e/ingress-intel-total-conversion/commits/portal-info-fix
(unfinished) 

**Edit**
Failed to get ellipsis this way.

johnd0e:
Alternative patch: #83 
It handles title length more correct, and does not depend on number of attached icons (and their widths).

modos189:
best way: https://github.com/IITC-CE/ingress-intel-total-conversion/pull/83

PR #83, opened 2019-01-31
 portalinfo: put bookmark+location into title to keep all single line

### LINKED ISSUE #44: Portal info: layout breaks down it title is long

johnd0e:
Fix #44 

Minor issue here: title text is not vertically aligned with icons.

### DIFF
--- a/code/portal_detail_display.js
+++ b/code/portal_detail_display.js
@@ -105,18 +105,18 @@ window.renderPortalDetails = function(guid) {
     .html('') //to ensure it's clear
     .attr('class', TEAM_TO_CSS[teamStringToId(data.team)])
     .append(
-      $('<svg><use xlink:href="#ic_place_24px"></use></svg>')
-        .attr({
-          title: 'Click to move to portal',
-          class: 'material-icons  icon-button',
-        })
-        .click(function() {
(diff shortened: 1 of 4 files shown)


modos189:
> title text is not vertically aligned with icons

This can be fixed by assigning `float: left` to icon.

But overall it’s better than my #82

johnd0e:
Wait..
Maybe using `float: left` we do need to put icons into title.

BTW isn't 24x24 larger than we want for icon here?

modos189:
Maybe, if it doesn't break compatibility. For example, bookmark icon is aligned using `position`.

I thing, 24x24 is good

PR #84, opened 2019-01-31
Removed `Copy to clipboard` from share activity in another apps

modos189:
For some reason, previously displayed in third-party applications, which should not be


johnd0e:
But why do you want to remove it?

johnd0e:
After this change will it be still possible to copy location in iitc itself?

modos189:
Yeap

johnd0e:
I found that `Copy to clipboard` for 3rd-party apps was intended https://github.com/iitc-project/ingress-intel-total-conversion/issues/624

Issue #47, opened 2019-01-06
ES2015+?


johnd0e:
It's 2019, and ES2015+ features supports is getting better among modern browsers.
Several patches from #2 already using some ([mostly](https://github.com/iitc-project/ingress-intel-total-conversion/pull/1288/files) `let`/`const`).

So it would be great if we decide which of the features are 'safe' to use in whole project.
But to select particular features we need to consider target platforms of IITC, as some of them could be limited.

Perhaps most universal approach would be to:
- [ ] Include [some](https://babeljs.io/) transpilling tool in our build chain.

modos189:
Babel requires a node.js, which I would like to avoid.

I found a project that implements the Javascript interpreter on Python. As stated, he is not production ready, but it works.

And if there were no problems with connecting DukPy, I would prefer to choose TypeScript, not Babel. Because he too implements the capabilities of the new ES, but it also supports types.

modos189:
https://github.com/amol-/dukpy

johnd0e:
>  requires a node.js, which I would like to avoid.

I'm curious: why?
As we possibly also need other node.js-powered tools (such eslint, jslint, jshint, JSCS, webpack,...)

> I would prefer to choose TypeScript.

Well, TypeScript may be excellent, but it is not subset of ES*, and there is no any direct browser support of it (and possible never be).
But _option_ for using TypeScript would be great, as well as CoffeeScript (just example).

johnd0e:
Anyway (whatever tool is finally chosen), I want to know if you are actually going to implement it in build chain soon?
Could I start using some of ES2015 features now, or better to wait when build chain get ready?

modos189:
I don't want to use nodejs and npm because it's a huge additional dependency for the sake of just supporting the JS transpiler.

I tested DukPy and it doesn't suit us because of too low speed. At the same time, I really want to use at least ES2015. This standard is supported by all devices except Android < 5.0.

What if we use ES2015 for project, and for Android < 5.0 connect the browser version of Babel?

johnd0e:
>  for the sake of just supporting the JS transpiler.

May be not just transpiler, but some other tools.

> This standard is supported by all devices except Android < 5.0.

As far as I know not all the features of the standard is equally supported, so we should choose carefully...

> What if we use ES2015 for project, and for Android < 5.0 connect the browser version of Babel?

Well, we should try.

P.S.
As for node.js.
[...]

Issue #39, opened 2018-12-27
Uniform code style


johnd0e:
In current code base we can see now a variety of different styles.
It would be great to bring everything to some common code style.

So let's discuss here:

1. Elements of iitc-code uniform style 

2. Tools to implement style checking, and ways to include them in our CI-chain.

johnd0e:
So if we prefer to keep existing style's elements then we should use [eslint](https://eslint.org).
Otherwise, we could use some 'opinionated' formatter, like [prettier](https://prettier.io) or [StandardJS](https://standardjs.com/).

modos189:
Project code style is specified in
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/HACKING.md#code-style

But eslint only reports discrepancy to code style? I think utility itself should format code to our style

johnd0e:
> Project code style is specified in
> https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/HACKING.md#code-style

Great. I will make `.eslintrc` with that rules.

> But eslint only reports discrepancy to code style?

There is `--fix` option.
But not all issues could be resolves automatically.

>  I think utility itself should format code to our style

`prettier` is even more automatic, but it's rules are predefined, so we can use it only if we agree with the whole style.

P.S.
Anyway, we need to merge all old PR's before any global changes.

johnd0e:
BTW, found inconsistency in the document.
Here is a rule:
> * there should be a space after `if`, `for`, etc. E.g. `if (true) { doStuff(); } else { dontDoStuff(); }`

But see at example two lines upper:
> * opening brace on the same line: `if(blub) {`

johnd0e:
And perhaps we need to add some other rules to the list.
E.g. https://eslint.org/docs/rules/semi

johnd0e:
Here is the same specification with eslint rules' references added (`.eslintrc` will follow).
Perhaps we should add some more rules, so consider it's still a draft.

Code Style
----------

Please follow the these guidelines. Some are just preference, others are good practice.
- IITC sources currently meet [ES 5.1][ES5] standard
- [eqeqeq]: use identity operators: `===` and `!==`. [Why do I want this?](http://stackoverflow.com/a/359509/1684530)
- jQuery is your friend
- [indent] using two spaces
- enforce [semi]colon
- _'one true [brace-style]'_:
  - opening brace on the same line: `if (blub) {`
[...]

PR #100, opened 2019-02-12
A cross link autodetect tool

magicbug2013:
When you click on a line, it will automatically find all the corss links and identify them.

### DIFF
--- a/plugins/cross-link-autodetect-tool.user.js
+++ b/plugins/cross-link-autodetect-tool.user.js
@@ -14,7 +14,7 @@
 // use own namespace for plugin
 window.plugin.CLADT = function() {};
 
-window.plugin.CLADT.IS_DEBUG = true;
+window.plugin.CLADT.IS_DEBUG = false;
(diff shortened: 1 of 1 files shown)


johnd0e:
Thank you for this contribution.
Please note that we still avoid using ecmascript 2015+ features (#47).
This includes `const`, promises, `await`/`async`, arrow functions, destructuring assignment etc.

johnd0e:
`varφ`, `varλ`, `varθ`, `varδ`/`10`/`20`, `varΣθ` (space missed)

And some style issues (https://github.com/IITC-CE/ingress-intel-total-conversion/issues/39#issuecomment-455122280), like `semi`, `eqeqeq`, `quotes`.

magicbug2013:
OK, I'll try to fix these.

johnd0e:
There was no need to close the PR, usually it's better just to fix your original branch.

Also note that instead of `markerTemplate` you could use upcoming #96.

PR #104, opened 2019-02-15
Improve styles of text labels (portal-level-numbers, portal-names, portal-names, regions, keys-on-map)

johnd0e:
Some small but noticeable enhancements (mostly soften text shadows).
It's time to make our project visually different from stock iitc!


johnd0e:
There is still 1 issue with numbers: mobile font is not the same as desktop.

Desktop: monospace, zeroes are striked.
Mobile: ???

johnd0e:
> There is still 1 issue with numbers: mobile font is not the same as desktop.

@modos189 
Could you check it?

PR #109, opened 2019-02-19
Update gmap-gray plugin to GoogleMutant library

### LINKED ISSUE #73: basemap-gmaps-gray is not functional
`error starting plugin: index 27, error: TypeError: L.Google is not a constructor`

We can do one of these:
* raw-include old Google plugin here.
* re-base this plugin on GoogleMutant
* include this gray version in core (drawback: layer chooser gets overfilled)

modos189:
Fix #73


johnd0e:
image

Gray google is for less distraction, but I see some features have too high contrast.

**Edit**: this is not mutant-specific, old plugin had the same issue.
Can be fixed by changing gamma value:
```
      {featureType:"poi",stylers:[{saturation:-100},{visibility:"on"},{gamma:.34}]},
```

Also there are spots that are too bright.

Issue #12, opened 2018-12-06
Unable to reinstall apk: INSTALL_FAILED_UPDATE_INCOMPATIBLE


johnd0e:
_Edited_

Unable to reinstall apk with changed signature.
The problem appeared after uninstalling previous build truing to keep it's data with this command:
```
adb uninstall [-k] <package> - remove this app package from the device
                               ('-k' means keep the data and cache directories)
```

Please help.

johnd0e:
BTW, did you change something in apk build (e.g. signature)?
I just failed to install latest test apk over previous one, so I had to uninstall it at first.

modos189:
I didn't change debug signature, but I didn't spesify it. It is possible that each build uses new signature

johnd0e:
I've just run into problem, after uninstalled apk with this command:
```
adb shell pm uninstall -k org.exarhteam.iitc_mobile.debug
```

Now it's impossible to install apk again, because I have not stored that previous apk from 05/12.
And any other version just refuses to install.
May be you still have that version?

Or may be you can tell what should I clear in my system in order to be able to install new apk?
Already have deleted `/data/app/org.exarhteam.iitc_mobile.debug`, but some data still remains somewhere..

modos189:
> -k: keep the data and cache directories around after package removal.

You can try without this key. If not, what is the error if you install using

> adb install IITC_Mobile-debug.apk

johnd0e:
> You can try without this key

Too late, it is not working

> If not, what is the error if you install using

Error message is brief, can't remember exact text now, but something about wrong signature.

modos189:
maybe
`adb shell pm clear org.exarhteam.iitc_mobile.debug`

johnd0e:
This fails too

modos189:
`adb install -r IITC_Mobile-debug.apk`
?
Don't know what the problem is, can the error text to this command?

      push package(s) to the device and install them
     -l: forward lock application
     -r: replace existing application
     -t: allow test packages
     -s: install application on sdcard
     -d: allow version code downgrade (debuggable packages only)
     -p: partial application install (install-multiple only)
     -g: grant all runtime permissions

johnd0e:
`adb install IITC_Mobile-debug.apk` (plus any of mentioned key):
```
   Failure [INSTALL_FAILED_UPDATE_INCOMPATIBLE]
```
`adb uninstall org.exarhteam.iitc_mobile.debug`
```
[...]

PR #116, opened 2019-02-23
Request write permissions when adding userscripts from external programs

### LINKED ISSUE #34: More cases to request write storage permissions
> Request write storage permissions only if you add scripts.

When I open *.user.js at Chrome webpage (or in file manager) it proposes list of applications.
I select IITC and after confirmation all seems ok, but then `Plugins settings` screen appears without user plugins.

There should be permission request, I suppose.

And I think this even will not be enough, consider this case:
IITC was once installed and we have user plugins in `IITC-Mobile/plugins` directory.
Then we uninstall apk, and install it again.
Will we get permission request?

_Originally posted by @johnd0e in https://github.com/IITC-CE/ingress-intel-total-conversion/issues/4#issuecomment-446499237_

modos189:
> When I open *.user.js at Chrome webpage (or in file manager) it proposes list of applications.
> I select IITC and after confirmation all seems ok, but then Plugins settings screen appears without user plugins.

fix #34

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
@@ -1,5 +1,6 @@
 package org.exarhteam.iitc_mobile;
 
+import android.Manifest;
 import android.annotation.TargetApi;
 import android.app.Activity;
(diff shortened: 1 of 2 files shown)


johnd0e:
Unfortunately I am unable to install test build (#12) to check if all issues gone.

Issue #99, opened 2019-02-09
Scripts versioning


johnd0e:
Currently every version number contains date/time stamp as it's last component.

So with every build every script get newer version, even if it's content stays the same.

With this system if we turn on auto-updating feature - it will fire unconditionally, on every new build.

I don't think that it is good design, as typically we want to update changed scripts only.

angel93ayora:
"ideal" solution would be to only change timestamp when first component is updated, complex because last component is calculated on build time and you need the last build to be able to read the last component (timestamp) you built time ago

another shoulution would be to not inject timestamp, or to do it only on test versions

johnd0e:
> another shoulution would be to not inject timestamp, or to do it only on test versions

Exactly.

modos189:
I'll add that we need to reduce the number of dots to the version number. Some userscripts managers report an error

`// @version        0.3.0.20190103.162834`

`Required value' version' is missing or invalid. It must be between 1-4 dot-separated integers...`

modos189:
> another shoulution would be to not inject timestamp, or to do it only on test versions

I like this solution. And I need to remove point between date and time.

johnd0e:
So I propose to remove `.@@DATETIMEVERSION@@` completely, both for Release and Test-builds.
May be we should add it to local builds (discussion needed).

johnd0e:
> So I propose to remove `.@@DATETIMEVERSION@@` completely, both for Release and Test-builds.

- 'release' scripts do not need 'datetime' as version alone is just enough
- in 'test-builds' we either also need increment version after each commit, or continue using 'datetime' as 4th component.
  But even if we change only some 1 script - all other scripts in the project also get new 'datetime', which lead to total unconditional autoupdating.
  I do not like it.

But there is another option: what if 'datetime' would be not just 'build time', but actual timestamp of every file?

Bonus: it let us to (re)build only actually changed files.
BTW, the idea is not new:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/d77562b91a5b1e0bf170b4257eb33316a0b9daa3/build.py#L174-L176

gusowski1:
One reason where the @@datetimeversion@@ could make sense is the following use case.

Some code in the @@PLUGINEND@@ or @@PLUGINSTART@@ is changed the script itself is a newer version. In This case you need to bump the version of each script manually, otherwise the script itself won't be updated and the new code in the start and endsection wouldn't be changed.

[...]

PR #117, opened 2019-02-23
Reduce the number of dots to the version number

modos189:
Some userscripts managers report an error #99

before PR:
// @version 0.3.0.20190103.162834
→ Required value' version' is missing or invalid. It must be between 1-4 dot-separated integers...

after PR:
// @version 0.3.0.20190103162834
→ You're beautiful :)

### DIFF
--- a/build.py
+++ b/build.py
@@ -52,7 +52,7 @@ settings = buildSettings[buildName]
 utcTime = time.gmtime()
 buildDate = time.strftime('%Y-%m-%d-%H%M%S', utcTime)
 # userscripts have specific specifications for version numbers - the above date format doesn't match
-dateTimeVersion = time.strftime('%Y%m%d.', utcTime) + time.strftime('%H%M%S', utcTime).lstrip('0')
+dateTimeVersion = time.strftime('%Y%m%d', utcTime) + time.strftime('%H%M%S', utcTime).lstrip('0')
 
 # extract required values from the settings entry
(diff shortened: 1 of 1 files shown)


johnd0e:
And we need remove datestamp completely from deployed iitc (both test/release), right?

johnd0e:
> Some userscripts managers report an error #99

What managers do you mean? I know only about one old issue: https://github.com/iitc-project/ingress-intel-total-conversion/issues/1097
Not sure that we should take action without checking current versions of software.

And after this patch we have more serious issue: version length is not constant.
So when I trying to install today's build of plugin, tampermonkey proposes downgrade:
```
-// @version        0.2.0.20190227151132
-// @description    [local-2019-02-27-151132] Show correct maps for China user by applying offset tweaks.
+// @version        0.2.0.2019022884444
+// @description    [local-2019-02-28-084444] Show correct maps for China user by applying offset tweaks.
```

So if you are sure that this patch is really needed, then you better pad last version components with zeroes.

modos189:
Clearly, this error is only when installing directly from Chrome, which is no longer available. I now checked.

So I cancel PR

johnd0e:
FYI, I've [succeed](https://superuser.com/questions/767286/re-enable-extensions-not-coming-from-chrome-web-store-on-chrome-v35-with-enhan) at installing userscript directly in chrome (w/o tampermonkey).
image

Though, it fails to work for unknown reason

PR #135, opened 2019-03-13
Replacing build time on the time the file being changed in version nu…

modos189:
…mber

We do not specify the type of build, so the time is specified including for releases. But I do not think this is bad. #99

### DIFF
--- a/build.py
+++ b/build.py
@@ -51,8 +51,6 @@ settings = buildSettings[buildName]
 
 utcTime = time.gmtime()
 buildDate = time.strftime('%Y-%m-%d-%H%M%S', utcTime)
-# userscripts have specific specifications for version numbers - the above date format doesn't match
-dateTimeVersion = time.strftime('%Y%m%d.', utcTime) + time.strftime('%H%M%S', utcTime).lstrip('0')
 
 # extract required values from the settings entry
 resourceUrlBase = settings.get('resourceUrlBase')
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -216,7 +219,7 @@ def saveScriptAndMeta(script, ourDir, filename, oldDir=None):
 
 downloadUrl = distUrlBase and distUrlBase + '/total-conversion-build.user.js' or 'none'
 updateUrl = distUrlBase and distUrlBase + '/total-conversion-build.meta.js' or 'none'
-main = doReplacements(main, downloadUrl=downloadUrl, updateUrl=updateUrl)
+main = doReplacements(main, downloadUrl=downloadUrl, updateUrl=updateUrl, dateTimeVersion=getDateTimeVersion('main.js'))
```
johnd0e, line 222: Instead of just `main.js` we should iterate over `code/*.js` to find newest file.

johnd0e:
I browse through my iitc working directory and see that file timestamps do not correspond to their real modification date (from log).
And e.g. after `clone` all timestamps will be set to current date/time.

More info: https://stackoverflow.com/questions/2458042/restore-a-files-modification-time-in-git.

So if we want real file timestamp - we need to extract it from git log, separately for every file.

I doubt that we really need such complex setup.

modos189:
> I doubt that we really need such complex setup.

Yes. It's not that important.

PR #145, opened 2019-03-28
[sync] fix error in handleError(e)

mailz32:
When the plugin JSON on Google Drive was missing, Sync did nothing without any warnings but TypeError in debug console. I've spent some time to figure out that "error" object was enclosed within "result" object.


modos189 (approved):
Thanks for fix

modos189:
Having considered more variants of the data it appeared that it is necessary to use
`(e.error.message) ? e.error.message : e.result.error.message`.

Fixed in PR https://github.com/IITC-CE/ingress-intel-total-conversion/pull/212

PR #153, opened 2019-04-04
Added argument to start web server after build

modos189:
This change will allow in browser extension IITC-Button select a local repository to work with it. Once the build is complete, the changes in IITC and plugins are immediately available in the browser, making it easier to debug.

UP:

Example of use:
`python web-meta-run.py local` (start on port 8000)
or
`python web-meta-run.py local --port=1331`

Will be created file local.json and will be started simple webserver

### DIFF
--- a/build.py
+++ b/build.py
@@ -35,23 +35,13 @@ except ImportError:
     defaultBuild = None
 
 buildName = defaultBuild
-startWebServer = False
-startWebServerPort = 8000
 
 # build name from command line
 if len(sys.argv) == 2:  # argv[0] = program, argv[1] = buildname, len=2
     buildName = sys.argv[1]
-if len(sys.argv) == 3 and sys.argv[2].startswith('--webserver'):  # argv[0] = program, argv[1] = buildname, argv[2] = start web server, len=3
-    buildName = sys.argv[1]
-    startWebServer = True
-    port = sys.argv[2].split('=')
-    if len(port) == 2:
-        startWebServerPort = int(port[1])
 
 if buildName is None or buildName not in buildSettings:
     print("Usage: build.py buildname")
-    print("or")
-    print("Usage: build.py buildname --webserver=8000")
     print(" available build names: %s" % ', '.join(buildSettings.keys()))
     sys.exit(1)
 
@@ -201,43 +191,6 @@ def saveScriptAndMeta(script, ourDir, filename, oldDir=None):
             f.write(meta)
 
 
-def parseUserScript(file):
-    data = {}
-    for line in file.split('\n'):
-        if "==UserScript==" in line:
-            continue
-        if "==/UserScript==" in line:
-            return data
-
-        line = line.strip()
-        sp = line.split()
-        data[sp[1]] = ' '.join(sp[2:])
-
-
-def addPluginToMeta(meta, filename, script):
-    info = parseUserScript(script)
-    category = info.get('@category')
-    if category:
-        category = re.sub('[^A-z0-9 -]', '', category).strip()
-    else:
-        category = "Misc"
-
-    if category not in meta:
-        meta[category] = {
-            'name': category,
-            'desc': "",
-            'plugins': []}
-
-    meta[category]['plugins'].append({
-        'name': info['@name'].replace("IITC plugin: ", "").replace("IITC Plugin: ", ""),
-        'id': info['@id'],
-        'version': info['@version'],
-        'filename': filename+'.user.js',
-        'desc': info['@description'],
-    })
-    return meta
-
-
 outDir = os.path.join('build', buildName)
 
 # create the build output
@@ -278,7 +231,6 @@ with io.open(os.path.join(outDir, '.build-timestamp'), 'w') as f:
 # for each plugin, load, parse, and save output
 os.mkdir(os.path.join(outDir, 'plugins'))
 
-meta = {}
 for fn in glob.glob("plugins/*.user.js"):
     script = readfile(fn)
 
@@ -288,9 +240,6 @@ for fn in glob.glob("plugins/*.user.js"):
     script = doReplacements(script, downloadUrl=downloadUrl, updateUrl=updateUrl, pluginName=pluginName)
 
     saveScriptAndMeta(script, outDir, fn, oldDir)
-
-    if startWebServer:
-        meta = addPluginToMeta(meta, pluginName, script)
 
 # if we're building mobile too
 if buildMobile:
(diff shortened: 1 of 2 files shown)


modos189:
Actually, this is main reason why I decided to create IITC-Button extension. For now there is no easy and fast way to test changes in IITC and plugins.

johnd0e:
> For now there is no easy and fast way to test changes in IITC and plugins.

But you was always able to use local webserver + grease/tampermonkey, wasn't you?

What is the purpose of that extension?

modos189:
> you was always able to use local webserver + grease/tampermonkey

Partially. 
You cannot change the update channel from release to local updates. You must remove all release scripts and manually install the local ones.
Back - the same procedure.
In addition, running your own web server reduces threshold for entry into the development.

These problems are solved by extension. In addition, the IITC-Button extension makes it easy to include any plugins by selecting them in categories.

johnd0e:
> You must remove all release scripts and manually install the local ones.

It's true.

> In addition, running your own web server reduces threshold for entry into the development.

Well, here I do not agree as every developer needs python to build from sources.

> In addition, the IITC-Button extension makes it easy to include any plugins by selecting them in categories.

BTW, instead of python server I use this https://github.com/McBen/ingress-intel-total-conversion/blob/master/fileserver.js
And it provides categorized view.

P.S.
I'd like to try this new easy-button extension. Do you have chrome version?
I do not need it in webstore, just raw sources.

modos189:
> just raw sources

In the adjacent repository: https://github.com/IITC-CE/IITC-Button

johnd0e:
Just tried it. It's really more user-friendly that grease/tampermonkey.
I found how to change update channel, but there is no `local` option.

johnd0e:
About PR itself.
I have not saw the result in action yet (will try soon), but have 3 preliminary thoughts:
- what if we make `webserver` not commandline arg, but property of build target (buildsettings.py/localbuildsettings.py)
- alternatively: make webserver independent from build.py, to be able to run it in any moment.
  * Ideally: not to be tied to build process, but listen for folder changes instead.
  * or generate .json for any build, unconditionally, then we will be able to run server later.

modos189:
> I found how to change update channel, but there is no local option.

I will add this feature soon

> what if we make webserver not commandline arg, but property of build target

Well, that's one way to do it.

johnd0e:
1.
   ```
   Update channel "local" opened. Start a web server on port 8000
   Last generation of categories: 2019-05-20 15:30:44 UTC
   Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
   ```
   Is this message about `0.0.0.0` correct?

2. In fact I'm able to open `http://localhost:8000/`, but then I see full `iitc-ce` repo root directory listing.
    Is it intended?

3. This script creates json in repo root, but why not in `build/*`?
    Imho it'd be more natural. And no need to generate file if it is already there.

4. We can skip settings reading, as all built targets are already in `build/*`

5. Instead of `update_meta` worker it would be simpler to generate json on build, unconditionally.
   For maintainability may be better left code in separate file, and call it from `build.py`.

johnd0e:
BTW, wouldn't it more consistent to do it in IITC-Button itself?
- we start server in needed directory
- iitc-button makes json list, and keep it updated (looking for `.build-timestamp`)

modos189:
> Is this message about 0.0.0.0 correct?

It's Ok, it means that access is open from all local addresses. https://stackoverflow.com/questions/20778771/what-is-the-difference-between-0-0-0-0-127-0-0-1-and-localhost

> but then I see full iitc-ce repo root directory listing.

Yeah, that's the way it's supposed to be. Well, it's enough to display only `build` and `*.json`, but I don't see any security threats here.

> This script creates json in repo root, but why not in build/*?

This is exactly how it is done by analogy with the architecture of the site. But this is a good idea.

> wouldn't it more consistent to do it in IITC-Button itself

I tried to do it same as downloading updates from site. It would be excessive to download updates of .meta.js files every time for all plugins, including those that are disabled.

johnd0e:
> I tried to do it same as downloading updates from site. It would be excessive to download updates of .meta.js files every time for all plugins, including those that are disabled.

Do you mean that same json is generating for release/test-builds on iitc.modos189.ru?

modos189:
Yeap https://iitc.modos189.ru/release.json

johnd0e:
> Yeah, that's the way it's supposed to be. Well, it's enough to display only `build` and `*.json`, but I don't see any security threats here.

So I propose:
- put json files into target directories, so background `worker` become unneeded.
- expose only target directory, as it is enough.

Alternatively, we can expose whole `build` directory, and implement target switching in IITC-Button.

johnd0e:
> I tried to do it same as downloading updates from site. It would be excessive to download updates of .meta.js files every time for all plugins, including those that are disabled.

I have another opinion.
- meta files are tiny enough, and every other userscript manager downloads them all every time.
- still we can avoid of unneeded downloads looking for `.build-timestamp` (download meta only when `.build-timestamp` changes)

I certainly do not insist, and you may continue using json, but IMHO it is just excessive intermediate entity.

modos189:
Other userscripts managers download .meta.js only for active (or installed) plugins. We have task of downloading meta files for all plugins. 58 requests for only plugins just to check updates? Including the fact that the connection may be unstable.

johnd0e:
> 58 requests for only plugins just to check updates?

No. Just 1 request, to get `.build-timestamp`

modos189:
Ok, when update 58 requests + on request for each installed plugin. Or make it even more difficult to request only meta.js file for uninstalled plugins and user.js for installed plugins. All this is too complicated, provoking errors and a large number of requests.

johnd0e:
> Ok, when update 58 requests

Well, it's 40 KB uncompressed.

> request only meta.js file for uninstalled plugins and user.js for installed plugins.

Considering that IITC-Button is mostly for desktop, I would update all plugins at once, including uninstalled. It is 930 KB, or ~200 KB zipped.

> All this is too complicated, provoking errors and a large number of requests.

I see it opposite, as more simple approach)

modos189:
It's time to get back to work :)

Suburbanno:
Go guy!! excited for new release!!

johnd0e:
I have not viewed last commits yet, but please note: we have some unmerged PRs for `build.py` that may (and will) conflict with this PR (no problem, but better merge this last).

johnd0e:
Instead of `web-meta-run.py` I propose to use `postBuild` action provided in `buildSettings`.

PR #161, opened 2019-04-11
IITC-Mobile: 'Support DeX desktop mode' option

Whomiga:
Useful for automatically making the display to be desktop when running on the Samsung DeX desktop system while allowing for showing in regular mode on a smartphone without having the switch back and forth between Desktop and regular when moving between the two environments.

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
@@ -247,8 +247,8 @@ public class IITC_Mobile extends AppCompatActivity
 
     @Override
     public void onSharedPreferenceChanged(final SharedPreferences sharedPreferences, final String key) {
-		if (key.equals("pref_force_desktop")) {
-			mDesktopMode = sharedPreferences.getBoolean("pref_force_desktop", false);
+        if (key.equals("pref_force_desktop")) {
+            mDesktopMode = sharedPreferences.getBoolean("pref_force_desktop", false);
             mNavigationHelper.onPrefChanged();
         } else if (key.equals("pref_dex_desktop")) {
             mDexDesktopMode = sharedPreferences.getBoolean( "pref_dex_desktop", true);
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
@@ -79,6 +79,8 @@ public class IITC_Mobile extends AppCompatActivity
     private IITC_MapSettings mMapSettings;
     private IITC_DeviceAccountLogin mLogin;
     private final Vector<ResponseHandler> mResponseHandlers = new Vector<ResponseHandler>();
+    private boolean mDexRunning = false;
+    private boolean mDexDesktopMode = true;
     private boolean mDesktopMode = false;
     private Set<String> mAdvancedMenu;
     private MenuItem mSearchMenuItem;
@@ -95,6 +97,7 @@ public class IITC_Mobile extends AppCompatActivity
     private final Stack<String> mDialogStack = new Stack<String>();
     private String mPermalink = null;
     private String mSearchTerm = "";
+    private IntentFilter mDesktopFilter;
 
     // Used for custom back stack handling
     private final Stack<Pane> mBackStack = new Stack<IITC_NavigationHelper.Pane>();
@@ -108,8 +111,43 @@ public class IITC_Mobile extends AppCompatActivity
         }
     };
 
+    // Setup receiver to detect if Samsung DeX mode has been changed
+	private final BroadcastReceiver mDesktopModeReceiver = new BroadcastReceiver() {
+		@Override
+    	public void onReceive(Context context, Intent intent) {
+        	String action = intent.getAction();
+
+        	if ("android.app.action.ENTER_KNOX_DESKTOP_MODE".equals(action)) {
+            	// Samsung DeX Mode has been entered
+            	mDexRunning = true;
+            	mNavigationHelper.onDexModeChanged(true);
+        	} else if ("android.app.action.EXIT_KNOX_DESKTOP_MODE".equals(action)) {
+            	// Samsung DeX mode has been exited
+            	mDexRunning = false;
+            	mNavigationHelper.onDexModeChanged(false);
+        	}
+    	}
+	};
+	
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
+
+        // get status of Samsung DeX Mode at creation
+        Configuration config = getResources().getConfiguration();
+        try {
+            Class configClass = config.getClass();
+            if(configClass.getField("SEM_DESKTOP_MODE_ENABLED").getInt(configClass)
+                    == configClass.getField("semDesktopModeEnabled").getInt(config)) {
+                mDexRunning = true; // Samsung DeX mode enabled
+            }
+        } catch(NoSuchFieldException e) {
+            //Handle the NoSuchFieldException
+        } catch(IllegalAccessException e) {
+            //Handle the IllegalAccessException
+        } catch(IllegalArgumentException e) {
+            //Handle the IllegalArgumentException
+        }
+
         // enable progress bar above action bar
         // must be called BEFORE calling parent method
         requestWindowFeature(Window.FEATURE_PROGRESS);
@@ -157,6 +195,9 @@ public class IITC_Mobile extends AppCompatActivity
         mSharedPrefs = PreferenceManager.getDefaultSharedPreferences(this);
         mSharedPrefs.registerOnSharedPreferenceChangeListener(this);
 
+        // enable/disable mDexDesktopMode mode on menu create and url load
+        mDexDesktopMode = mSharedPrefs.getBoolean("pref_dex_desktop", true);
+
         // enable/disable mDesktopMode mode on menu create and url load
         mDesktopMode = mSharedPrefs.getBoolean("pref_force_desktop", false);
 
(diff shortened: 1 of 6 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
```diff
@@ -522,7 +572,7 @@ public void setCurrentPane(final Pane pane) {
     }
 
     public void switchToPane(final Pane pane) {
-        if (mDesktopMode) return;
+        if ((mDesktopMode) || (mDexRunning && mDexDesktopMode)) return;
```
johnd0e: The same condition repeats in different parts of code.
Shouldn't it be separate function?

Whomiga: I added the Dex code to where it was already being determined for the original Desktop, if I put it to a separate function should I pull the entire if statement?

Least amount of changes was by adding it to the original Desktop determination, this also allows for the original Desktop determination to override the DeX determination.

I didn't want to break something if I modified that and lost proper scope.

Whomiga:
Somewhere in the functionized change, I lost the DeX mode change...

johnd0e:
@Whomiga 
I have no Samsung to test, so I just ask:
- Is this patch finished?
- Does it completely solve your original issue?
- Is IITC fully usable in DeX-mode (considering #162)?

Whomiga:
This patch, as written, does solve the original issue. Noticed #162 while testing the changes, but I believe the problems with DeX was preexisting, and unconnected to my changes. The IITC problems in DeX Mode will require further testing.

johnd0e:
> I believe the problems with DeX was preexisting, and unconnected to my changes.

Sure, unconnected.
But if IITC in DeX is barely usable than this patch is not needed too.

Whomiga:
I would recommend keeping the patch on the inactive list by cancelling the pull request until more has been determined about the problems with DeX mode. In any case, I have a testable situation here and can continue to test with my branch as necessary.

Whomiga:
I've found that the DeX problems were a Red Herring - the problem is that the draw tools plugin, and maybe other stuff, has trouble being used with a connected mouse and appears to only work correctly when used by touch.

Whomiga:
@johnd0e - If IITC Mobile is used in DeX mode with the Phone acting as a Touchpad, it works perfectly, I believe that this Pull Request is quite valid as it does what it says and only showed an evidently preexisting problem.

johnd0e:
So this PR is ready to merge, right?

Whomiga:
Yes, I see no problem with merging this pull request, if there are no further questions, I'll merge it (If I can, not sure how that goes)

Whomiga:
@johnd0e Question: How do I properly merge this pull request and close it. It has been a while and I don't know if I have Write Access to Merge, etc...

johnd0e:
We will merge it later, with the rest of PRs.

Whomiga:
Any way that I can move this pull request from the master branch to another branch, like "DeX-Updates", so I can do other work while leaving this to be merged later? (Didn't plan ahead)

johnd0e:
Well, you could (by creating new PR), but it'd be totally pointless.
Just leave `master` and create new branch in your fork to continue work.

johnd0e:
@Whomiga
This option needs better name/description.
And do you want it to be enabled by default?

Is it possible to remove it for DeX-less systems?

Whomiga:
The option is totally ignored on DeX-less systems and defaults to not-selected in default settings.

The main feature of this is that it will automatically come up in a Desktop more if the phone is in DeX mode, otherwise it will come up in whatever mode is selected in settings. For a phone, that setting is usually in regular mode due to the display on the phone.

If selected it will 'force' the default behavior of IITC to be displayed in a Desktop mode when the DeX is active (This would be a phone connected to a DeX setup) so that it acts as you would expect on a Desktop environment. If not in a DeX environment or selected, it would default to the normal settings which could be set to a Desktop environment or a Phone environment depending on the user's requirements.

I mostly added this because I use my phone in both a normal environment and a DeX environment and got tired of resetting to/from Desktop every time I switched from Dex to Phone. This way it would come up as a Desktop in the DeX environment and a regular mobile environment outside of DeX.

I figured anyone who uses their phone in both DeX and Mobile environments would be able to use this capability.

> 
> @Whomiga
> This option needs better name/description.
> And do you want it to be enabled by default?

It is disabled by default, as that causes it to use the original setting mode with no respect to the phone even having the DeX mode capabilities.
 
> Is it possible to remove it for DeX-less systems?

See no reason why it is necessary, as the setting value is totally ignored on phones that can't do DeX
(maybe a note in the release indicating that it has no effect except on DeX systems)

johnd0e:
> I figured anyone who uses their phone in both DeX and Mobile environments would be able to use this capability.

Sure. I know it's purpose.


> It is disabled by default

When I open Settings - I see it in enabled state.


> See no reason why it is necessary

Reason: to not confuse users of most other systems: they have no clue what is that "DeX".

Whomiga:
> 
> 
> > I figured anyone who uses their phone in both DeX and Mobile environments would be able to use this capability.
> 
> Sure. I know it's purpose.
> 
> > It is disabled by default
> 
> When I open Settings - I see it in enabled state.

Oops, sorry it appears I was wrong.  But if the phone doesn't have a Samsung DeX mode, it wouldn't matter what the value was set to.

> > See no reason why it is necessary
> 
> Reason: to not confuse users of most other systems: they have no clue what is that "DeX".

You could make it "Samsung DeX" vs "DeX" - if anyone wanted more into they could look it up online or make the name and description...
"Force Samsung DeX desktop mode" 
"Force desktop mode on systems when Samsung DeX mode is active"

I mostly picked the title "Force DeX desktop mode" because of the "Force desktop mode" settings title of the previous settings item.

Whomiga:
I changed the preference to disabled by default, but I really think that anyone who uses Samsung DeX would want the desktop mode by default if they are using the Samsung DeX mode.

mobile/app/src/main/res/values/strings.xml
```diff
@@ -113,6 +113,8 @@
         Note: IITCm can still be controlled via the Navigation Drawers</string>
     <string name="pref_force_desktop">Force desktop mode</string>
     <string name="pref_force_desktop_sum">Nice for tablets, looks awful on smartphones</string>
+    <string name="pref_dex_desktop">Force Samsung DeX desktop mode</string>
```
johnd0e:
```suggestion
    <string name="pref_dex_desktop">Support Samsung DeX Mode</string>
```

mobile/app/src/main/res/values/strings.xml
```diff
@@ -113,6 +113,8 @@
...
+    <string name="pref_dex_desktop_sum">Force desktop mode on systems where Samsung DeX mode is available and active</string>
```
johnd0e:
```suggestion
    <string name="pref_dex_desktop_sum">Force desktop mode when DeX mode is active</string>
```

johnd0e (changes requested):
How about this wording?

mobile/app/src/main/res/values/strings.xml
Whomiga: To nit-pick:
Technically DeX is supported whether this is selected or not, this setting just forces IITC to always open in Desktop mode when running on an Active DeX system rather than having to toggle Desktop on and off when switching between and Active DeX system and normal mobile phone mode. 

But, this would work.

Whomiga: I can easily make the change in my copy and submit it. What is the thought of making it on/off by default?  I really think that anyone using a DeX mode - which is the only time this would be of any use would want it on.

Whomiga:
Default on/off is the only remaining question, I think.

johnd0e:
Well, make it 'on' if you wish. 
And should we merge it squashed or maybe you want to make some cleanup in this branch?

johnd0e:
@Whomiga?

Whomiga:
> 
> 
> @Whomiga?

Sorry, been away for a few days...

Do I need to do any cleanup? I believe all the changes are in this pull request to be merged.

mobile/app/src/main/res/xml/preferences.xml
```diff
@@ -31,6 +31,12 @@
             android:key="pref_force_desktop"
             android:summary="@string/pref_force_desktop_sum"
             android:title="@string/pref_force_desktop"/>
+
+        <CheckBoxPreference
+            android:defaultValue="false"
```
johnd0e: So you've decided to keep it disabled

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
```diff
@@ -200,8 +247,11 @@ public boolean onEditorAction(final TextView v, final int actionId, final KeyEve
 
     @Override
     public void onSharedPreferenceChanged(final SharedPreferences sharedPreferences, final String key) {
-        if (key.equals("pref_force_desktop")) {
-            mDesktopMode = sharedPreferences.getBoolean("pref_force_desktop", false);
+		if (key.equals("pref_force_desktop")) {
+			mDesktopMode = sharedPreferences.getBoolean("pref_force_desktop", false);
```
johnd0e: Wrong indentation

johnd0e (commented):
@Whomiga

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
Whomiga: Hmmm... my local copy shows that it is enabled, but diffs don't show it as being different than on the server...

Whomiga: I just committed a new copy that should have a fixed indentation and the item enabled by default, but the change doesn't show in diffs on the server

Whomiga: Looks like I finally got all those changes pushed to the server

Issue #163, opened 2019-04-12
Some minor bugs to fix later


johnd0e:
(It's a reminder for myself to fix the bugs later, during refactoring)

johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/256f9c53a866a254b40e40b1f766e58b4d2a89f4/main.js#L28
TypeError when user is not logged in, see https://github.com/iitc-project/ingress-intel-total-conversion/pull/1249#discussion_r274823201

johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/256f9c53a866a254b40e40b1f766e58b4d2a89f4/plugins/farms-find.user.js#L254

1. This part of code executed on script initial load, before `setup`, so `google` could be not defined
2. I doubt that we really need to change prototype
3. I doubt that we ever need `google` api for this plugin.
   And what if current baselayer is not google (need to check).

johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/256f9c53a866a254b40e40b1f766e58b4d2a89f4/code/utils_misc.js#L411-L412

1. `layerGroup._leaflet_id` is private property
2. And it does not correspond to `layerChooser._layers` index, so this function just cannot work properly
3. Actually it is not used in any known plugin
4. All this `layerChooser` numerous stuff needs to be refactored and placed in single class (instead of random parts of source files).

johnd0e:
Here is missed styler for default ingress map:
```
{featureType:"road",elementType:"labels.icon",stylers:[{invert_lightness:!0}]}
```

PR #164, opened 2019-04-15
IITC-Mobile: fix bugs on login page

### LINKED ISSUE #154: Google authorization does't work on Android KitKat
The authorization form is not displayed correctly, data cannot be sent.

Temporary solution:
Enable *Fake User Agent* in *Settings → Advanced Settings*.

Explanation:
This results to loading of another authorization form that runs on an outdated KitKat browser engine.

### LINKED ISSUE #160: IITCm signin layout
After recent intel update there is new button which does not fit well:
<details>

IITC screenshot-1288478869

</details>

modos189:
Fix sign-in buttons layout, which was broken after Niantic added facebook authorization, and more.
Fix #160, fix #154, fix (part of) https://github.com/IITC-CE/ingress-intel-total-conversion/issues/163#issuecomment-482516494

### DIFF
--- a/main.js
+++ b/main.js
@@ -26,24 +26,25 @@ document.body.onload = function() {};
 //however, that's already been executed, so we can just access PLAYER - no messing around needed!
 
 var PLAYER = window.PLAYER || (typeof unsafeWindow !== 'undefined' && unsafeWindow.PLAYER);
-if (typeof(PLAYER)!="object" || typeof(PLAYER.nickname) != "string") {
+if (!PLAYER || !PLAYER.nickname) {
   // page doesn’t have a script tag with player information.
-  if(document.getElementById('header_email')) {
+  if (document.getElementById('header_email')) {
     // however, we are logged in.
     // it used to be regularly common to get temporary 'account not enabled' messages from the intel site.
     // however, this is no longer common. more common is users getting account suspended/banned - and this
     // currently shows the 'not enabled' message. so it's safer to not repeatedly reload in this case
-//    setTimeout('location.reload();', 3*1000);
-    throw("Page doesn't have player data, but you are logged in.");
+    // //setTimeout('location.reload();', 3*1000);
+    throw new Error("Logged in but page doesn't have player data");
   }
   // FIXME: handle nia takedown in progress
   
   // add login form stylesheet
-  headHTML = document.getElementsByTagName('head')[0].innerHTML;
-  headHTML += '<style>@@INCLUDESTRING:login.css@@</style>';
-  document.getElementsByTagName('head')[0].innerHTML = headHTML;
+  var style = document.createElement('style');
+  style.type = 'text/css';
+  style.appendChild(document.createTextNode('@@INCLUDESTRING:login.css@@'));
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -25,7 +25,7 @@ document.body.onload = function() {};
 //originally code here parsed the <Script> tags from the page to find the one that defined the PLAYER object
 //however, that's already been executed, so we can just access PLAYER - no messing around needed!
 
-var PLAYER = window.PLAYER || (unsafeWindow && unsafeWindow.PLAYER);
+var PLAYER = window.PLAYER || (typeof unsafeWindow !== 'undefined' && unsafeWindow && unsafeWindow.PLAYER);
```
johnd0e: The matter of taste, but I still think that one condition is excessive
```suggestion
var PLAYER = window.PLAYER || (typeof unsafeWindow !== 'undefined' &&  unsafeWindow.PLAYER);
```

main.js
```diff
@@ -37,6 +37,12 @@ if (typeof(PLAYER)!="object" || typeof(PLAYER.nickname) != "string") {
     throw("Page doesn't have player data, but you are logged in.");
   }
   // FIXME: handle nia takedown in progress
+  
+  // add login form stylesheet
```
johnd0e, line 41: Should we change style for desktop IITC?

main.js
modos189: what if `typeof unsafeWindow !== 'object'`,

main.js
modos189: Well, why not? CSS `media` can create an adaptive view for all screen resolutions

main.js
johnd0e: Why? Can you imaging real case?

main.js
johnd0e: Ok. I have no objections (though not seen new design yet).

main.js
modos189: Oh. `===`, not `!==`.
But I agree, it makes no sense, because in JavaScript everything is an objects.

johnd0e:
@modos189 

<details><summary>width >=970 (responsive)</summary>

image
</details>

<details><summary>width: <970 (too wide)</summary>

image
</details>

<details><summary>width: <969 (unresponsive)</summary>

image
image

</details>

Caused by 
```css

#dashboard_container {
    min-height: 500px;
    min-width: 930px;
}
```

I suppose we should allow lower values, say, `min-width: 245px; min-height: 365px`.

modos189:
on mobile:

<details>
<summary>before #f222ab6</summary>

изображение

</details>


<details>
<summary>after #f222ab6</summary>

изображение


</details>

I suggest add

```
@media (max-width: 970px) { 		
    .button { 		
        margin: 5px 5%;	
    } 		
}
```

<details>
<summary>then in look like this</summary>

изображение


</details>

johnd0e:
> on mobile:

I do not see much difference on mobile (both variants seem acceptable).
But at desktop I do not like too wide buttons with small margins.

modos189:
Ok

PR #167, opened 2019-04-20
Region Scoreboard Enhances

angel93ayora:
Added texts for:
TEAM leads by N
N checkpoints remaining
Next checkpoint:    DATE
Cycle ends:    DATE

Improved the table to show accumulated MUs and difference in MUs with previous checkpoint

### DIFF
--- a/code/region_scoreboard.js
+++ b/code/region_scoreboard.js
@@ -175,7 +175,7 @@ function regionScoreboardScoreHistoryTable(result) {
   html += '</div><br>';
 
   // table build
-  html += '<div><table class="checkpoint_table"><thead><tr><th>CP.</th><th colspan="2">Enlightened</th><th colspan="2">Resistance</th></tr>';
+  html += '<div><table class="checkpoint_table"><thead><tr><th>CP.</th><th colspan="3">Enlightened</th><th colspan="3">Resistance</th></tr>';
   html += '<tr><th></th><th>Sum.</th><th>Score</th><th>Diff.</th><th>Sum.</th><th>Score</th><th>Diff.</th></tr></thead>'
   for(var i=0; i<history.length; i++) {
 
--- a/code/region_scoreboard.js
+++ b/code/region_scoreboard.js
@@ -128,20 +128,77 @@ function regionScoreboardScoreHistoryChart(result, logscale) {
 
 function regionScoreboardScoreHistoryTable(result) {
   var history = result.scoreHistory;
-  var table = '<table class="checkpoint_table"><thead><tr><th>Checkpoint</th><th>Enlightened</th><th>Resistance</th></tr></thead>';
+  var max_i = history.length - 1;
 
-  var sum = [0, 0];
+  var html = '';
 
-  for(var i=0; i<history.length; i++) {
-    sum[0] +=  parseInt(history[i][1]);
-    sum[1] +=  parseInt(history[i][2]);
-    table += '<tr><td>' + history[i][0] + '</td><td>' + digits(history[i][1]) + '</td><td>' + digits(history[i][2]) + '</td></tr>';
+  // Summatories and diffs calculation
+  var summatory = [];
+  var diffs = [];
+  
+  summatory[max_i] = [parseInt(history[max_i][1]), parseInt(history[max_i][2])]
+  diffs[max_i] = [0, 0];
+  for (var i = max_i - 1; i >= 0; i--) {
+    summatory[i] = [];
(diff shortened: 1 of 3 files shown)


johnd0e:
Hi!
We [are going](#2) to merge this: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1045 (by original author of the plugin).

What do you think about it?
Is your patch applicable in this case?

angel93ayora:
I like https://github.com/iitc-project/ingress-intel-total-conversion/pull/1045 i can implement my changes along with them.

angel93ayora:
Should i commit to this PR including the changes of https://github.com/iitc-project/ingress-intel-total-conversion/pull/1045 ??

johnd0e:
> Improved the table to show accumulated MUs and difference in MUs with previous checkpoint

Could you attach sample screenshot here?

> Should i commit to this PR including the changes of [iitc-project/ingress-intel-total-conversion#1045](https://github.com/iitc-project/ingress-intel-total-conversion/pull/1045) ??

Make PR vs [`scoreboard-improved`](https://github.com/IITC-CE/ingress-intel-total-conversion/tree/scoreboard-improved) branch.

angel93ayora:
> Could you attach sample screenshot here?
Yes, i thought i've uploaded one

2019-04-29 00_22_00-Ingress Intel Map

this is with enlarged width (+30px)

image
this is with original width

i think it'll be tighter and more readable to make numbers shorter with K / M suffixes, example: 1.5M MU

johnd0e:
Someone may need full unshortened values..

I have an idea: what if your new table data would be optional?
And we could turn it on via checkbox, or just longtap

angel93ayora:
> Someone may need full unshortened values..

them can be put on a title attribute


> I have an idea: what if your new table data would be optional?
> And we could turn it on via checkbox, or just longtap

you mean to switch full values vs short?

i could do two switchable views, compact and full

johnd0e:
> them can be put on a title attribute

It is hardly accessible for touch devices


> you mean to switch full values vs short?

I mean `full without diff` vs `short with diff`

angel93ayora:
so it could be:

compact
- short numbers
- cp# | cp_score | diff

full:
- full numbers
- cp# | cp_score | summatory

or togglers for each column, this can be done easy with css

johnd0e:
> cp# | cp_score | diff

IMHO 'diff' does not have much sense without 'summatory', so I mean `cp# | cp_score | summatory | diff`

And looking again to this.. Is that 'diff' really useful? Could you explain?

angel93ayora:
> And looking again to this.. Is that 'diff' really useful? Could you explain?

i.e when you made some fields before check and you want to know how much they scored

summatory is probably less useful:
i.e when you want to know how much MUs are needed to win a cycle, but only last summatories are useful in this case

johnd0e:
Probably I understand wrong.
There are 3 columns on your screenshot: Sum. | Score | Diff.

1. We need Sum: it's historic value for every checkpoint
2. Score gain seems useful too
3. Diff is difference between last and next Score gains
   Actually I do not see how I could use this precise values.

angel93ayora:
diff numbers are the best metric for measuring MU activity in comparison with previous checkpoint

> i.e when you made some fields before check and you want to know how much they scored

or when you destroy some fields and want to see how much mu you destroyed (aprox. because other players may have built/destroyed some fields)

johnd0e:
> diff numbers are the best metric for measuring MU activity in comparison with previous checkpoint

It's not clear why should we compare with prev. CP, why not with average gain, or some other random metric.

> or when you destroy some fields and want to see how much mu you destroyed (aprox. because other players may have built/destroyed some fields)

Here you mean Score column, right?

angel93ayora:
>It's not clear why should we compare with prev. CP, why not with average gain, or some other random metric.

because it (diffs) reflects the MU variation on the previous 5h


>>or when you destroy some fields and want to see how much mu you destroyed (aprox. because other >>players may have built/destroyed some fields)

>Here you mean Score column, right?

here i meant diff column also

yes, you can figure out what was the increase/decrease of MUs comparing with the previous checkpoint, the diff column is that but you dont have to subtract scores in your mind

angel93ayora:
i've seen that it is already done on the graphic, so i'll delete it on the table in order to save width pixels. Also it's not clear if it is useful.

angel93ayora:
Closed, i'll PR from a branch forked from https://github.com/iitc-project/ingress-intel-total-conversion/pull/1045

PR #168, opened 2019-04-20
fix 'build' badge in README.md

### LINKED ISSUE #98: v0.29
@modos189
So what features do you want to include in nearest release?
Would you like to use Project#1 to track the progress?

modos189:
fix 'build' badge https://github.com/IITC-CE/ingress-intel-total-conversion/issues/98#issuecomment-477723056

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ To build the mobile app, along with python, you will need
 
 Run `build.py mobile` to build IITC Mobile in debug mode.
 
+### IITC-Button
+
+The source codes of the IITC-Button browser extension are in a nearby repository:
+https://github.com/IITC-CE/IITC-Button
+
 ### Website
 
 The source codes of the iitc.modos189.ru site are in a nearby repository:
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Ingress Intel Total Conversion (IITC)
 =====================================
 
 <!-- [START badges] -->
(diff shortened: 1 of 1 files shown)


johnd0e:
1. I've updated description 'cause former notation has special action (autoclose corresponding issue), which is not relevant in this case, and #98 still has unresolved questions.

2. I do not think that proposed fix is right.
Our main branch is `master` and I suppose that badge should display `master`'s build state.

modos189:
1. I updated the description in the telegram channel, but forgot to report it.

2. `master` branch is not used for building, and it always contains a working version, as it is a release version.

johnd0e:
> 2\. `master` branch is not used for building, and it always contains a working version, as it is a release version.

Exactly: it is release version.

If we decide to use badge for some other branch - it should be explicitly pointed.

BTW, currently this badge is not useful, because our build system lacks tests.
Badge shows success even when there are errors preventing from page loading.
(the same apply to docs badge)

johnd0e:
> 1. I updated the description in the telegram channel, but forgot to report it.

But iitc site also need this warning.
And I even think we should keep it there pinned at least several months.

johnd0e:
Something wrong:
image

Why 2 badges?

modos189:
Because you opened Diff. Correct link: [README.md](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/628dd1d06d5b1e19ec84f4961b0796fd574d4476/README.md)

johnd0e:
Ok then. Let's merge it soon.

PR #169, opened 2019-04-20
IITC-Mobile: fix 'Send screenshot' function

### LINKED ISSUE #31: Mobile debug console: command line background color
It is white.
And both it's buttons (map/clear) is also white, so it's hard to view them.

In 'official' iitc command line background color is dark, the same as background of console itself.

### LINKED ISSUE #140: Screenshots has no portal and links
In mobile version, when sharing screenshot with map view, there is no portal circles and links/fields on a result image. Portal names if it turned on in plugins appears.

modos189:
Canvas mode is default in IITC-CE, and due to hardware acceleration, portals and links were not displayed when using build-in screenshot function.
Fix #140.

Also:
- fix file sharing on Android >6
- (unrelated) fix #31: background color of debug command line

Note: current implementation is not reliable enough, so 'Send screenshot' menu item is not shown by default.

### DIFF
--- a/mobile/app/src/main/AndroidManifest.xml
+++ b/mobile/app/src/main/AndroidManifest.xml
@@ -155,6 +155,16 @@
             android:noHistory="false"
             android:theme="@android:style/Theme.Holo.Dialog"/>
 
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="${applicationId}.provider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/provider_paths"/>
+        </provider>
+
         <!-- START Used for Samsung Multi-Window support -->
         <uses-library
             android:name="com.sec.android.app.multiwindow"
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
@@ -15,6 +15,7 @@ import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.Bitmap;
 import android.graphics.Bitmap.CompressFormat;
+import android.graphics.Canvas;
 import android.net.Uri;
 import android.nfc.NdefMessage;
 import android.nfc.NdefRecord;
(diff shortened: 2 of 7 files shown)


johnd0e:
> Disable screenshot button.
> After enabling hardware acceleration for HTML Canvas in new browsers, portals and links are not displayed on screenshots. More: #140

I propose:

1. Leave item in disabled state, instead of invisible.
2. Reintroduce Canvas mode option, with inverted value (I mean 'enabled' by default).
3. Do not disable item if canvas off.

I could make javascript part.

modos189:
What for? Screenshot can be done using Android

johnd0e:
> What for? Screenshot can be done using Android

1. Perhaps not every android version has screenshot feature enabled. Why ever this function was initially created?

2. IITC buildin screenshot function can be more convenient, as it is fullscreen (no need to manual enter fullscreen), and do not leave temporary files.

3. [need to check more]

modos189:
I agree that it is more convenient to click on button and immediately send a screenshot to some chat.
But I thing this is not so important a feature that the user decides to go into settings and deliberately disable hardware acceleration.
Besides, it looks like... low-quality stuff...

modos189:
I reread [android documentation](https://developer.android.com/reference/android/webkit/WebView#enableSlowWholeDocumentDraw()) and found out I was wrong.

Calling `WebView.enableSlowWholeDocumentDraw()` method does not disable hardware acceleration.

This only affects the optimization of the visible part of the page. For our application, theoretically, it absolutely does not matter.

Need to be tested more thoroughly, but it looks like we can leave the screenshot button.

johnd0e:
> But I thing this is not so important a feature that the user decides to go into settings and deliberately disable hardware acceleration.

Well, I'd prefer hardware acceleration. But some user might have other opinion. So why brake this thing for him.

> For our application, theoretically, it absolutely does not matter.

We should try (but I really doubt that iitc can keep same performance with this method).

modos189:
I was able to fix the function of creating screenshots. Tested on my devices and in emulator.
Hardware acceleration is switched off and on, so there is a delay before taking a screenshot.
Maybe someone can improve this hack.

johnd0e:
@modos189 
Travis-ci fails to build this.

https://github.com/IITC-CE/ingress-intel-total-conversion/commits/test-builds

johnd0e:
Pushed this to `test-builds`.
But testing it is up to you, because I'm unable to install apk.

johnd0e:
> But testing it is up to you, because I'm unable to install apk.

At least on some systems this fix does not help..
https://t.me/iitc_group/2107

MysticJay:
Not working:
see https://github.com/IITC-CE/ingress-intel-total-conversion/pull/245#issuecomment-528392679

johnd0e:
@MysticJay 
It is not related to #245, so better copy your report into this issue, and I clean up there.

MysticJay:
Android 9.0.5 (OxigenOS on Oneplus3)
build 0.9.1.2190905.142213

Screen shot does not work (clears layers, counts down, then offers apps to share with). Creating empty map.

MysticJay:
Recommendation: drop Screenshot, use Android's build in fetaure.
its not fullscreen anyway, limited to current frame size.

johnd0e:
What we can do:
- apply this PR, as it works for some users
- remove `Send screenshot` item from menu _by default_. It still can be added with `Advanced Settings / Configure IITCm menu`
- try to find some other way, e.g. using https://github.com/tsayen/dom-to-image (or [similar](https://github.com/tsayen/dom-to-image/issues/278))

johnd0e:
Currently 'Fix share...' commit fails in some cases.
1. On my phone (Android 6) I get sharing menu with several applications, but sharing fails with most of them.
Previously used method (without `FileProvider`) worked fine.

2. ~Here described some additional code: https://medium.com/@benexus/dealing-with-permissions-when-sharing-files-android-m-cee9ecc287bf#8729
I haven't tried, but it may help with issue 1.~

P.S.
Google guide: [Sending simple data to other apps](https://developer.android.com/training/sharing/send#java)

johnd0e:
Todo: try Android MediaProjection
- https://blog.csdn.net/u010057266/article/details/53470239
- https://blog.csdn.net/u010057266/article/details/72760442
  - https://github.com/shark5/ScreenCapture

PR #157, opened 2019-04-08
Fix miscellaneous bugs in core and plugins 

johnd0e:
- cache-portals-on-map: fix #103: _TypeError_ while running hook `mapDataEntityInject`
- layer-count: fix #149 (error accessing private property)
- bookmarks: fix #166: Bookmarks reorder error
- draw-tools: fix #174: Circle edit is broken (actually `GeodesicCircle` was lost at some point)
- draw-tools: fix #178: Circle is hard to draw
  Notes:
  - it's workaround only, actual bug is somewhere deeper
  - same problem is circumvented also by #175
- utils_misc.js:
  - fix `removeLayerGroup` (#190)
  - fix `escapeHtmlSpecialChars`  (#211)
- missions:
  - fix exception in `onWaypointsRefreshed`/`onMissionsRefreshed`
- boot.js: `setupLargeImagePreview`
  - fix title (#215)
  - fix dialog was not properly centered
- boot.js: fix `layerChooser.getLayers`
  to be again compatible with [existent](https://apps.apple.com/app/iitc-mobile/id1032695947) iOS mobile version

PR #175, opened 2019-04-26
draw-tools: fix mobile-related bugs; add Circle again

### LINKED ISSUE #162: drawtools doesn't appear to work correctly with mouse under IITC Mobile
Works just fine on same phone in regular phone mode

### LINKED ISSUE #1294: draw-tools on mobile: unable to draw circle
We can live without it, so (until fixed) it is better to remove icon from toolbar in order to save some place (it's important on small mobile screen).

It would be even better if we could customize draw panel defining which icons to hide.

johnd0e:
### Summary:
* Fix Circle drawing on touch devices, so it's returned to toolbar.
* Fix drawing with mouse on touch device.
* Fix: it wasn't possible to cancel Circle drawing.

### Details:

1. To work on touch Leaflet.draw (for some internal purposes) uses `touchExtend` handler.
   Previously it was not initialized as map was created before Leaflet.draw loading
   (https://github.com/Leaflet/Leaflet.draw/issues/923)
   So we enable it:
   - It's now possible to draw Circle/Rectangle on touchscreens
      (fix https://github.com/iitc-project/ingress-intel-total-conversion/issues/1294)
   - It's now possible to use mouse on touchscreen devices
      (fix https://github.com/IITC-CE/ingress-intel-total-conversion/issues/162)

2. Unfortunately `touchExtend` handler is broken (seems in many ways), that leads to misc. issues,
    the most serious is https://github.com/Leaflet/Leaflet.draw/issues/789 (unable to draw polylines/polygons).
    Some workaround proposed in discussion of that issue, but none is perfect.
    Here in draw-tools we use another workaround: enable `touchExtend` handler only when it is needed (see 1)

3. And last fix required patching of a upstream sources 
   https://github.com/Leaflet/Leaflet.draw/issues/922
   - It's now possible to cancel Circle/Rectangle drawing.
     This fix works fine in IITC, but is not perfect (see discussion)

P.S.
Some more draw-related fixes here: #157

### DIFF
--- a/external/leaflet.draw-src.js
+++ b/external/leaflet.draw-src.js
@@ -1258,7 +1258,7 @@ L.Draw.SimpleShape = L.Draw.Feature.extend({
 			// (update): we have to send passive now to prevent scroll, because by default it is {passive: true} now, which means,
 			// handler can't event.preventDefault
 			// check the news https://developers.google.com/web/updates/2016/06/passive-event-listeners
-			document.addEventListener('touchstart', L.DomEvent.preventDefault, {passive: false});
+			this._map.getPanes().mapPane.addEventListener('touchstart', L.DomEvent.preventDefault, {passive: false});
 		}
 	},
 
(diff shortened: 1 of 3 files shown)


modos189:
Now the objects are drawn immediately when you click, which makes it impossible to move the map without drawing. Can this be fixed?

johnd0e:
> Can this be fixed?

Unfortunately, it is default behavior of Leaflet.draw: https://github.com/Leaflet/Leaflet.draw/issues/935
So we have to return to previous workarounds: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/162#issuecomment-487042972

It is sad..

modos189:
So this behavior can't be done only for a Circle tool?

johnd0e:
> So this behavior can't be done only for a Circle tool?

It can, and it was done so in the very first commit (6ea92c6b639c4f1010a99b1de49db98ec336ff47).
And after your report I've returned to that code again.

Note, that this behavior is not for Circle only.

Issue #143, opened 2019-03-28
Performance issues


johnd0e:
When we have a lot of features active on map - leaflet performance degrades dramatically.

It's really hard to use map with a lot of portals. And it prevents us to work on #121 (and similar).

It's not clear yet what is the main source of performance lost.
I'm pretty sure that it's not portal markers themselves, as `L.CircleMarker` uses _Canvas_ renderer.

For start, here are some known solutions:
- https://github.com/cyrilcherian/Leaflet-Fullcanvas
- https://github.com/eJuke/Leaflet.Canvas-Markers
- https://github.com/yakitoritabetai/Leaflet.LabelTextCollision

Related issue: #129

johnd0e:
Recently I've seen degraded performance at low zoom levels during Ingress event, when two Europe cities had a lot of portals with ornaments.

johnd0e:
Right now we can observe performance lost over Amsterdam, and it is caused by ornaments. Zooming, panning are extremely slow.
After turning ornaments off performance become normal.

There are now ~1500 ornaments, visualizing by `L.Icon`, and visible even on low zoom levels.

modos189:
Leaflet.Canvas-Markers plugin significantly speeds up work of map with ornaments. But I could not connect it correctly. It seems that there is some conflict over two canvas in page.

johnd0e:
> But I could not connect it correctly.

Have you set up a branch? I would like to look.

johnd0e:
> Related issue: #129

For ornaments we could also try implement some custom 'clustering': detect overlapping markers, and do not draw icons for them.

modos189:
> > But I could not connect it correctly.
> 
> Have you set up a branch? I would like to look.

Yes, here's what I got.

https://github.com/IITC-CE/ingress-intel-total-conversion/commit/1543be0080a22ab58cf52db548df3663b87a1039

johnd0e:
> 1543be0

I see now.
And there are some errors in dev console.
The upstream repo is archived, but can search for fixes in forks: https://github.com/eJuke/Leaflet.Canvas-Markers/network

modos189:
Good news: I took refactored version from https://github.com/Spaction/Leaflet.Canvas-Markers (had to fix a typo, now can't compile) and then there is no conflict between canvas.

Now I need to integrate this with IITC layers

johnd0e:
Great!

> Now I need to integrate this with IITC layers

[...]

PR #180, opened 2019-04-28
A significant increase in performance of map with clusters

modos189:
Partial solution of issue #143

Now ornaments are drawn as circleMarker on canvas.

### DIFF
--- a/code/ornaments.js
+++ b/code/ornaments.js
@@ -45,7 +45,7 @@ window.ornaments.addPortal = function(portal) {
   if (portal.options.data.ornaments) {
     window.ornaments._portals[guid] = portal.options.data.ornaments.map(function(ornament) {
       var layer = window.ornaments._layer;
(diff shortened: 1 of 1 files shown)


johnd0e:
What if we make plugin for this?

modos189:
Noo. Why?

Only a small number of users will know about problem and turn on plugin, but most users if they open area with a lot of ornaments will receive a stop responding mobile application (at least, I have so) and will swear what a bad IITC.

johnd0e:
Ok.
Compare my implementation https://github.com/johnd0e/ingress-intel-total-conversion/commit/e9e82e12c0ef489b86716cc43e928662be2c20f4

modos189:
Your implementation is very good, but new PR https://github.com/IITC-CE/ingress-intel-total-conversion/pull/181 is better )

PR #189, opened 2019-05-04
Leaflet controls: prevent mobile style on desktop

johnd0e:
1. Modify `L.Browser.touch` in order to prevent mobile-styled controls in desktop (Leaflet v1 issue)

2. Fix leaflet.draw toolbar icons alignment in desktop mode

### DIFF
--- a/code/boot.js
+++ b/code/boot.js
@@ -784,6 +784,7 @@ function boot() {
 @@INCLUDERAW:external/load.js@@
 
 try { console.log('Loading included JS now'); } catch(e) {}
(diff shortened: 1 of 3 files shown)


johnd0e:
(1)
After upgrading Leaflet to v1 we get mobile-styled controls in desktop Chrome/Opera/etc.

Background info: https://github.com/nrenner/brouter-web/issues/69

I am pretty sure that Leaflet behavior (at least) is not completely right.
The simplest method would be to just remove `leaflet-touch` style from `#map` div.

Unfortunately out there is other code (e.g. in `L.Draw`) using `L.Browser.touch` property directly, so I propose to disable it.

Edit: actually it is not webkit-specific anymore, as I observe the same in Firefox.

(2) Toolbar in desktop mode:
image

You see draw toolbar is not ok.
It's definitely Leaflet.draw bug (easy to fix).

PR #198, opened 2019-05-13
Refactor to use common function window.makePermalink (utils_misc.js)

johnd0e:
Portal permalink is now shorter as it contain only `pll=` part.


johnd0e:
(continued in #267 and #281)

PR #201, opened 2019-05-15
Increase map renderer padding (and make value customizable)

### LINKED ISSUE #184: padding option
We have old code [changing](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/code/boot.js#L184) `L.Path.CLIP_PADDING`, but this property is not used in current Leaflet, there is Renderer option [`padding`](https://leafletjs.com/reference-1.4.0.html#renderer-padding) instead.

Question: should we ever change defaults?

johnd0e:
- use renderer option instead of `L.Path.CLIP_PADDING` (which is not used in Leaflet 1.x)
- remove value customization code, as currently it doesn't make much sense
- experimental: set padding to `0.5`, both for Canvas and SVG
  This value can be overridden by option `window.PADDING`
  (need to be set in some plugin, before `setup`)

Fix https://github.com/IITC-CE/ingress-intel-total-conversion/issues/184

Other changes:
- possibility to not prefer canvas
- fix some code where canvas-related options need to be adapted lo leaflet 1.x
- change some values in `RenderDebugTiles`

### DIFF
--- a/code/boot.js
+++ b/code/boot.js
@@ -175,7 +175,9 @@ window.setupMap = function() {
 //    zoomAnimation: false,
     markerZoomAnimation: false,
     bounceAtZoomLimits: false,
-    preferCanvas: true // Set to true if Leaflet should draw things using Canvas instead of SVG
+    preferCanvas: 'PREFER_CANVAS' in window
+      ? window.PREFER_CANVAS
+      : true // default
   });
   if (L.CRS.S2) { map.options.crs = L.CRS.S2; }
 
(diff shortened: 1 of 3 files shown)


johnd0e:
This PR needs to be tested, may be not `1` but other value better suits our needs.
- test performance (mobile/desktop)
- test experience

As for experience, we should compare with stock intel.
(IMHO there is large padding area)

modos189:
In stock Intel each tile is a separate canvas, but if approximately, it equals our PADDING = 1

modos189:
I don't notice any drop in performance, but I see an increase in the average RAM usage from 128mb (padding 0.1) to 256mb (padding 1) in IITC-Mobile. I think padding 0.5 will be the best compromise.

johnd0e:
> I think padding 0.5 will be the best compromise.

done

Issue #76, opened 2019-01-22
Plugins loading order


johnd0e:
Some things are not obvious and should be documented somewhere.

In fact IITC core may be loaded before plugins, after plugins, or somewhere in the middle.
Although in some environments we could control the order, but if scripts are loading asynchronously, then bigger scripts will finish loading later (no matter of initial order).

So in general we can't expect particular order of IITC/plugins loading.

But plugin's function `setup` is called when iitc is booted, and we can rely on this.

So the rule:
 **Plugin can use all service API's (IITC/jQuery/Leaflet/...) only after `setup` called**

johnd0e:
But sometimes order is important, and here we can run into issues, because there is no reliable way just to execute some plugins before others.

So it's essential to have some control under execution order.
- some plugins provide service for others
- some plugins 'rewrite' functions others (and need to be executed after)
  - or they rewrite options (and so need to be executed before)

I want to discuss here all possible issues.

johnd0e:
1. There are plugins which rewrite/extend functions of other plugins.

It is possible if other plugin's functions is already defined at the moment of 'extending'.
So it's not usually a problem, if following conditions met:
* Plugin to be extended must define functions in it's namespace **outside** of `setup`.
* Other (extending) plugin must rewrite functions **inside** it's `setup`.

Otherwise we need some additional tricks to ensure correct order of operations.

johnd0e:
2. There are plugins which provide service via hooks (pluginCreateHook/addHook).

And we can run into issue if some plugin calls `addHook` before such hook created:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/d77562b91a5b1e0bf170b4257eb33316a0b9daa3/code/hooks.js#L99-L104

I suppose this check should be removed (do not see any other way to resolve the issue).
**Edit**:  the same was proposed here: https://github.com/iitc-project/ingress-intel-total-conversion/issues/1248

johnd0e:
3. Some plugins may want to extend Leaflet, or load some other lib, or provide utility function, or any 'feature'.

And for plugins which are in need of that 'feature' we have to use special handling, like this:
* At first, plugin should check (in some relevant way) if the 'feature' is already available. If it is then immediately **use** it.
* If not then **wait** until it will get available.

Currently there is no nice way to resolve this.
The best we can do now: wait by putting pending actions into hook (`addHook`), and that hook should be provided by service plugin (when the issue from previous comment get resolved).

[...]

PR #205, opened 2019-05-19
Implement plugins priority control

johnd0e:
Improvements related to https://github.com/IITC-CE/ingress-intel-total-conversion/issues/76:

* Plugin priority order is now controlled.
  Priority can be specified as `setup` property (see details in commit msg).
* On error in setup:
   - show plugin name in log
   - strike-through plugin name in About window
* `pluginCreateHook` function deprecated,
  it's now possible to use hooks without explicit creation.
* Related fixes and refactoring


johnd0e:
Screenshot_2019-07-29-11-26-34-880_org exarhteam iitc_mobile

PR #212, opened 2019-06-27
sync: fix and update

modos189:
Fixed synchronization errors. Updated log view. 

* Fix of the first synchronization

* Removed an unnecessary timer that duplicates a repeated connection attempt in case of a network error

* Change the display of the change log. Only the last message is shown or each synchronized file.

* Disable error output with alert()

изображение

### DIFF
--- a/plugins/sync.user.js
+++ b/plugins/sync.user.js
@@ -203,21 +203,22 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
     _this.initialized = true;
     _this.initializing = false;
     plugin.sync.logger.log('Data loaded: ' + _this.pluginName + '[' + _this.fieldName + ']');
-    if(callback) callback();
+    if(_this.callback) _this.callback();
     if(_this.initializedCallback) _this.initializedCallback(_this.pluginName, _this.fieldName);
   };
 
   // Stop the sync if any error occur and try to re-authorize
   handleError = function(e) {
-    var error_type = e.type;
+    var isNetworkError = e.type;
+    var errorMessage = (e.error !== undefined) ? e.error.message : e.result.error.message;
     
-    if(e.error.message === "A network error occurred, and the request could not be completed.") {
-      error_type = "network error";
+    if(errorMessage === "A network error occurred, and the request could not be completed.") {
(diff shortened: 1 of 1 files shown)


plugins/sync.user.js
```diff
@@ -2,7 +2,7 @@
 // @id             iitc-plugin-sync@xelio
 // @name           IITC plugin: Sync
 // @category       Misc
-// @version        0.3.1.@@DATETIMEVERSION@@
+// @version        0.4.0.@@DATETIMEVERSION@@
 // @description    [@@BUILDNAME@@-@@BUILDDATE@@] Sync data between clients via Google Drive API. Only syncs data from specific plugins (currently: Keys, Bookmarks). Sign in via the 'Sync' link. Data is synchronized every 3 minutes.
```
johnd0e: Should we add 'uniques' to description?

MysticJay:
sync does not work with Multi Project Extension unless it only updates the default namespace.
Otherwise It might overwrite the data of a different project.
Needs to be investigated.
Quick fix: ensure sync only handles the default namespace not alternate NS.
First thought: sync must save the name of the active namespace with the data.

MysticJay:
Tested and confirming that SYNC is always SYNCing the current namespace of the plugin.
This is in confflicht with MPE.
MPE is an extension for operators and planners extension, enbling IITC to handle different projects.
SYNC is primarily NOT an Operator extension but one for Agents that plan at home and then want to sync to their mobile device.

So target audience is different.

I propose to only sync the default namespace of a plugin. This will ensure that an operator does not accidently share a the wrong project and that the agent does not destroy a project in planning phase.

window.mpe.projects.<nameof plugin>.defaultKey is holding the value to be saved.

@modos189 please modify accordingly.

johnd0e:
We consider this together with other MPE-related changes.
But it is offtopic **here**.

MysticJay:
As I have looked at the code, I tend to disagree, that this is OffTopic here. Adding it to MPE's PR does not make sense either as the new code base for SYNC is not given.
Creating a new issue would be possible but then you have to take care the issue gets added to the next build (but it would not change anything to the number of things that need to be done).

PR #213, opened 2019-06-29
Scoreboard improved

angel93ayora:
Removed Diffs from table
Removed times from `Checkpoint overview`collapsible
Changed the bottom texts to be more readable.
Some code refactor and comment deletions.
[fix] remainig cps were calculated wrong

image

### DIFF
--- a/code/region_scoreboard.js
+++ b/code/region_scoreboard.js
@@ -13,7 +13,10 @@ var RegionScoreboard = (function() {
     this.median=[-1,-1,-1];
     this.MAX_CYCLES = 35;
 
+    // arrays where index is the RegionScore's cp number (starting at 1)
     this.checkpoints = [];
+    this.summatory = [];
+    this.diff = [];
 
     this.hasNoTopAgents = function() {
       return this.topAgents.length===0;
@@ -31,6 +34,14 @@ var RegionScoreboard = (function() {
       return this.checkpoints[cp];
     };
 
+    this.getCPSummatory = function(cp) {
+      return this.summatory[cp];
+    };
+
+    this.getCPDiff = function(cp) {
+      return this.diff[cp];
+    };
+
     this.getScoreMax = function(min_value) {
       var max = min_value || 0;
       for (var i=1; i<this.checkpoints.length; i++) {
@@ -40,6 +51,7 @@ var RegionScoreboard = (function() {
       return max;
     };
 
+    // TODO: remove references, sums are calculated on object construction
     this.getCPSum = function() {
       var sums=[0,0];
       for (var i=1; i<this.checkpoints.length; i++) {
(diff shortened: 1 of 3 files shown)


johnd0e:
Could you attach also screen of plugin before your modification, to let us compare them easily?

Update
And you've made this PR over wrong branch: use `scoreboard-improved` (not `test-builds`)

modos189:
Region scores in release: 
изображение

in last test build:
изображение

johnd0e:
> Removed Diffs from table

What were that 'Diffs'?

> Removed times from `Checkpoint overview`collapsible

Why?

> Changed the bottom texts to be more readable.

Here I agree.

angel93ayora:
On [this comment](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/167#issuecomment-487420614) you can see the diff columns on the table and the times inside the Checkpoint overview collapsible.

>>Removed times from Checkpoint overviewcollapsible

>Why?

Because the same info. is shown at the bottom of the dialog.

> in last test build:

i could insert the time column if desired, or maybe show the time when the CP was reached on a title/tooltip

total scores text is also removed, is the same score as the last checkpoint summatory, but it could be placed at the bottom of the dialog

johnd0e:
> diff columns

I understand now.
But let's compare with base branch.
You've added `Sum` column, right?

> Because the same info. is shown at the bottom of the dialog.

At the bottom there is time of _next) CP.
It is not the same as time column for every _passed_ CPs.

> or maybe show the time when the CP was reached on a title/tooltip

It would be hard to access on touch devices.

> is the same score as the last checkpoint summatory

This summatory column seems not very useful for me.
If you really need it - let it be optional.

angel93ayora:
> You've added Sum column, right?

yes

> It is not the same as time column for every passed CPs.

>> or maybe show the time when the CP was reached on a title/tooltip

> It would be hard to access on touch devices.

so the best solution will be to add a button to show/hide those columns (sum and time)

johnd0e:
> so the best solution will be to add a button to show/hide those columns (sum and time)

Time was always there, so I suppose button for sum is enough.

angel93ayora:
button added

time column
image

sum columns
image

other changes included in PR:
Changed time texts at bottom of the dialog
Added difference between totals at top of checkpoint overview collapsible
image

**edit: style improvements, on next commit**

johnd0e:
@angel93ayora 
If you are going to support this PR, then please rebase it on updated `master`, and fix target.

johnd0e:
@angel93ayora 

https://github.com/IITC-CE/ingress-intel-total-conversion/pull/213#issuecomment-547833208

johnd0e:
Some minor changes implemented in #318.

Closing this PR for now.

PR #221, opened 2019-07-10
Multi project extension

MysticJay:
This is the first and IMHO most important plugin ZasoGD has created.
The script itself is unchanged compared to the original file.
Headers and footers are changed (not sure about the `@match` line)

Zaso's startup wrapper has been commented out, although I do not yet understand why.

The script has been tested in IITC-CE - Menu item load.


plugins/multi-projects-extension.user.js
```diff
@@ -0,0 +1,544 @@
+// ==UserScript==
+// @id             iitc-plugin-multi-projects-extension@Zaso
+// @name           IITC plugin: Multi Projects Extension
+// @category       Control
+// @version        0.0.6.@@DATETIMEVERSION@@
+// @namespace      http://www.giacintogarcea.com/ingress/items/
```
modos189: Please remove lines @namespace, @updateURL, @downloadURL, @match and @grand

plugins/multi-projects-extension.user.js
```diff
@@ -0,0 +1,544 @@
...
+// @updateURL      http://www.giacintogarcea.com/ingress/iitc/multi-projects-extension-by-zaso.meta.js
+// @downloadURL    http://www.giacintogarcea.com/ingress/iitc/multi-projects-extension-by-zaso.user.js
+// @description    [@@BUILDNAME@@-@@BUILDDATE@@] Create separated projects in some plugins.
+// @match          https://intel.ingress.com/*
+// @grant          none
+@@METAINFO@@
+// ==/UserScript==
+@@PLUGINSTART@@
+// function wrapper(){
```
modos189: The wrapper is added when building the project, please remove these comments

plugins/multi-projects-extension.user.js
```diff
@@ -0,0 +1,544 @@
...
+		window.plugin.mpe.setupCSS();
+		window.plugin.mpe.ui.addControl();
+		window.plugin.mpe.ui.appendContainerInSidebar();
+	}
+
+// PLUGIN END //////////////////////////////////////////////////////////
+
+@@PLUGINEND@@
+
+// if(window.iitcLoaded && typeof setup === 'function'){
```
modos189: and this comments

MysticJay:
IMHO ready to merge

plugins/multi-projects-extension.user.js
```diff
@@ -0,0 +1,544 @@
...
+// ==/UserScript==
+@@PLUGINSTART@@
+// function wrapper(){
+// // ensure plugin framework is there, even if iitc is not yet loaded
+// if(typeof window.plugin !== 'function') window.plugin = function(){};
+
+// PLUGIN START ////////////////////////////////////////////////////////
+
+	// use own namespace for plugin
+	window.plugin.mpe = function() {};
```
modos189:
```suggestion
window.plugin.mpe = function() {};
```

modos189: Reduce the indent for all code.

Everything else is fine.

modos189:
Contents of `Multi-Projects-Extension.doc` file please place as comment in multi-projects-extension.user.js file, and delete Multi-Projects-Extension.doc

MysticJay:
Does IITC-CE work with Block-Comments? /* and */?

modos189:
You can use multiline comments. Yeah, like you just did.

We have documentation (in the doc folder) but I don't like it very much, it should be generated based on comments in the code.

modos189:
Pay attention to my two remaining comments https://github.com/IITC-CE/ingress-intel-total-conversion/pull/221/files

modos189 (dismissed):

modos189:
Thanks, this PR will be merged into a new release

johnd0e:
> Thanks, this PR will be merged into a new release

I'd rather think about include support of this in IITC core.
May be in some other form (need to investigate functionality deeper).

MysticJay:
To test the functionality:
1) import Multi-Project-Extension.user.js to externals
2) import DRAW2-beta.user.js to externals
3) disable regular DRAW-Tools
4) enable MPE and Draw2-beta
Create a drawing
from MultiProjects Menu click on the "+" and create a new project. 
Existing drawing should be cleared.
create a new draw.
Use the dialogue to switch between drawings
Use the settings dialogue to add the selection menu to the MainMenu.

MysticJay:
To test:
disable standard Bookmarks-Plugin,
import bkmks2.beta to externals

johnd0e:
1. I do not get why you need separate plugin (draw2). You may just extend single draw-tools (in runtime)

2. But if you still need draw2 then please make your changes reviewable:
   - copy `draw-tools.user.js` to `draw-tools2.user.js`
   - make your changes in separate commit(s)

MysticJay:
Bare with me, I am new to GIT, so it seemed to be safer to create a separate draw than to fiddle around in the original draw. At least this way does not break the existing tools and existing extensions to it. When I started porting Zaso's changes it was not clear to me how deep the changes would be.

The changes can easily be followed by just renaming DRAW2-beta.user.js to draw-tools.user.js
Also the contributions should show the changes step by step, as the first contribution of the file was the original draw-tools (just a copy)

Draw2 and Bkmarks2 is "only" a Draft to show how MPE is working. Due to the same namespace both use with the old Versions, the data is completely compatible. If you find a problem with MPE it is easy to switch back to the old plugins without loosing anything.

Also there are many other proposals that need to be reviewed and I think it is easier to add these to a forked plugin and finally just replace the old plugin than to daily send in new contributions.

As you wrote earlier we should rather think of adding MPE to the code as "Projects" or "Project-Manager".

johnd0e:
> Also the contributions should show the changes step by step, as the first contribution of the file was the original draw-tools (just a copy)

Yeah, but for `draw-tools` I see only single commit https://github.com/IITC-CE/ingress-intel-total-conversion/pull/221/commits/9d89089e8f3b14fada5be8e1e25543940d69297a

Better if there were (1) initial (2) changes for MPE

> Also there are many other proposals that need to be reviewed and I think it is easier to add these to a forked plugin

We can play with forked plugins, but I do not think that we should release them as they are. It'd be confusing.

> As you wrote earlier we should rather think of adding MPE to the code as "Projects" or "Project-Manager".

Yes. It is general principle.
But there may be different implementations of Multi-Projects idea.
I need time to check it.

MysticJay:
Changes applied on original files. removed development (beta) files.

johnd0e:
Ok. I will review proposed changes ASAP (but do not expect it to be fast, definitely not earlier than 0.30 release).

johnd0e:
Some general thoughts:

MPE allows to group Draws/Bookmarks by related project.
What if we need some particular set of Draws or Bookmarks in several projects?
Oops. We can't.

I suppose we could improve project idea.
E.g.:
Bookmarks already has folders. So we might just select existent folders to add them into project.

plugins/font-awesome.user.js
```diff
@@ -0,0 +1,30 @@
+// ==UserScript==
+// @id             iitc-plugin-font-awesome@Zaso
+// @name           IITC plugin: Font Awesome
```
johnd0e, line 3: If we include web font - we should include it in core, to make it available out-of-the-box.
(and it is still unclear if should include - #236)

plugins/font-awesome.user.js
```diff
@@ -0,0 +1,30 @@
...
+
+// PLUGIN START ////////////////////////////////////////////////////////
+
+window.plugin.faIcon = {};
+
+var setup = function(){
+	$('<link>')
+		.prop('type', 'text/css')
+		.prop('rel', 'stylesheet')
+		.prop('href', 'https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css')
```
johnd0e, line 19: Why use 4.7 instead of current (5.10.1)?
https://fontawesome.com/

MysticJay: No objection to use 5.10.1.
Just not tested it.

MysticJay:
MPE is about replacing the complete namespace of a plugin. Sharing parts between different namespaces would require them to be in an additional namespace.
Whatever I do not think this is needed Copy paste between different projects is possible and helps to avoid accidental sharing of confidential Information between projects.
I highly recommend to not share anything between projects.

johnd0e:
I still don't get what do you mean. Explain please.

MysticJay:
You asked about sharing parts of draw or bookmarks between projects.

MysticJay:
What about moving this to the core?

johnd0e:
It is planned

PR #229, opened 2019-07-30
Refactor build system, simplify templates and escape macros in code

### LINKED ISSUE #50: @@-substitutions break validity of plugins' sources

### LINKED ISSUE #99: Scripts versioning

johnd0e:
- Refactor build system into several modules (can be used as cli utilities). Implement `watch` mode to auto-rebuild on source changes.
- Escape macros in code in order to keep js-validity, fix #50.
  Use template from settings instead of url harcoding (closes #150).
- Simplify userscripts source template, rename plugin sources:  `*.user.js` -> `*.js`, rearrange source directories.
- Remove timestamp component from `version` of Release scripts (closes #99).
- Additional dev tools: `web_meta_gen.py`, `web_server_local.py`, `tampermonkey_stubs_gen.py`.


johnd0e:
.

build.py
```diff
@@ -20,19 +20,18 @@
 # load settings file
 from buildsettings import buildSettings
 
+defaultBuild = None
 # load option local settings file
-try:
+if os.path.isfile('./localbuildsettings.py'):
```
modos189: If file is missing, there will be an exception. Not great, not terrible.

BTW, we can use:
```
# load option local settings file and default build
try:
    from localbuildsettings import buildSettings as localBuildSettings, defaultBuild
    buildSettings.update(localBuildSettings)
except ImportError:
    localBuildSettings = {}
    defaultBuild = None
```

johnd0e:
I am also going to enhance build script to be able to build single plugin (even 3rd-party).

code/boot.js
```diff
@@ -829,23 +829,23 @@ window.registerMarkerForOMS = function(marker) {
 }
 
 try {
-@@INCLUDERAW:external/autolink-min.js@@
+@include_raw:external/autolink-min.js@
```
modos189: Maybe we should add quotes, make a string?

plugins/debug-raw-portal-data.user.js
```diff
@@ -1,15 +1,8 @@
-// ==UserScript==
 // @id             iitc-plugin-raw-portal-data
```
modos189: Did you leave your id here on purpose, or did you forget to remove it?

plugins/periodic-refresh.user.js
```diff
@@ -1,15 +1,8 @@
-// ==UserScript==
 // @id             iitc-plugin-periodic-refresh
```
modos189: This is probably where you forgot to remove ID

total-conversion-build.js
```diff
@@ -1,13 +1,10 @@
...
-// @id             ingress-intel-total-conversion@jonatkins
+// @author         jonatkins
 // @name           IITC: Ingress intel map total conversion
-// @version        0.29.1.@@DATETIMEVERSION@@
-// @description    [@@BUILDNAME@@-@@BUILDDATE@@] Total conversion for the ingress intel map.
-@@METAINFO@@
+// @version        0.29.1
+// @description    Total conversion for the ingress intel map.
 // @run-at         document-end
-// ==/UserScript==
```
modos189: BTW, maybe we should move this file to `code` folder?

plugins/fix-china-map-offset.user.js
```diff
@@ -1,15 +1,8 @@
-﻿// ==UserScript==
-// @id             iitc-plugin-fix-china-map-offset
+﻿// @id             iitc-plugin-fix-china-map-offset
```
modos189: again

build_plugin.py
```diff
@@ -0,0 +1,182 @@
...
+    meta, is_plugin = fill_meta(meta, name, dist_path)
+    settings.plugin_id = name
+
+    filename = os.path.join(out_dir, name) if out_dir else name
+    if settings.update_file == '.meta.js':
+        io.open(filename + settings.update_file, 'w', encoding='utf8').write(meta)
+
+    script = re.sub(r"'@bundle_code@';", bundle_code, script)
+    wrapper = import_module(settings.plugin_wrapper)
+    template = r"'@(\w+)(?::([0-9a-zA-Z_./-]+))?@'"
```
modos189: Ooh. Uh...)
At least add a comment, please, that it's for finding `@type:string@` tags

build_plugin.py
```diff
@@ -0,0 +1,182 @@
...
+        if settings.update_file in {'.user.js', '.meta.js'}:
+            append_line('updateURL', path + settings.update_file)
+        append_line('downloadURL', path + '.user.js')
+    append_line('id', name)
+    append_line('namespace', settings.namespace)
+    if keys.isdisjoint({'match', 'include'}):
+        append_line('match', settings.match)
+    append_line('grant', 'none')
+    meta.append('// ==/UserScript==\n')
+    return '\n'.join(meta), is_plugin
```
modos189: Please add blank lines, everything is now merging into a solid block

plugins/basemap-gaode.user.js
```diff
@@ -1,15 +1,8 @@
-﻿// ==UserScript==
-// @id             iitc-plugin-basemap-gaode
+﻿// @id             iitc-plugin-basemap-gaode
```
modos189: id?

plugins/basemap-kartverket.user.js
```diff
@@ -1,15 +1,8 @@
-﻿// ==UserScript==
-// @id             iitc-plugin-basemap-kartverket
+﻿// @id             iitc-plugin-basemap-kartverket
```
modos189: id?

plugins/link-show-direction.user.js
```diff
@@ -1,15 +1,8 @@
-// ==UserScript==
 // @id             iitc-plugin-link-show-direction
```
modos189: id?

modos189 (dismissed):
It's pretty cool on the whole. I left a few questions in the comments

plugins/bookmarks.css
```diff
@@ -280,7 +280,7 @@
 	cursor:pointer;
 }
 #bkmrksTrigger, .bkmrksStar span, .portal-list-bookmark span {
-	background-image:url(@@INCLUDEIMAGE:images/bookmarks-img.png@@);
+	background-image:url(images/bookmarks-img.png);
```
modos189, line 283: Should we use `@include_img` here?

modos189 (changes requested):
Good job!

plugins/bookmarks.css
johnd0e: No, that is handled in another way.
Urls in CSS are easy to extract with regexp, so no need for special templates.  

See corresponding change in `bookmarks.js`.

modos189: Ok

modos189 (approved):

PR #233, opened 2019-08-03
Portal info: display coordinates without angled brackets

Nipheon:
If Firefox it was hard to select coordinates (for copy), as doubleclick selected brackets too.


johnd0e:
> Allows copy and pasting the raw portal coordinates into other routing apps.
> --
> In Portal View -> Map Links the coordinates are shown enclosed in angled brackets. I have no idea why and it's annoying because I either have to 
> * carefully drag-select only the coordinates
> * triple click the line, copy, paste and then remove the brackets before any map software that I know can process them
>
>  Because fielding ops mean I have to copy and paste one or two dozen coordinates this is really annoying.

Currently doubleclick does exactly what expected: select part inside brackets (Chrome).

And we can improve it slightly more: select on single click (only needed part).

But I suppose those brackets are not needed anyway, so I tend to accept this.

Nipheon:
Interesting, in Firefox a doubleclick selects the brackets too (which is why it's so annoying for me).

johnd0e:
BTW, there are plugins that place coordinates directly to portal info (so no need to use **Map links** for that).

Nipheon:
I correct myself, a double click selects the current number up to the prevoius/next period or comma (totally useless), a triple click selects everything including the brackets.

Oh, what are these plugins? I was thinking about doing this feature myself, but if that already exists...

johnd0e:
> Oh, what are these plugins?

Google it.
I've found https://github.com/vrabcak/iitc-portal-coords

Issue #52, opened 2019-01-13
Modular code design


johnd0e:
(Well, I think it's rather our goal manifest than an issue)

Currently IITC code is divided by source files, but still not organized well enough, as different parts are not encapsulated, and all inter-dependencies are not obvious.

All IITC-specific objects now resides in global `window` namespace, polluting it with multiple rather chaotic entities, and it's often hard to identify where they come from (which source file, or may be even from stock intel?).

Even more: namespace is now heavily polluted with IITC's private functions, which are never meant to be exposed (e.g. because it's impossible to reuse them).

So I propose:
- [ ] Hide all private stuff, expose only that we are really in need.
- [ ] Transform each source file into independent module with own namespace.
Use IIFE everywhere (like in `entity_decode.js`, `portal_detail.js`).
- [ ] Put all IITC objects into single `IITC` namespace object.

I believe that IITC own code should be exemplary, in order to facilitate clean and easy plugin writing for the whole community.
(Here is closely related issue: #46)

modos189:
I propose to maintain backward compatibility for now. And then, when the majority will use our project, make more global changes (preferably even with a script that automatically updates third-party userscripts). =)

johnd0e:
You are right, we cannot break compatibility now.
But we could change code in small portions, leaving compatibility stubs for plugins.

Anyway, it is long-term issue.

PR #234, opened 2019-08-06
Isolate every core module in separate IIFE

johnd0e:
this is the 1st step to isolate modules without applying namespace to every single private function (ough!)

Ref: #52


johnd0e:
Next steps will be:

1. Refactoring of every core module, carefully picking which of properties/methods/functions should be exposed, and which should become private.
    Perhaps, in result we'll get even different modules.

2. All standard plugins also need to be refactored, as currently they are full of bad coding patterns, thus - unable to serve as good examples for new plugins.

3. Stop abusing `window` for sharing functions, in favor of some modern modules system.
    This let us to apply some modern build system (instead of current custom python script)

4. Rethink plugins API from scratch.
    Again, we should limit `window` usage. Better if plugin get api object as argument of `setup`.
    Still, we can't break existent plugins, so there should be additional compatibility layer for a while.

5. ???

Also I'd like to decrease jQuery-dependancy, as it is not so helpful for modern browsers.

johnd0e:
@johnd0e .

build.py
```diff
@@ -133,8 +133,18 @@ def loaderSVG(var):
     return 'data:svg+xml;utf8,' + loaderString(var)
 
 
+def wrapInIIFE(fn):
+    str = readfile(fn)
```
modos189: `str` is reserved word, it is better to choose another variable name

modos189 (dismissed):

PR #235, opened 2019-08-06
Stop console 'spam' using customizable logger

### LINKED ISSUE #26: Debug output: make conditional
iitc writes a lot of debug info to console log, but it's not useful for general use, and only adds 'noice' to log.
It would be more useful if such thing were wrapped in something like this:

```
function window.debug.log (info) {
  if (window.debug.ENABLE_LOG) {
    console.log (info)
  }
}
```
Condition could be more complex:

```
function window.debug.log (level,info) {
  if (debug.ENABLE_LOG.ALL or debug.ENABLE_LOG[level]) {
    console.log (info)
  }
}
```

In order to control the condition ENABLE_LOG we could use additional plugin.

johnd0e:
All console functions calls in core modules are proxified with [ulog].
Fix #26: now by default only warnings and errors come to console.

To see more details - logging level should be changed (it's also possible to filter output by specific modules).
There are several ways to control logging, see [docs] for more info.

[ulog]: https://github.com/Download/ulog
[docs]: https://ulog.js.org/


johnd0e:
...e.g. it can be controlled by query string: https://intel.ingress.com/intel?log=debug
or filter by module: https://intel.ingress.com/intel?debug=map_data_render,boot
Notes:
- It is proof-of-concept, so in final version ulog may be replaced with some other logger, if it will suit our needs better.
- This PR is based on #234 in order to:
  - isolate modules to let them use own `log` function for each module (file-based).
    And in each module can add more granularity, defining more loggers for every specific purpose;
  - automate loggers instantiation: it is part of build template, so modules' sources require no changes.
- Todo: versions.md, attribution.md, plugins API.

PR #237, opened 2019-08-07
Rewrite code use class

HubertZhang:


johnd0e:
artifact.js

1. Why don't you use [methods syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions)?

2. Why not put `setup` actions into constructor?

3. And the main question: do we ever need `class` here?
   I suppose we don't, as we need 1 instance only.

So I propose to refactor this into plain module, which will expose only necessary functions.

johnd0e:
3. I mean that instead of class we could use plain module
```artifact.js
var REFRESH_JITTER,REFRESH_SUCCESS; // ...

function setup () { /* ... */ }
function idleResume () { /* ... */ }
function handleSuccess (data) { /* ... */ }
function handleFailure (data) { /* ... */ }
function processData (data) { /* ... */ }
function clearData () { /* ... */ }
function processResult (portals) { /* ... */ }
function getArtifactTypes () { /* ... */ }
function isArtifact (type) { /* ... */ }
function getArtifactEntities () { /* ... */ }
function getInterestingPortals () { /* ... */ }
function isInterestingPortal (guid) { /* ... */ }
function getPortalData (guid,artifactId) { /* ... */ }
function updateLayer () { /* ... */ }
function showArtifactList () { /* ... */ }

window.artifact = { // this will be export some day
  setup, isInterestingPortal, getInterestingPortals, getArtifactEntities
}
```
Remember, that whole content of module is put into IIFE
```
(function () {
  // module content
})()
```
```

**Comment by johnd0e** [comment]
I like this rewrite.
We also need to get rid of unneeded jquery (like `$.each`), and refactor in general.

And carefully limit which of functions need to be exposed.
We should think also what internal functions and constants may be useful to make customizable (and expose them too).

File: code/chat.js [hunk 1]
```diff
@@ -888,4 +888,30 @@ function postMsg() {
...
\ No newline at end of file
+}
+
+window.chat = {
+  setup,
+  show,
+  backgroundChannelData,
+  toggle,
+  getChatPortalName,
+  get _requestAlertsRunning() {
```
HubertZhang: @johnd0e could these getters work properly?

johnd0e: Why not?
(I see massive refactoring, so all this needs to be tested)

johnd0e:
@HubertZhang 
I have to remind that we still use ES5.1 in this project.
And we can't apply this PR until our build system upgraded to use transpiler.

For now I see this PR as _proof-of-concept_.

But: most of your changes are good even for ES5.1, if we stick to `var` and avoid (for now) _arrow functions_ (and some other cool features).
So may be do your changes in steps?
1. Refactor, keeping ES5.1
2. Merge
3. When we get transpiler in our build chain - do rest (minimal) changes for ES2015+?

P.S.
Our current code base does not work even in `'use strict'` mode

johnd0e:
@johnd0e So question is:
Do you want to merge most of your changes ASAP? In nearest release?

johnd0e:
@johnd0e 👆

johnd0e:
Here is something unrelated now

PR #238, opened 2019-08-08
Move wrapper template to separate file

johnd0e:
- Default template separated to `pluginwrapper.py`
- Custom wrapper can be set in `localbuildsettings.py`
- Same wrapper used for main script.
- Misc (un)related changes


.gitignore
```diff
@@ -1,5 +1,6 @@
 build
 localbuildsettings.py
+*custom.py
```
johnd0e: May be use some other template?
E.g. `local*.py`?

PR #239, opened 2019-08-16
IITC-Mobile: fix error that made it difficult to pick plugin file

modos189:
It was impossible to select a file in a standard android file picker

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
@@ -391,7 +391,7 @@ public class IITC_FileManager {
 
             // create the chooser Intent
             final Intent target = new Intent(Intent.ACTION_GET_CONTENT)
-                    .setType("text/*")
+                    .setType("*/*")
                     .addCategory(Intent.CATEGORY_OPENABLE);
             final IITC_Mobile iitc = (IITC_Mobile) mActivity;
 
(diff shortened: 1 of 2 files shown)


johnd0e:
Here was only one error — wrong type `file/*`.
And I am not sure that we should change it to generic `application/*`, as filechooser filtering feature is convenient — just set proper data type.

As for 2nd change — in function `FileRequest` we may need customizable type filtering, as that function is (indirectly) available for user scripts.
May be we need data type as additional argument in this function.

mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/PluginPreferenceActivity.java
```diff
@@ -145,7 +145,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                     // create the chooser Intent
                     final Intent target = new Intent(Intent.ACTION_GET_CONTENT);
                     // iitcm only parses *.user.js scripts
-                    target.setType("file/*");
+                    target.setType("application/*");
```
johnd0e:
```suggestion
                    target.setType("application/javascript");
```

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
```diff
@@ -391,7 +391,7 @@ private FileRequest(final Uri uri) {
 
             // create the chooser Intent
             final Intent target = new Intent(Intent.ACTION_GET_CONTENT)
-                    .setType("text/*")
+                    .setType("application/*")
```
johnd0e:
```suggestion
                    .setType("*/*")
```

PR #248, opened 2019-09-04
Consistent errors throwing

### LINKED ISSUE #191: Inconsistent errors throwing
1. `throw 'string'`
2. `throw new Error 'string'`
3. `throw {error: 'string'}`

Mostly 1st used.
2nd is better, but mobile debug console is able to handle plain strings only.

I suppose we should fix console.

johnd0e:
Now using `throw new Error('message')`
fix #191

Note:
For easy merging, some parts are left for #157 (utils_misc.js), #164 (main.js), #205 (hooks.js)

### DIFF
--- a/code/chat.js
+++ b/code/chat.js
@@ -54,7 +54,9 @@ window.chat.handleTabCompletion = function() {
 
 window.chat._oldBBox = null;
 window.chat.genPostData = function(channel, storageHash, getOlderMsgs) {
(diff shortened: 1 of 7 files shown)


angel93ayora:
what do you think about creating an Error method at iitc core, so you can send there the error, log it with the chosen logger from #235, and maybe do some other stuff if needed

johnd0e:
Could you provide sample code demonstrating any advantage of such function?

Issue #246, opened 2019-08-30
[New Feature] Display actual area (proxy for MU) for polygons.


Athanasius:
This will probably be best implemented as a direct part of DrawTools.

When trying to judge the exact portals to use for pre-planned fields it would be good to know the area covered by a given Polygon (easier to implement than "click on map, check for any triangle there made by DrawTools lines...").  Perhaps a popup/hover when editing a Polygon?

Keep in mind that for larger polygons proper "on the surface of a sphere" calculations will be necessary, and it's easiest to just do this anyway given it would be working off of Lat/Long co-ordinates of the corners.

If I find time I'll investigate and may provide a PR.

johnd0e:
May be better in layer-count?

Athanasius:
Layer Count operates only on existing fields, not DrawTools polygons, doesn't it ?

I'll investigate the best mechanism.  I'm specifically thinking of this sequence:

1. Draw the largest field
2. See area count
3. Adjust one portal to another candidate
4. See new area count and compare mentally.

Rather than the total area by all draw polygons (which also might not take fly links into account).

So, in theory, this just means hooking into when a polygon has just been completed, or an edit on it finished, to display the area.  I may well look at how Layer Count does its display for inspiration, although plugin-order can sometimes mean that displays in an inconvenient location (I'll see about opening another Issue about that). 

I'll work towards this for now, just to see how well it works, then could consider applying the same code to Layer Count as well, but remember that Layer Count is on *existing* fields, not on DrawTools polygons.

johnd0e:
> Layer Count operates only on existing fields, not DrawTools polygons, doesn't it ?

Drawn polygons too.



> but remember that Layer Count is on _existing_ fields, not on DrawTools polygons.

You better try

Athanasius:
It would help if Layer Count was actually working (Firefox and Linux), which prevented me making sure.  I'll go file an issue with the error.

johnd0e:
@Athanasius Make sure with test build

Athanasius:
With Layer Count working on the latest test build I can see that it does indeed also work on Polygons, so adding area count to that output would certainly be doable and useful to some.

But the extra clicks of drawing, using layer count, edit polygon, use layer count again just seems awkward, so I'll see about either an option to Layer Count to have a permanent display, or the separate plugin as well.

johnd0e:
> an option to Layer Count to have a permanent display

Explain how it might work

> or the separate plugin

[...]

Issue #247, opened 2019-08-31
[Layer Count] - Error and doesn't work on Firefox 68.0.2 on Linux


Athanasius:
// @version        0.1.0.20190315.122355
// @description    [release-2019-03-15-122355] Allow users to count nested fields
// @updateURL      https://iitc.modos189.ru/build/release/plugins/layer-count.meta.js
// @downloadURL    https://iitc.modos189.ru/build/release/plugins/layer-count.user.js
// @namespace      https://github.com/IITC-CE/ingress-intel-total-conversion

Nothing appears to happen when pressing the Layer Count icon/button.  Firefox Console shows:

TypeError: el is undefined line 25310 > scriptElement:3724:6
    removeClass https://intel.ingress.com/ line 25310 > scriptElement:3724
    onBtnClick https://intel.ingress.com/ line 235 > scriptElement:44
    eachLayer https://intel.ingress.com/ line 25310 > scriptElement:8260
    onBtnClick https://intel.ingress.com/ line 235 > scriptElement:39

Which doesn't match up with anything in Issue #149

Athanasius:
And yes, I'll sort out nuking all data and trying latest test build....

johnd0e:
@Athanasius https://iitc.modos189.ru/test_builds.html

Athanasius:
Yup, this works on the test build.  I've not been keeping track, do you have a release cadence/plan ?

PR #254, opened 2019-09-12
Layer count area

Athanasius:
Implements showing the area of selected existing fields or drawn triangles.  Addresses #246 

NB: Assumes a drawn polygon is a triangle and only considers the first three points.  To generalise this to any polygon would be a lot more work (either splitting it up into triangles and summing them, or doing a 'line scan' different method).

### DIFF
--- a/plugins/layer-count.user.js
+++ b/plugins/layer-count.user.js
@@ -20,30 +20,9 @@ plugin.layerCount.onBtnClick = function(ev) {
 		layer = plugin.layerCount.layer;
 
 	if(btn.classList.contains("active")) {
-		if(window.plugin.drawTools !== undefined) {
-			window.plugin.drawTools.drawnItems.eachLayer(function(layer) {
-				if (layer instanceof L.GeodesicPolygon) {
-					L.DomUtil.addClass(layer._path, "leaflet-clickable");
-					layer._path.setAttribute("pointer-events", layer.options.pointerEventsBackup);
-					layer.options.pointerEvents = layer.options.pointerEventsBackup;
-					layer.options.interactive = true;
-				}
-			});
-		}
 		map.off("click", plugin.layerCount.calculate);
 		btn.classList.remove("active");
 	} else {
-		if(window.plugin.drawTools !== undefined) {
-			window.plugin.drawTools.drawnItems.eachLayer(function(layer) {
-				if (layer instanceof L.GeodesicPolygon) {
-					layer.options.pointerEventsBackup = layer.options.pointerEvents;
-					layer.options.pointerEvents = null;
-					layer.options.interactive = false;
-					L.DomUtil.removeClass(layer._path, "leaflet-clickable");
-					layer._path.setAttribute("pointer-events", "none");
-				}
-			});
-		}
 		map.on("click", plugin.layerCount.calculate);
 		btn.classList.add("active");
 		setTimeout(function(){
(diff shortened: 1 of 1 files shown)


johnd0e:
Please rebase your additions on `master` branch.
And it'd be nice to include here some screenshot of plugin in work.

MysticJay:
We do not want ANY polygon, we want triangles only! So just use those polygons that have exactly 3 nodes.

johnd0e:
> and only considers the first three points.

Wrong. All non-triangles should be rejected.

Athanasius:
So, show an area of 0 m^2 if a polygon had more than 3 points, or adjust the output to not even show an area string?

Keeping in mind that if there are multiple polygons at the point clicked, with some being triangles and others with 4+ points, it will still count the triangles and show their total area.

Athanasius:
I could make it explicit with s/Area/Triangles Area/ in the polygon output.

johnd0e:
I have no suggestion on this, as I never seen your plugin at work.
Consider this https://github.com/IITC-CE/ingress-intel-total-conversion/pull/254#issuecomment-530770226

Athanasius:
353cac732857768da0e1e1c774bfbd4a76635e55 so we only work on triangles (and explicitly state polygon area is only for triangles)
2867a628c7e74dfc7931f493e58be553c6a93310 comments out most of the debug console.log'ing, leaving just a few in case of bugs.
(Edit: I'd forgotten to commit and cited the wrong commit IDs)

Now I'll see if this will rebase cleanly on master.

Athanasius:
That's it rebased on master, but now it doesn't work because master doesn't work on Firefox 69.0 on Linux.  That's why I was basing this on test-builds in the first place.  See #247

Athanasius:
So, if you want to check this you'll need to use https://github.com/Athanasius/iitc-ce_ingress-intel-total-conversion/pull/new/layer-count-area_test-builds

Athanasius:
Working on some screenshots now.

johnd0e:
> That's it rebased on master, but now it doesn't work

To make it work can cherry-pick this commit https://github.com/johnd0e/ingress-intel-total-conversion/commit/55c52bcbf2f57a2d201443033379f748ea4463f0

Athanasius:
> Working on some screenshots now.

https://miggy.org/games/ingress/IITC/layer-count-area/

Hopefully the filenames are self-explanatory.  Yes, the red squares and arrows were "added in post" and aren't part of the changes.  They're there just to make it clear where I clicked (cursor wouldn't capture) for each output.

Athanasius:
> > That's it rebased on master, but now it doesn't work
> 
> To make it work can cherry-pick this commit [johnd0e@55c52bc](https://github.com/johnd0e/ingress-intel-total-conversion/commit/55c52bcbf2f57a2d201443033379f748ea4463f0)

Confirmed working with this commit, but I assume I shouldn't push that into my branch for this PR.

johnd0e:
Why not. It will merge clean.

Athanasius:
> Why not. It will merge clean.

Done.

johnd0e:
No, that will not merge clean. You should first cherry-pick https://github.com/johnd0e/ingress-intel-total-conversion/commit/55c52bcbf2f57a2d201443033379f748ea4463f0, than rebase your changes on top of it.

Athanasius:
> No, that will not merge clean. You should first cherry-pick [johnd0e@55c52bc](https://github.com/johnd0e/ingress-intel-total-conversion/commit/55c52bcbf2f57a2d201443033379f748ea4463f0), than rebase your changes on top of it.

Now acceptable ?

johnd0e:
Now we are at least able to review the PR.
(I cannot give any forward promise that we accept it)

PR #255, opened 2019-09-15
Refactoring in build.py

johnd0e:
Separated from #153

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/255)

### DIFF
--- a/build.py
+++ b/build.py
@@ -9,8 +9,6 @@ import sys
 import os
 import shutil
 import json
-import shelve
-import hashlib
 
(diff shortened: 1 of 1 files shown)


johnd0e:
Todo:
- do not use globals in `main()`, pass data as arguments.
  Here we need build name or complete settings hash.
- apk building should be separate function

johnd0e:
Closed in favor #229

Issue #258, opened 2019-09-20
FlyLinks some modify


durck:
I slightly modified the plugin. In toolbox, added a button that moves to the draw tools links from flyLinks

durck:
It is interesting for your?

johnd0e:
Hard to say from that brief description.
Could you show some code, or screenshots?

durck:
1. Enable flyLinks
![Enable plugin](https://user-images.githubusercontent.com/31553806/65335019-4b05e680-dbcc-11e9-890a-67ba58eb15b5.png)

2. When we set view where we want work with plugin push button "Drow FlyLinks"
![InkedPush Button](https://user-images.githubusercontent.com/31553806/65335294-d7180e00-dbcc-11e9-8793-147996dd81f3.jpg)

3. We can disable plugin. We have:
![dt](https://user-images.githubusercontent.com/31553806/65335379-00389e80-dbcd-11e9-8789-fe4334bec355.png)

durck:
[fly-links.user.zip](https://github.com/IITC-CE/ingress-intel-total-conversion/files/3636432/fly-links.user.zip)

durck:
is it interesting or no?

johnd0e:
@durck Sure it is.
Similar features were requested long time ago.

durck:
you are will add this feature in the plugin? or i can to do it?

johnd0e:
@durck I have not tried your implementation yet. So I do not know if we can add it 'as is'.

You could submit a PR. It'd be useful for further discussion.

PR #260, opened 2019-09-21
added upload to the drowtool

durck:


johnd0e:
Ok, I see now. Thank your for this concept.

Actually, it's not necessary to modify Fly-links plugin.
Your new function can be implemented independently, as separate plugin.
And instead of using drawtools' function `import` better copy entities directly from `window.plugin.flyLinks.linksLayerGroup` and `window.plugin.flyLinks.fieldsLayerGroup` to `window.plugin.drawTools.drawnItems`.

durck:
Ok, i will doing it. But where i can make request with new plugin?

johnd0e:
Anywhere.
In separate PR, or even in this one.

johnd0e:
Ref: #258

PR #267, opened 2019-10-25
Change default intel url (to match stock)

modos189:
Before this PR: our entry point was `intel.ingress.com/intel`.
Reason: that where stock intel redirects to when we try to use old deprecated link (www.ingress.com/intel).

But stock intel has bug there: 'sign out' link leads to https://intel.ingress.com/intel/logout, which seems broken (redirect back to logout even after re-login, over and over again).

As appeared, currently `intel.ingress.com/` is proper intel url (as linked from ingress.com).
It does not have sign-out issues.

So now we change base url in iitc and plugins.


johnd0e:
fixup: 3b34a63f714bbcf643f36f4bc1b1dffcdee12393

PR #268, opened 2019-10-28
Build IITC use webpack

HubertZhang:


McBen:
I really love this PR. 
It's still missing plugin support and old-build script should be changed too - especially for mobile build.

But sad to say that it won't work out-of-the-box anymore. 

After rebase and (add missing `number-is-nan` package) I still got a "resolve-dir" not found. ATM IDK how to solve this.

PR #279, opened 2019-11-25
Round shield mitigation to cope with lack of js float precision

appletalk:


code/portal_info.js
```diff
@@ -266,7 +266,7 @@ window.getPortalMitigationDetails = function(d,linkCount) {
   // mitigation is limited to 95% (as confirmed by Brandon Badger on G+)
   mitigation.total = Math.min(95, mitigation.shields+mitigation.links);
 
-  mitigation.excess = (mitigation.shields+mitigation.links) - mitigation.total;
+  mitigation.excess = Math.round(10 * ((mitigation.shields+mitigation.links) - mitigation.total)) / 10;
```
johnd0e, line 269:
```suggestion
  var excess = (mitigation.shields+mitigation.links) - mitigation.total;
  mitigation.excess = Math.round(10*excess) / 10;
```

And better write add more info to commit message, like this:

> JS float precision is not good, so without rounding we'd get `125.8 - 95 = 30.799999999999997`
  (More background info: https://floating-point-gui.de/)

johnd0e (changes requested):

johnd0e:
Applied in a56929308ad8f99d29baa0fa6cb46486f9cee97d

PR #286, opened 2019-12-08
Set portal permalinks to zoom level 17

5310:
After the common permalink generator refactor [!198](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/198) the portals permalinks are certainly neater and shorter, but they are much more zoomed out than before. This change adds the zoom level 17 argument back into the permalink, which just adds four characters to the URL itself, and has better usability.

### DIFF
--- a/core/code/utils_misc.js
+++ b/core/code/utils_misc.js
@@ -500,7 +500,10 @@ window.makePermalink = function (latlng, options) {
   }
   if (latlng) {
     if ('lat' in latlng) { latlng = [latlng.lat, latlng.lng]; }
-    args.push('pll='+latlng.join(','));
+    args.push(
+      'pll='+latlng.join(','),
+      'z=17'
+    );
   }
   var url = options.fullURL ? '@url_intel_base@' : '/';
   return url + '?' + args.join('&');


johnd0e:
What do you mean by "much more zoomed"?
Default zoom was indeed changed, but it is not related to #198, it's #281.

5310:
I meant much more zoomed _out_ actually. I thought the default zoom level was always 15, and after #198 the portal permalinks were using it (15) instead of 17.

johnd0e:
If you prefer zoom level 17 then just change constant. 
No need to patch sources, you can do it in plugin

perringaiden:
> I meant much more zoomed _out_ actually. I thought the default zoom level was always 15, and after #198 the portal permalinks were using it (15) instead of 17.

15 is the level at which all portals are visible, but prior to the links vs portals change, 17 used to be the level where uncaptured portals appeared, and people were used to the old view being zoomed to 17.

perringaiden:
Overall, *a* zoom level should be specified in the links, especially when providing them outside of IITC.  Whether that's 15 or 17 is largely a style choice.

Of note, not providing the "ll" value in a permalink also means that Stock may select the portal, without properly moving the map to display the portal.

johnd0e:
> Overall, a zoom level should be specified in the links, especially when providing them outside of IITC.

Have you ever tried?
When zoom not specified - stock intel itself assumes default value (15 or 17).

> Of note, not providing the "ll" value in a permalink also means that Stock may select the portal, without properly moving the map to display the portal.

Have you tried?
When `ll` is not specified, it is assumed equal to `pll` by stock intel.

johnd0e:
But I see another issue: sometimes we need EXACT permalink, including `ll` and `pll` (with values that differ).

We can get this in stock intel. But it is not possible in iitc (and never was).

johnd0e:
To summarize:
- it is not clear if we really need permalinks with explicit zoom
- if it will be found that we do - it should be implemented in different way (additional option of the function.)

PR #290, opened 2019-12-10
ITC-Mobile: option to use external browser for Google OAuth

johnd0e:
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/290)


le-jeu:
do you succeed to log with custom tabs ? I fail to get the redirection back to IITCm

Edit: ok, got it working on Intel login, but fails with plugin login (even with the button to get back to iitcm)

johnd0e:
> but fails with plugin login 

I had the same issue, and was going to forbid external login for sync.

johnd0e:
There are possible 2 different approaches
Either detect sync login by some url param (easy to implement for sync, but not cover other possible plugins), or automatically forbid external login after iitc is already booted.

PR #296, opened 2019-12-22
Prevent loop redirects on some older versions of WebView

modos189:
It was found out that Google authorization loops occur on some older versions of WebView. Therefore, when detecting an endless loop it is proposed to update Android WebView from Google Play

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewClient.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewClient.java
@@ -3,12 +3,15 @@ package org.exarhteam.iitc_mobile;
 import android.annotation.SuppressLint;
 import android.app.AlertDialog;
 import android.app.Dialog;
+import android.content.ActivityNotFoundException;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.net.Uri;
 import android.net.http.SslError;
+import android.os.Build;
 import android.os.Environment;
+import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
 import android.view.View;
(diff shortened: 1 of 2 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewClient.java
```diff
@@ -292,6 +309,14 @@ else if (uriHost.equals("accounts.google.com") ||
                  uriPath.contains("conflogin") ||
                  uriPath.contains("ServiceLogin")) {
             Log.d("Google login");
+
+            // Prevent loop redirects on some older versions of WebView
+            if (lastGoogleUrl.equals(url)) {
+                Log.d("Google login loop");
```
johnd0e:
```suggestion
                Log.d("Endless loop detected: Google login redirected to same URL:");
                Log.d(url);
```

mobile/app/src/main/res/values/strings.xml
```diff
@@ -207,5 +207,9 @@
         <p><b>Be careful:</b> Javascript from external sources may contain harmful code (spyware etc.)!
         <p>Are you sure you want to proceed?]]>
     </string>
+    <string name="webview_requires_update_top">Requires update</string>
```
johnd0e:
```suggestion
    <string name="webview_requires_update_top">Unable to login</string>
```

mobile/app/src/main/res/values/strings.xml
```diff
@@ -207,5 +207,9 @@
...
+    <string name="webview_requires_update_msg">"IITC uses external component Android System WebView to display web content. Please update it to continue."</string>
```
johnd0e:
```suggestion
    <string name="webview_requires_update_msg">"Google login continues redirecting to same URL. This is most likely caused by outdated Android System WebView component."</string>
```

mobile/app/src/main/res/values/strings.xml
```diff
@@ -207,5 +207,9 @@
...
+    <string name="open_google_play">Open Google Play</string>
```
johnd0e:
```suggestion
    <string name="open_google_play">Update WebView</string>
```

johnd0e:
Also, after updating WebView we need force [close](https://stackoverflow.com/questions/6330200/how-to-quit-android-application-programmatically) iitc.

johnd0e:
It seems that original issue is somehow fixed by #289.

PR #302, opened 2019-12-27
IITC-Mobile: fix google login

johnd0e:


modos189 (approved):
In the future (#290), it will be simplified, but for now, Ok

Issue #300, opened 2019-12-24
Android KitKat support


modos189:
Now, I was doing a test to make sure that Custom Tabs were disabled correctly on KitKat (because they are not supported there) and it turned out that IITC was not working at all.

![Screenshot_kitkat](https://user-images.githubusercontent.com/1785196/71426360-5ed45380-26b9-11ea-8cd0-fed420c35e33.png)
(related with https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/core/external/ulog.min.js)

Of course, we can take time to fix that bug, but is it worth it?

Pros of not supporting KitKat:

* Ability to work with modern WebView and less kludges.

* Support for modern JS standards, as for third-party plugins as well.

Minus of not supporting KitKat:

* According to Google Play we will lose 62 users.

It turns out that the number of pluses exceeds the number of minuses. And since there is no developer to provide support for KitKat, I suggest:

1. Raise minimum requirements in IITC-Mobile to Android 21 (Lollipop)

2. When adding new features, do not avoid using modern JavaScript features (e.g. const, let, async).

# Roadmap:

1. Update SDK min version to 21 (Lollipop) in release v0.30.1 https://github.com/IITC-CE/ingress-intel-total-conversion/pull/304

2. Cleaning up obsolete code specific to older versions of android in release v0.31.0

Korb:
Currently [the app page in F-Droid](https://f-droid.org/en/packages/org.exarhteam.iitc_mobile/) displays a warning when viewed on Google Android 14

> This app was built for an older version of Android and cannot be updated automatically.

During installation, another warning appears:

> This app was built for an older version of Android and doesn't include the latest privacy protections. Installing this app may put your device at risk. Learn more about Play Protect. [Install anyway]

I suggest you read the article "[Meet Google Play's target API level requirement](https://developer.android.com/google/play/requirements/target-sdk)", which is primarily intended for Google Play users, but F-Droid users will also have difficulty finding, installing, and updating the app in situations where OS security requirements conflict with the user app.

Also, I recommend keeping in mind [the changes coming in 2025](https://android-developers.googleblog.com/2024/10/android-sdk-release-update.html).

Suburbanno:
I really don't see much point in maintaining support for such old versions.

[...]

PR #304, opened 2019-12-30
Update SDK min version to 21 (Lollipop)

modos189:
In current form IITC is not working on android before lollipop. Given significant limitations in debugging and small audience of users of this version of Android, it was decided to increase the requirements for Android version. See https://github.com/IITC-CE/ingress-intel-total-conversion/issues/300


johnd0e:
Is there anyone with KitKat? We need to check if Google still allows to login into intel (we doubt).

PR #306, opened 2020-01-11
Plugins: fix handling of placeholder portals

### LINKED ISSUE #120: ap-stats: does not work at zoom <15
```
Available AP in this area:
Enlightened | NaN
Resistance | NaN
```

There should be more descriptive info, e.g. `Zoom closer to get all portals loaded`.

johnd0e:
Placeholder portals data lack of details expected by plugins.
Despite that details have disappeared years ago, some plugins still not completely adapted.

- [x] portals-list
- [x] portal-counts
- [x] ap-stats (fix #120)

### DIFF
--- a/plugins/ap-stats.js
+++ b/plugins/ap-stats.js
@@ -1,7 +1,7 @@
 // @author         Hollow011
 // @name           Available AP statistics
 // @category       Info
-// @version        0.4.1
+// @version        0.4.2
 // @description    Displays the per-team AP gains available in the current view.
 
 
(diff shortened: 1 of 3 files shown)


johnd0e:
todo: check if error in portal-counts is fixed: `Error: <rect> attribute height: Expected length, "NaN".`

johnd0e:
Btw, we have not defined exact 'placeholders' [meaning](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a844ea192dbd92bad7e3bbcb990f99a6899c6328/core/code/map_data_render.js#L251-L256).
They are portal entities usually created from indirect intel data (links, fields), and lacking most of the properties.

And what is most important: placeholders presence mean that current map zoom is not enough to retrieve all portals data, that makes some plugins processing useless (`portal-count`, `ap-stats`, `tidy/fly-links`, etc).

There are also other reasons why we may encounter placeholders in `window.portals` even on zoom >=15:
- off-screen portal
- some intel request is not finished or failed
- ???

As placeholders usually do not contain data required for correct processing - we may need to filter them out.
- And correct way to do it - rely on exact data presence (which may be different depending on task)
- Wrong way - unconditionally rely on some single general property (e.g. `!portal.title`, or `portal.team === TEAM_NONE && !portal.level`)

ATM placeholders provide such data: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a844ea192dbd92bad7e3bbcb990f99a6899c6328/core/code/map_data_render.js#L259-L268
Also, [level](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a844ea192dbd92bad7e3bbcb990f99a6899c6328/core/code/map_data_render.js#L319-L322) is set to `0`.

modos189 (approved):

PR #308, opened 2020-01-11
draw-tools: fix touch on iOS

johnd0e:
As it appeared, our attempt to fix Leaflet.draw [bug] in #175
is not good for some environments, e.g. it prevents proper touch
detecting on iOS (observed while drawing polygons)

[bug]: https://github.com/Leaflet/Leaflet.draw/issues/789

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/308)

### DIFF
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -573,7 +573,7 @@ window.plugin.drawTools.snapToPortals = function() {
   window.plugin.drawTools.save();
 }
 
-window.plugin.drawTools.drawstart = function (e) {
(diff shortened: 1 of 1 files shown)


modos189:
Does this enable or disable touch mode? The safari browser is also desktop (OSX).

johnd0e:
It just reduces our intervention into original Leaflet.draw code.
BTW, no matter desktop or mobile, as either can be equiped with touchscreen.

modos189 (approved):

PR #309, opened 2020-01-14
quick-fix for old android versions with pre-ES2015 webview

johnd0e:
Issue observed under x86 android kitkat/lollypop emulator:
as appeared, ulog module fails trying to redefine function name property.
https://github.com/Download/ulog/issues/24

So now we detect this case and then replace ulog with a stub.

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/309)


modos189:
It works

Issue #203, opened 2019-05-15
Fly links / Tidy links: ideas to improvements


johnd0e:
There is one idea which is discussed multiple times, and implemented in PR's:
we need the possibility to limit area (with draw-tools or anything), and exclude some portals.

But here I want to discuss other aspects of these plugins using.

johnd0e:
1. We need a possibility to freeze current state of drawn tidy/fly links

johnd0e:
1a. We do not need auto-redrawing of layers at all, better run draw on demand (button, hotkey, or toolbox link)
(related old issue: https://github.com/iitc-project/ingress-intel-total-conversion/issues/340#issuecomment-18795183)

johnd0e:
2. Export drawn layers to draw-tools (or GeoJSON).
See also #260

modos189:
3. Take into account already existing links #202

johnd0e:
> 1. Take into account already existing links #202

Well, but we already have issue for that)
Anyway, I would prefer to focus here on HOW we operate with such plugins in general.

Fly33:
About fly links:
1) 🤔 
2) It will be implemented someday )
3) Partially implemented in 0.4.0 through draw tools lines. (https://www.reddit.com/r/Ingress/comments/erfauj/suggesting_links_to_flylinks_plugin_coming_soon/) I think this is much enough.

johnd0e:
3. I do not feel it's enough, because saying 'existing links' I meant real intel links, not drawn.

johnd0e:
4. Facilities to filter portals (like limit to polygon and exclude portals with markers) should be shared, because they be useful for a lot of plugins. And it's not good to 'invent the wheel' in every such plugin.
  I am thinking about sort of 'portal filtering api' (will open separate issue about this).

Fly33:
> 
> 
>     1. I do not feel it's enough, because saying 'existing links' I meant real intel links, not drawn.

What if you don't need to pay attention to existing links? Should you go and destroy them?)

johnd0e:
@Fly33 Well, it's easy to destroy some enemy portals nearby.
But it's harder with own team portals. 
Distant crosses also can prevent effective linking.

I agree, nothing is impossible, but if I am planning operation that requires serious preparations - I'd rather apply Maxfield.

From the opposite side - fly-links is much easier for casual linking.

johnd0e:
2. See also #260, #258

[...]

PR #310, opened 2020-01-14
Fly-link update to 0.4.0

Fly33:
+ region selection with the polygon draw tool
+ exclusion of portals with the marker draw tool
+ steady drawing (portals is not excluded from the drawing if they are
moved out of boundaries)

### DIFF
--- a/plugins/fly-links.js
+++ b/plugins/fly-links.js
@@ -1,7 +1,7 @@
 // @author         fly
 // @name           Fly Links
 // @category       Draw
-// @version        0.2.1
+// @version        0.3.0
 // @description    Calculate how to link the portals to create the largest tidy set of nested fields. Enable from the layer chooser.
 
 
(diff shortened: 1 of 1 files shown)


modos189:
Looks great. I see no obstacles to merger after the corrective release that will be released the other day.

johnd0e:
@Fly33 
Could you comment ideas of #203?
Would it be hard to implement?

plugins/fly-links.js
```diff
@@ -221,32 +217,168 @@ window.plugin.flyLinks.updateLayer = function() {
...
+    }
+  }
+  
   $.each(edges, function(idx, edge) {
     drawLink(edge.a, edge.b, {
       color: '#FF0000',
       opacity: 1,
       weight: 1.5,
-      interactive: false,
+      clickable: false,
```
modos189:
```suggestion
      interactive: false,
```

plugins/fly-links.js
```diff
@@ -221,32 +217,168 @@ window.plugin.flyLinks.updateLayer = function() {
...
-  
+
   $.each(triangles, function(idx, triangle) {
     drawField(triangle.a, triangle.b, triangle.c, {
       stroke: false,
       fill: true,
       fillColor: '#FF0000',
       fillOpacity: 1 - Math.pow(0.85, triangle.depth),
-      interactive: false,
+      clickable: false,
```
modos189:
```suggestion
      interactive: false,
```

modos189 (changes requested):
New version of Leaflet uses the keyword *interactive*. Please do not return outdated *clickable*

modos189 (approved):

Issue #293, opened 2019-12-17
scale bar is hidden on mobile


johnd0e:
With any basemap except ingress/google scale bar get hidden under statusline.

* Complete solution: fix map container bounds, or `.leaflet-bottom` style, considering related issues #89, #151

* Just quick fix: set `plugin.scaleBar.mobileOptions.position` to `topleft`.

PR #311, opened 2020-01-14
Fix scale bar on mobile on non-google maps

modos189:
fix issue https://github.com/IITC-CE/ingress-intel-total-conversion/issues/293

screenshot

### DIFF
--- a/core/smartphone.css
+++ b/core/smartphone.css
@@ -102,6 +102,13 @@ body {
   margin-left: 5px !important;
 }
 
+.leaflet-control-container .leaflet-bottom {
+  display: flex;
+  flex-direction: column-reverse;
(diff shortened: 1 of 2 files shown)


johnd0e:
Haven't tested this myself. But if this will be appeared good solution - we should push it to [upstream](https://gitlab.com/IvanSanchez/Leaflet.GridLayer.GoogleMutant/-/merge_requests).

Also I'm not sure that the fix should be tied to `smartphone.css`.

johnd0e:
As appeared, upstream already has some solution merged: https://gitlab.com/IvanSanchez/Leaflet.GridLayer.GoogleMutant/-/merge_requests/49

So it's possible that current fix should be reverted, after updating #365.

Issue #236, opened 2019-08-07
Include icons font?


johnd0e:
#221 (besides other features) also includes [fontawesome](https://fontawesome.com/).

**Pros:**
For plugins developers it's very convenient to have some icons out-of-the-box

**Cons**
If we need only some icons, it is more efficient to embed only that icons.

So questions:
1. What is actual overhead of including whole font?
2. IITC-CE already contains (some) icons from https://materialdesignicons.com/.
What set should we prefer?

modos189:
We now have connected some material icons and can always expand their number. We can even integrate all the icons so that the plugin developers can use them when needed.

Since we use svg icons, it's no problem for us to add a new icon or remove unused ones.
If we use fontawesome, we will need to rebuild the entire bundle with fontawesome.

SVG allows us to use multicolor icons, which are easier to position on page, and do not cause problems when you force the browser to specify the font used by the browser. In addition, icons as a font are text, so browsers can (depending on the browser) apply different antialiases.

johnd0e:
What is important, imho
- icons should be available 'out-of-the-box', not 'on request'
- icons should be easy to use (fonts are easier here)
  Yeah, svg has own advantages, but who really cares?
- we should waste too much resources

About last point: I do not know yet how that fonts are implemented.
When we include css - do all icons loaded immediately, or only when actually used?
css itself is small, much smaller than set of svgs.

inexcode:
> Do all icons loaded immediately?

Font also has to be downloaded, with all icons.

In case with materialdesignicons.com, it's ~300 KB uncompressed with woff2, ~200 KB compressed from CDN. Old browsers would fall back to other formats, ~1 MB uncompressed.

johnd0e:
> Old browsers would fall back to other formats, ~1 MB uncompressed.

What exact browsers/versions do you mean?

inexcode:
> What exact browsers/versions do you mean?

https://caniuse.com/#feat=woff2

johnd0e:
> https://caniuse.com/#feat=woff2

Most important: Android 2.1 - 4.4.4 - not supported
Well, I suppose those versions are almost gone.

modos189:
I think, discussion should be divided into two questions:

1. Using svg or font material icons?

2. Connecting also fontawesome library?

modos189:
> Android 2.1 - 4.4.4 - not supported
>  Well, I suppose those versions are almost gone.

[...]

PR #312, opened 2020-01-16
Multi-Project-Extension

MysticJay:
After V0.30 was released the code, that was initially sumbmitted for PR#221 got corrupted.
Revamped the branch and fixed the code.


plugins/bookmarks.js
```diff
@@ -1,12 +1,17 @@
+// @id             Bookmarks
```
modos189: In our build system, if no ID is specified, the filename is used. Since the ID and filename match here, there is no point in specifying the ID

plugins/bookmarks.js
```diff
@@ -1,12 +1,17 @@
...
 // @author         ZasoGD
 // @name           Bookmarks for maps and portals
 // @category       Controls
-// @version        0.3.0
+// @version        0.4.0
 // @description    Save your favorite Maps and Portals and move the intel map with a click. Works with sync.
```
modos189:
```suggestion
 // @description    Save your favorite Maps and Portals and move the intel map with a click. Works with sync. With support Multi Projects Extension.
```

plugins/draw-tools.js
```diff
@@ -1,13 +1,21 @@
+// @id             draw-tools
 // @author         breunigs
 // @name           Draw tools
 // @category       Draw
-// @version        0.7.1
+// @version        0.8.0
 // @description    Allow drawing things onto the current map so you may plan your next move.
```
modos189:
```suggestion
 // @description    Allow drawing things onto the current map so you may plan your next move. With support Multi Projects Extension.
```

plugins/bookmarks.js
```diff
@@ -1,12 +1,17 @@
...
 
 
-/***********************************************************************
+// History ***********************************************
```
modos189: We're using git history, so please remove these lines

plugins/bookmarks.js
```diff
@@ -1303,4 +1352,11 @@
     if (window.plugin.portalslist) {
       window.plugin.bookmarks.setupPortalsList();
     }
+// Initilaize MPE-Support only if MPE-Module is available
```
modos189:
```suggestion
    // Initilaize MPE-Support only if MPE-Module is available
```

plugins/draw-tools.js
```diff
@@ -198,15 +206,14 @@ window.plugin.drawTools.save = function() {
...
-  localStorage['plugin-draw-tools-layer'] = JSON.stringify(data);
+  localStorage[window.plugin.drawTools.KEY_STORAGE] = JSON.stringify(data);
 
   console.log('draw-tools: saved to localStorage');
 }
 
 window.plugin.drawTools.load = function() {
   try {
-    var dataStr = localStorage['plugin-draw-tools-layer'];
+        var dataStr = localStorage[window.plugin.drawTools.KEY_STORAGE];
```
modos189:
```suggestion
    var dataStr = localStorage[window.plugin.drawTools.KEY_STORAGE];
```

plugins/draw-tools.js
```diff
@@ -363,10 +379,11 @@ window.plugin.drawTools.optCopy = function() {
         });
     }
 }
-
```
modos189, line 368: Apparently, a line was mistakenly deleted? It looks worse

plugins/draw-tools.js
```diff
@@ -658,11 +675,95 @@ window.plugin.drawTools.boot = function() {
...
+}
+
+
 function setup () {
-  loadExternals();
-  window.plugin.drawTools.boot();
+  loadExternals();                              // initialize leaflet
+  window.plugin.drawTools.boot();               // initialize drawtools
+  window.pluginCreateHook('pluginDrawTools');
+  window.plugin.drawTools.setupCSS();           // create additional CSS elements
```
modos189: If this function is performed not only when MPE is active but always, maybe you should transfer the CSS code to a separate CSS file?

modos189 (changes requested):

plugins/bookmarks.js
MysticJay: will be removed

plugins/bookmarks.js
MysticJay: will be added

plugins/bookmarks.js
MysticJay: will be removed

plugins/bookmarks.js
MysticJay: will be indented

plugins/draw-tools.js
MysticJay: Will be changed

plugins/draw-tools.js
MysticJay: Indentation will be removed

plugins/draw-tools.js
MysticJay: will be added

plugins/draw-tools.js
MysticJay: Need to check, where this code is needed. Might be slipped in from upcomming features in DrawToolsPlus.
If kept it should move to the other CSS code in DT.

MysticJay: The CSS changes are used to remove the GAP between drawitems and drawedit to safe space.
Maybe reintroduce at a different time.
line 759: setupCSS () removed.

modos189:
In the first commit I fixed the syntax, and in the second I removed the font-awesome. Why? In the task https://github.com/IITC-CE/ingress-intel-total-conversion/issues/236 we discussed the use of icons - they should be font or svg.
I'm in favor of using svg, but anyway - icons should not be in the form of a plugin, but in the core and become standard, otherwise only there two plugins will use icons and that will be the end of it.

Also note https://github.com/IITC-CE/ingress-intel-total-conversion/pull/312/files#diff-3073581708df23c3f1bd259521cb3021R8.
I replaced the code sample with a link to documentation - you just need to create this page and move the sample there

MysticJay:
Syntax changes are ok, while I prefer way more comments and explanations to each function. The dividers between the functions helped a lot to reverse engineer the code. But this is a different discussion.

FontAwesome: the current usage of SVGs already have negative side effects in some faction specific plugins, esp. if combined with "alternative text". But that is a different discussion too. 
FA is widely used in ZasoItems. I am fine keeping FA as a separate plugin with the other ZasoItems.

modos189 (approved):

PR #322, opened 2020-02-06
Close a tag after FAQ link

mvolfik:
Came across this when translating, e.g. German translation is wrong, as it has to place some word behind "FAQ", which would currently still work as a link.


modos189:
Thanks for PR, but this long text should be divided into separate short sentences

johnd0e:
Obsolete after #323

PR #331, opened 2020-02-16
Fix Modal Dialogs to be top-most

McBen:
When multiple dialog are open a "top most" dialog can be below other dialogs.

Dialogs are created with up counting z-index. They start with 9999 and
quickly reaches 10001.  Modal Dialog are always on z-index 10001 and so
interference with these if more the 2 dialogs are open.

### DIFF
--- a/core/style.css
+++ b/core/style.css
@@ -861,7 +861,7 @@ h3.title {
  */
 .ui-tooltip, .ui-dialog {
   position: absolute;
-  z-index: 9999;
+  z-index: 9500;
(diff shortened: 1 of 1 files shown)


johnd0e:
@McBen Thank you for recent PRs.
But I see (in all 3) there is extra merge commit.
Could you avoid it?

McBen:
> 
> 
> @McBen Thank you for recent PRs.
> But I see (in all 3) there is extra merge commit.
> Could you avoid it?

oops...fixed

johnd0e (approved):

Issue #122, opened 2019-02-27
Earth radius


johnd0e:
```
commit 522ef34f77c86c54cc7586a298d5a6f74f75aef4
Author: Jon Atkins <github@jonatkins.com>
Date:   Thu Mar 12 20:47:05 2015 +0000

    earth radius: change leaflet and leaflet-draw to use the radius value used by the s2 geometry library

diff --git a/external/L.Geodesic.js b/external/L.Geodesic.js
diff --git a/external/leaflet-src.js b/external/leaflet-src.js
diff --git a/external/leaflet.draw.js b/external/leaflet.draw.js
```

We need to investigate if this change was necessary.

_Originally posted by @johnd0e in https://github.com/IITC-CE/ingress-intel-total-conversion/pull/78#issuecomment-462323839_

johnd0e:
More related links:
- https://github.com/iitc-project/ingress-intel-total-conversion/issues/1236

johnd0e:
> diff --git a/external/L.Geodesic.js b/external/L.Geodesic.js

Radius value here does not have to be precise, so that change had no sense.

McBen:
S2-Earth-Radius is used in distance calculation.
You'll see the difference in draw-tools (or other plugins when they use LatLng.distanceTo)

Current used radius result im more accurate distances but Ingress uses the S2 value for link-distance calculation  - which is sometimes a important value.

PR #332, opened 2020-02-16
S2 coordinate system

McBen:
Reactivating the S2-Earth value will fix distance calculation.

Current calculation (shown when drawing a drawtool polyline) is of by ~0.2%

related: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/122

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -617,7 +617,6 @@ window.extendLeaflet = function() {
     return new L.DivIcon.ColoredSvg(color, options);
   };
 
-  /* !!This block is commented out as it's unclear if we really need this patch
 
   // See https://github.com/IITC-CE/ingress-intel-total-conversion/issues/122
 
@@ -628,9 +627,9 @@ window.extendLeaflet = function() {
   L.CRS.Earth.R = 6367000;
 
   var s2SphericalMercator = L.Util.extend({}, L.Projection.SphericalMercator, {
-    R: window.EARTH_RADIUS,
+    R: 6378137,
     bounds: (function () {
-      var d = window.EARTH_RADIUS * Math.PI;
+      var d = this.R * Math.PI;
       return L.bounds([-d, -d], [d, d]);
     })()
   });
@@ -644,7 +643,6 @@ window.extendLeaflet = function() {
     }())
   });
 
-  */
 
   // Fix Leaflet: handle touchcancel events in Draggable
   L.Draggable.prototype._onDownOrig = L.Draggable.prototype._onDown;


johnd0e:
@McBen Please provide some sample steps we should use to compare draw-tools value with ingress value.

McBen:
the distance between  https://intel.ingress.com/intel?pll=18.887762,98.862276
and https://intel.ingress.com/intel?pll=51.464355,6.735444

client shows: **8514.9km**
iitc-ce shows: **8520,20km**

/edit note: old iitc shows correct value

johnd0e:
In fact we know nothing about Ingress backend.
But this constant can be found in https://developers.google.com/maps/documentation/javascript/reference/geometry

> `google.maps.geometry.spherical` namespace

> Utility functions for computing geodesic angles, distances and areas. The default radius is Earth's radius of **6378137** meters.  

P.S.
I'm going to apply this and some more related changes

McBen:
we know that Nia uses s2 everywhere
and constant taken from:
https://github.com/google/s2-geometry-library-java/blob/c28f287b996c0cedc5516a0426fbd49f6c9611ec/src/com/google/common/geometry/S2LatLng.java#L31

johnd0e:
@McBen But now you can clearly see that S2 value is different

McBen:
code never lies, comments/docs (sometimes) do :)

Jormund:
> In fact we know nothing about Ingress backend.

True. But then the mobile app is the next best thing. The distance showed in the game is probably the same used to know if portal range is enough.

I've a key showing 10031.4km in my inventory.
IITC 0.26 tells me 10031.41km
IITC 0.30.1 tells me 10037.71km
google.maps.geometry.spherical says 10048.96km

`a=new google.maps.LatLng(48.864478,2.352135);
b=new google.maps.LatLng(-13.174728,-72.541879);
google.maps.geometry.spherical.computeDistanceBetween(a,b);`
https://intel.ingress.com/intel?pll=48.864478,2.352135
https://intel.ingress.com/intel?pll=-13.174728,-72.541879

johnd0e:
@Jormund That function has 3rd optional argument. So try with some values, mentioned here.

Jormund:
`a=new google.maps.LatLng(48.864478,2.352135);
b=new google.maps.LatLng(-13.174728,-72.541879);
c=6378137;
google.maps.geometry.spherical.computeDistanceBetween(a,b,c);`
10048.96km

`a=new google.maps.LatLng(48.864478,2.352135);
b=new google.maps.LatLng(-13.174728,-72.541879);
c=6367000;
google.maps.geometry.spherical.computeDistanceBetween(a,b,c);`
10031.41km

`a=new google.maps.LatLng(48.864478,2.352135);
b=new google.maps.LatLng(-13.174728,-72.541879);
c=6371008.7714;//mean radius from IUGG
google.maps.geometry.spherical.computeDistanceBetween(a,b,c);`
10037.73km

`a=new google.maps.LatLng(48.864478,2.352135);
b=new google.maps.LatLng(-13.174728,-72.541879);
c=6371007.1810;//radius of sphere of same surface from IUGG
google.maps.geometry.spherical.computeDistanceBetween(a,b,c);`
10037.72km

https://en.wikipedia.org/wiki/Earth_radius#Published_values

johnd0e:
So the only needed line of code here is `L.CRS.Earth.R = 6367000;`

johnd0e:
BTW, actual S2 sources are here https://github.com/google/s2geometry, and there is no this constant now.

Java port mentioned by McBen seems abandoned.
Anyway, we have enough evidence that at least Ingress client uses that old constant for (at least some) calculations, so let's follow it.

McBen:
The radius value I use here is the value used be the good old iitc.me.
iitc.ce introduced the wrong calculation and I tried to fix it. I also tried to explain why iitc.me used this way of calculation. but this is getting boring. 

take it or leave it. do you own work on base of this or prove this is wrong.

> So the only needed line of code here is L.CRS.Earth.R = 6367000;

nope. thats not enough. You need to setup a CRS which uses the 'correct' values.

johnd0e:
> but this is getting boring.

Relax. I am leaving all these notes to myself.

> nope. thats not enough. You need to setup a CRS which uses the 'correct' values.

But it is not in your PR.

AFAIK currently iitc (and plugins) code is affected by L.CRS.Earth value only.
Old iitc had also patch for Mercator, but in wrong place.

PR #333, opened 2020-02-16
cross-links: improved calculation

McBen:
By converting latlng into the cartesian coordination system we can simplify the whole
calculation.

- lesser trigonometric functions calls for a little more multiplications
- don't have to deal with antimeridian
- doesn't use a internal class 

In my personal test it was 2-5times faster then the old algorithm. But this surely depends on system and is hard to measure.

### DIFF
--- a/plugins/cross-links.js
+++ b/plugins/cross-links.js
@@ -111,159 +111,158 @@ function dot(t, n) {
 }
 
 
-window.plugin.crossLinks.testPolyLine = function (polyline, link,closed) {
+window.plugin.crossLinks.testPolyLine = function (polyline, link, closed) {
 
-    var a = link.getLatLngs();
-    var b = polyline.getLatLngs();
+  var a = link.getLatLngs();
+  var b = polyline.getLatLngs();
 
-    for (var i=0;i<b.length-1;++i) {
-        if (window.plugin.crossLinks.greatCircleArcIntersect(a[0],a[1],b[i],b[i+1])) return true;
-    }
+  for (var i = 0; i < b.length - 1; ++i) {
+    if (window.plugin.crossLinks.greatCircleArcIntersect(a[0], a[1], b[i], b[i + 1])) return true;
+  }
 
-    if (closed) {
-        if (window.plugin.crossLinks.greatCircleArcIntersect(a[0],a[1],b[b.length-1],b[0])) return true;
-    }
+  if (closed) {
+    if (window.plugin.crossLinks.greatCircleArcIntersect(a[0], a[1], b[b.length - 1], b[0])) return true;
+  }
 
-    return false;
+  return false;
 }
 
 window.plugin.crossLinks.onLinkAdded = function (data) {
-    if (window.plugin.crossLinks.disabled) return;
+  if (window.plugin.crossLinks.disabled) return;
 
-    plugin.crossLinks.testLink(data.link);
+  plugin.crossLinks.testLink(data.link);
 }
 
-window.plugin.crossLinks.checkAllLinks = function() {
-    if (window.plugin.crossLinks.disabled) return;
+window.plugin.crossLinks.checkAllLinks = function () {
+  if (window.plugin.crossLinks.disabled) return;
 
-    console.debug("Cross-Links: checking all links");
-    plugin.crossLinks.linkLayer.clearLayers();
-    plugin.crossLinks.linkLayerGuids = {};
+  console.debug('Cross-Links: checking all links');
+  plugin.crossLinks.linkLayer.clearLayers();
+  plugin.crossLinks.linkLayerGuids = {};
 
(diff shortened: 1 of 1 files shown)


johnd0e:
Looks great for me.
Just minor remark: consider https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Code-Style

- we are still following es 5.1, and avoiding features like const. See #47
- indent should be 2 spaces

johnd0e (approved):
LGFM

plugins/cross-links.js
```diff
@@ -1,293 +1,268 @@
 // @author         mcben
 // @name           Cross links
 // @category       Draw
-// @version        1.2.0
+// @version        1.3.0
 // @description    Checks for existing links that cross planned links. Requires draw-tools plugin.
 
 
-window.plugin.crossLinks = function() {};
+window.plugin.crossLinks = function () { };
```
johnd0e, line 8: What if we make all namespace references shorter?
```js
var crossLinks = window.plugin.window.plugin.crossLinks = function () {};
```

And then:
```js
crossLinks.greatCircleArcIntersect = function...
crossLinks.testPolyLine = function...
// etc
```
BTW, is it necessary to expose all this functions in namespace?
May be better left some of them private?

plugins/cross-links.js
```diff
@@ -1,293 +1,268 @@
...
 
-window.plugin.crossLinks.greatCircleArcIntersect = function(a0,a1,b0,b1) {
-  // based on the formula at http://williams.best.vwh.net/avform.htm#Int
+/**
+ * greatCircleArcIntersect
+ */
+window.plugin.crossLinks.greatCircleArcIntersect = function (a0, a1, b0, b1) {
```
johnd0e, line 13: If this (and others) function can be useful for 3rd-party plugins, then better to add short description of it's action and arguments.

plugins/cross-links.js
```diff
@@ -1,293 +1,268 @@
...
-       opacity: 0.7,
-       weight: 5,
-       interactive: false,
-       dashArray: '8,8',
+  var poly = L.geodesicPolyline(link.getLatLngs(), {
+    color: '#d22',
+    opacity: 0.7,
+    weight: 5,
+    interactive: false,
+    dashArray: '8,8',
```
johnd0e, line 175: Why not put line options into namespace, for easy customizing?

plugins/cross-links.js
```diff
@@ -1,293 +1,268 @@
...
 }
 
-var setup = function() {
-    if (window.plugin.drawTools === undefined) {
-       alert("'Cross-Links' requires 'draw-tools'");
-       return;
-    }
+var setup = function () {
+  if (window.plugin.drawTools === undefined) {
+    alert('\'Cross-Links\' requires \'draw-tools\'');
```
johnd0e, line 248:
```suggestion
  if (!window.plugin) {
    alert("'Cross-Links' requires 'draw-tools'");
```

plugins/cross-links.js
```diff
@@ -1,293 +1,268 @@
...
+      window.plugin.crossLinks.testAllLinksAgainstLayer(e.layer);
+    } else {
+      // all other event types - assume anything could have been modified and re-check all links
+      window.plugin.crossLinks.checkAllLinks();
+    }
+  });
 
-    window.addHook('linkAdded', window.plugin.crossLinks.onLinkAdded);
-    window.addHook('mapDataRefreshEnd', window.plugin.crossLinks.onMapDataRefreshEnd);
+  window.addHook('linkAdded', window.plugin.crossLinks.onLinkAdded);
```
johnd0e, line 265: Shouldn't `linkRemoved` be handled too?

johnd0e:
@McBen 
We can merge this PR as is, without addressing upper comments.
Please consider them as general development discussion.

McBen:
This PR is about "greatCircleArcIntersect".

The other changes where made to match your format rules as good as possible (and were done by my IDE)

plugins/cross-links.js
McBen: I don't like chained assignments

johnd0e: Well, it's actually does not matter.
Consider this:
```js
var crossLinks = function () {};
window.plugin.window.plugin.crossLinks = crossLinks;
```

plugins/cross-links.js
McBen: Rule nr.8:  Avoid Negative Conditionals

johnd0e:
@McBen 
It's true, and I have zero objections about it's merging.

Still, I'd like to continue discussion somewere.
So should I open separate issue, like _"Some ways to enhance iitc plugins code, on the example of `'cross-links'`"_?

modos189 (approved):
good job!

PR #335, opened 2020-02-17
add ESLint + rules

McBen:
ESLint and rules based on  https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Code-Style


package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "iitc-ce",
+  "version": "1.0.0",
+  "main": "core/total-conversion-build.js",
+  "repository": "https://github.com/IITC-CE/ingress-intel-total-conversion.git",
+  "author": ".",
+  "license": "MIT",
+  "dependencies": {
```
johnd0e:
```suggestion
  "devDependencies": {
```

package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "iitc-ce",
+  "version": "1.0.0",
+  "main": "core/total-conversion-build.js",
+  "repository": "https://github.com/IITC-CE/ingress-intel-total-conversion.git",
+  "author": ".",
+  "license": "MIT",
```
johnd0e:
```suggestion
  "license": "ISC",
```

package.json
```diff
@@ -0,0 +1,13 @@
...
+  "dependencies": {
+    "eslint": "^6.8.0",
+    "eslint-plugin-es5": "^1.5.0",
+    "eslint-plugin-vue": "^6.2.0"
```
johnd0e: Do we really need these?

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
+{
+    "env": {
+        "browser": true,
+        "es6": false,
```
johnd0e, line 4: It is `false` by default

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "SharedArrayBuffer": "readonly"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018,
+        "sourceType": "module"
+    },
+    "plugins": [
+        "vue",
+        "es5"
+    ],
```
johnd0e: Do we really need these?

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "eslint:recommended",
+        "plugin:vue/essential",
+        "plugin:es5/no-es2016"
+    ],
+    "globals": {
+        "Atomics": "readonly",
+        "SharedArrayBuffer": "readonly"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018,
```
johnd0e: Why not just `"ecmaVersion": 5`

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "greasemonkey": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:vue/essential",
+        "plugin:es5/no-es2016"
+    ],
+    "globals": {
+        "Atomics": "readonly",
+        "SharedArrayBuffer": "readonly"
```
johnd0e: What are these?

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018,
+        "sourceType": "module"
```
johnd0e: Our sources are not ECMAScript modules

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+    "env": {
+        "browser": true,
+        "es6": false,
+        "jquery": true,
+        "greasemonkey": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:vue/essential",
+        "plugin:es5/no-es2016"
```
johnd0e: Do we really need these?

package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "iitc-ce",
+  "version": "1.0.0",
```
johnd0e:
```suggestion
  "version": "0.1.0",
```

package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "iitc-ce",
```
johnd0e, line 2: Maybe also
```
  "description": "ingress intel total conversion (IITC-CE)",
```
```
  "keywords": [
    "iitc",
    "ingress"
  ],
```
```
  "private": true,
```

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+    "parserOptions": {
+        "ecmaVersion": 2018,
+        "sourceType": "module"
+    },
+    "plugins": [
+        "vue",
+        "es5"
+    ],
+    "rules": {
+        "no-undef": 0,
```
johnd0e:
```suggestion
        "no-undef": "error"
```
Bot it is default value, so it is not necessary to specify it explicitly.

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "eqeqeq": "error",
+        "indent": ["warn", 2],
+        "brace-style": "error",
```
johnd0e:
```suggestion
        "brace-style": ["error","1tbs", { "allowSingleLine": true }],
```

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "keyword-spacing": ["error", { "before": true }],
```
johnd0e:
```suggestion
        "keyword-spacing": "error",
```
No need in `before` as it by default.

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "spaced-comment": ["error", "always"],
```
johnd0e:
```suggestion
        "spaced-comment": "error",
```

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "quotes": ["error", "single", { "avoidEscape": true }],
+        "no-const-assign": "error",
```
johnd0e: No need in es5, as it is anyway  handled like this:
```
Parsing error: The keyword 'const' is reserved [Error]
```

.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "no-trailing-spaces": "error",
+        "max-len": ["warn", { "code": 160 }]
+        // "no-var": "error"
```
johnd0e:
```suggestion
```

johnd0e (changes requested):
And more:
- add https://eslint.org/docs/rules/semi
- remove `yarn.lock`

johnd0e:
Todo: may be add more rules, like:
- [eol-last](https://eslint.org/docs/rules/eol-last)
- [quote-props](https://eslint.org/docs/rules/quote-props) (`"quote-props": ["error","consistent-as-needed"],`)
- [no-unused-expressions](https://eslint.org/docs/rules/no-unused-expressions)
- etc

package.json
McBen: changed

package.json
McBen: removed

package.json
McBen: fixed

.eslintrc.json
McBen: it's just to express we are using es5

.eslintrc.json
McBen: removed

.eslintrc.json
McBen: add

.eslintrc.json
McBen: note sure. were added by eslint --init. I guess related to vue?

.eslintrc.json
McBen: removed

.eslintrc.json
McBen: done

.eslintrc.json
McBen: done

.eslintrc.json
McBen: see below

.eslintrc.json
McBen: removed

package.json
McBen: removed version - unused anyway

package.json
McBen: what ever you wish ... unused anyway

.eslintrc.json
McBen: removed

.eslintrc.json
McBen: removed

.eslintrc.json
McBen: removed

.gitignore
```diff
@@ -8,6 +8,8 @@ local*.py
 .pydevproject
 .DS_Store
 
+node_modules
+
```
johnd0e, line 12: Shouldn't be part of this PR

.eslintrc.json
```diff
@@ -0,0 +1,25 @@
...
+        "spaced-comment": "error",
+        "quotes": ["error", "single", { "avoidEscape": true }],
+        "no-trailing-spaces": "error",
+        "max-len": ["warn", { "code": 160 }],
+        "semi": "error",
+        "eol-last": "error",
+        "quote-props": ["error","consistent-as-needed"],
+        "no-unused-expressions": "error"
+    }
+}
```
johnd0e: Perhaps EOL needed here

package.json
```diff
@@ -0,0 +1,16 @@
...
+    "iitc",
+    "ingress"
+  ],
+  "main": "core/total-conversion-build.js",
+  "repository": "https://github.com/IITC-CE/ingress-intel-total-conversion.git",
+  "author": ".",
+  "license": "ISC",
+  "private": true,
+  "devDependencies": {
+    "eslint": "^7.5.0"
```
johnd0e: Not sure about it, because in fact it is not used yet.
And current version is 7.16, so may be better just keep `^7` (no difference, but more expessive)

.eslintrc.json
```diff
@@ -0,0 +1,25 @@
...
+    "rules": {
+        "eqeqeq": "error",
+        "indent": ["warn", 2],
+        "brace-style": ["error","1tbs", { "allowSingleLine": true }],
+        "keyword-spacing": "error",
+        "spaced-comment": "error",
+        "quotes": ["error", "single", { "avoidEscape": true }],
+        "no-trailing-spaces": "error",
+        "max-len": ["warn", { "code": 160 }],
+        "semi": "error",
```
johnd0e, line 25: It would be more convenient to have the same rules order as we have is style guide: https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Code-Style

johnd0e (changes requested):

johnd0e:
@modos189 
Please review these rules: https://github.com/IITC-CE/ingress-intel-total-conversion/pull/335#issuecomment-588163570
They weren't part of original style guide (but we include if you are ok with them).

Alternatively, we may set them as warnings (not errors).

.gitignore
McBen: This PR introduces eslint node_package.
Adding packages without this would be a really bad idea.

johnd0e: Ah, right.

package.json
McBen: could be updated, yes - when doing this PR 7.5 was latest-release. 

I don't like automatic version updates of underlying libs. (eslint might be an exception here - but I don't like exception either :) )

.eslintrc.json
McBen: not required but could be added

.eslintrc.json
McBen: it's a configuration file - noone will every 'read' these files. There is a loosly connetion to the wikipage not a dependency.

McBen:
Maybe the rule `no-undef` and `no-unused-vars` should be disabled. 

They will generate false-positives because the custom build-script concept isn't really supported by any other tool.

johnd0e:
> Maybe the rule `no-undef` and `no-unused-vars` should be disabled.

Do you mean 'enabled'? Because you do not have those in the PR.

> They will generate false-positives because the custom build-script concept isn't really supported by any other tool.

Explain this please. I am not sure what false positives do you mean, and how they are related to some custom build-script.

McBen:
> > Maybe the rule `no-undef` and `no-unused-vars` should be disabled.
> 
> Do you mean 'enabled'? Because you do not have those in the PR.

This rule-set is based on `eslint:recommended`

 
> > They will generate false-positives because the custom build-script concept isn't really supported by any other tool.
> 
> Explain this please. I am not sure what false positives do you mean, and how they are related to some custom build-script.

Examples:
```
./core/total-conversion-build.js
    8:22  error    'plugin_info' is not defined                                           no-undef
   23:24  error    'PLAYER' is not defined                                                no-undef
  177:5   error    'portalsFactionLayers' is defined but never used                       no-unused-vars
  177:27  error    'linksFactionLayers' is defined but never used                         no-unused-vars
  177:47  error    'fieldsFactionLayers' is defined but never used                        no-unused-vars
```

johnd0e:
I see now. But I do not agree: those are not false-positives, those are places for improvements, as we definitely are able to clearly express our intentions, in number of ways, e.g.:
- use `window.` notation
- use exclusion list (custom list of globals)
- use inline exclusions (in comments)

~~And I still do not understand how this is related to custom build script.
We have no linting step now, true. But it is easy to add (when the sources will be ready).~~

**Update**
Perhaps you mean that in webpack there will be explicit import/export statements.
BTW:
```js
var portalsFactionLayers = [];
// ...........................
window.portalsFactionLayers = portalsFactionLayers; // example of some primitive form of export
```
Then:
```js
var portalsFactionLayers = window.portalsFactionLayers;  // example of some primitive form of import
```

johnd0e:
https://eslint.org/docs/rules/curly#all
?

package.json
```diff
@@ -0,0 +1,16 @@
...
+    "iitc",
+    "ingress"
+  ],
+  "main": "core/total-conversion-build.js",
+  "repository": "https://github.com/IITC-CE/ingress-intel-total-conversion.git",
+  "author": ".",
+  "license": "ISC",
+  "private": true,
+  "devDependencies": {
+    "eslint": "^7.16.0"
```
johnd0e: https://eslint.org/blog/2021/02/eslint-v7.20.0-released

McBen: Thats the job of the package manager. There is no need to keep track of every update.
`npm install`/`yarn install` (or upgrade) will handle it by themself.

McBen:
> 
> 
> https://eslint.org/docs/rules/curly#all
> ?


d9f4e7f

johnd0e:
@McBen 
Do you insist on `multi-line` option?
I'd prefer `all`.

McBen:
> Do you insist on `multi` option?
> I'd prefer `all`.

I don't insist on anything. I just like to use constructs like 
`if (error) return;` 
(instead of `if (error) { return; }`)

modos189:
I like the following options:

`eol-last: always`
`quote-props: always`
`curly: multi-line`

It hardly makes sense to use `no-undef` and `no-unused-vars`, because we are using our own build system instead of webpack and this should confuse eslint.

But what about third-party libraries? Possible incompatibilities with their style won't cause problems?

McBen:
> `eol-last: always`

is default

> `quote-props: always`

code like this would trigger an error:
`marker.setStyle({ opacity: 0.4, fillOpacity: 0.1 });`

> `curly: multi-line`

makes 2 votes for multi-line 
 
> But what about third-party libraries? Possible incompatibilities with their style won't cause problems?

commit 55066a0 - ESLint: Ignore externals
but since we never edit external source this should not be a prolbem ;)

johnd0e:
> > `eol-last: always`
> 
> is default

Where do you see such defaults?
Because it is not check-marked in [rules list](https://eslint.org/docs/rules/).

McBen:
> 
> 
> > > `eol-last: always`
> > 
> > 
> > is default
> 
> Where do you see such defaults?
> Because it is not check-marked in [rules list](https://eslint.org/docs/rules/).

https://eslint.org/docs/rules/eol-last

eol-last is not a default rule. But 'always' is the default option for eol-last.

package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "iitc-ce",
+  "description": "ingress intel total conversion (IITC-CE)",
+  "keywords": [
+    "iitc",
+    "ingress"
+  ],
+  "main": "core/total-conversion-build.js",
```
johnd0e: I'm not sure about this field.
E.g. I see that Leaflet uses built script path for it: https://github.com/Leaflet/Leaflet/blob/234e00bcab0d89319f4c6874f3063ac311ce8cc6/package.json#L33

package.json
```diff
@@ -0,0 +1,16 @@
...
+  "repository": "https://github.com/IITC-CE/ingress-intel-total-conversion.git",
+  "author": ".",
```
johnd0e:
```suggestion
```
What's meant here?

package.json
le-jeu: same as `author`, is this usefull ? this is not a package (yet?)

johnd0e (approved):

modos189 (approved):

Issue #241, opened 2019-08-28
[Sync issue] Bookmark does not remember the order of bookmarked maps


djserj:
Hi All,
I did not find the issue with the plug-in so creating this one here. I've created a couple of folders in Maps section of the Bookmarks and several maps there. I tried to order maps in my map folders but once I create a new map bookmark, all bookmarks fall fown into the bottom list outside my folders.
I'm using IITC-CE button add-in for Firefox with Bookmarks plug-in.
For the portals bookmarks it is the same - if you order your portals through folders, and then create a new bookmark for a portal, other bookmarks fall down into the bottom list outside your folders. You have to sort them from scratch.

johnd0e:
@djserj Is that behavior specific to IITC-CE, or it stays the same when using old IITC?

djserj:
It is specific to IITC CE, because I have another browser with an old IITC version with the same add-in, there is no such behavior, I just tested it to double check. I did not face the issue previously in that version.

johnd0e:
@djserj Have you tried latest test build?
https://iitc.modos189.ru/test_builds.html
(update both main script AND bookmarks plugin)

djserj:
Yes, just tried test build and the same

johnd0e:
I am unable to reproduce the issue.

Try to disable all unrelated plugins, and clean all the data.

mvolfik:
I also experienced this, what happens is that you save some portals/maps to bookmarks, then sync saves them to cloud. Then you move them to folders etc., but then sync not updates the cloud data, but loads them and overwrites your bookmarks. Then if you add one the bookmarks window refreshes and you see them all out of folders.

angel93ayora:
https://github.com/IITC-CE/ingress-intel-total-conversion/issues/166
is this related? ^ maybe more jquery incompatibilities

@djserj have you tried latest test build of bookmarks plugin?

PR #346, opened 2020-02-26
WIP: Bookmarks with SortableJS

Aradiv:
This PR is a WIP rewrite of the bookmarks plugin.

There are still some issues open and stuff that needs to be fixed.

Known Bug/ToDo List

- [ ]  folders can be sorted below idOthers but this will be reverted on the next refresh.
  ToDo: disallow sorting directory below idOthers
- [ ] ToDo: test Mobile
- [ ] ToDo:  (optional) check loadStorage() call and remove where they are not neccessary
- [ ] ToDo: a lot more testing


plugins/bookmarks.js
```diff
@@ -1,10 +1,13 @@
-// @author         ZasoGD
+// @author         ZasoGD / Aradiv
```
johnd0e, line 1: Let's reserve this field to original author, and do not change unless you rewrite plugin from scratch.

You may want to introduce separate field `@coauthor` or `@collaborator` or `@contributor`, but then we should keep same approach through all plugins, which currently is not in case.

johnd0e:
While working on this PR - may be you will be able to find what is root source of #241

johnd0e:
As I can see SortableJS should support touch.
jQueryUI fails with this (though there are plugins like https://github.com/heyman/jquery-draggable-touch and https://www.npmjs.com/package/touch-dnd).

plugins/bookmarks.js
MysticJay: While this is a different issue, I added myself to a script internal history (which I still prefer), but it was not excepted either.
As scripts tend to be privately improved, extended or modified in a different ways and get shared outside of GIT and version numbers have their own lives anyway, we need to address this somehow. @johnd0e please create a separate discussion issue on this.

PR #351, opened 2020-02-29
fix-china-map-offset: apply fix from upstream

johnd0e:
https://github.com/Artoria2e5/PRCoords/issues/10
https://github.com/Artoria2e5/PRCoords/commit/026c571b528177ebb8fe0348474385298499c6b9#diff-de8eeac97a630cd112642780ec1d948cR124

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/351)


modos189 (approved):
I notice a slight shift of a few decimeters compared to previous behavior. Apparently, that's how it's supposed to be.

PR #356, opened 2020-03-05
build_plugin.py: some code can be placed outside of wrapper

### LINKED ISSUE #354: can't put code outside of wrapper with new build system
i need to place some code outside of wrapper() since i have various things running in GM sandbox that don't want to expose.

Functions that need to be accessible are placed inside wrapper everything else is outside wrapper

with the old build system i just placed the code before the @@PLUGINSTART@@ and could control where my code is running.

johnd0e:
Close #354

To leave some code out of wrapper use special marker (on separate line):

```js
'this_is_unwrapped';
/*wrapped-from-here*/
```

### DIFF
--- a/build_plugin.py
+++ b/build_plugin.py
@@ -166,18 +166,22 @@ def process_file(source, out_dir, dist_path=None, deps_list=None):
 
     path = source.parent  # used as root for all (relative) paths
     script = re.sub(r"'@bundle_code@';", partial(bundle_code, path=path), script)
+    try:
+        script_before_wrapper, script = script.split('\n/*wrapped-from-here*/\n', 1)
+    except ValueError:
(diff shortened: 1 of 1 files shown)


johnd0e:
@Aradiv

Please check if it works as expected by you.

johnd0e:
It seems that #358 has better solution

modos189 (approved):
Looks bad, but as a temporary solution is acceptable. Especially, as I understand it, after #358 there will simply be no need to comment /*wrapped-from-here*/ - it will work without this comment.

johnd0e:
No, #358 was initially based on wrong assumption.
It cannot be used in GreaseMonkey and IITC-Button.

Issue #298, opened 2019-12-23
IITC-Mobile: depends on exact plugin wrapper code


johnd0e:
IITCm tries to remove wrapped relying on this hardcoded values:

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/4e0f388d9cdeb50519f356a5d90807006f43f9c8/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java#L59-L61

Aradiv:
So every plugin not using this exact wrapper code doesn't work?

johnd0e:
It will work in any case, but not in optimal way.

PR #358, opened 2020-03-07
better plugin wrapper

johnd0e:
Use `unsafeWindow`, thus no need to 'inject script in site context' anymore.
- ~~wrapper is simpler now~~
  Just modernized a bit.
  Main goal: avoid script re-injection if main context is directly accessible (like in Tampermonkey, Violentmonkey, and mobile app).
  + Direct script execution make debug easier:
    To set breakpoints switch to Web Developer Tools' Debugger/Sources tab
    and find related script in Tampermonkey's list of `userscript.html?id=<guid>`
  + no need to cut wrapper out in mobile app (see #298)
    (but we keep that code for all legacy plugins)
  + ~~it's possible to use GM_* functions directly (#354)~~
    still not possible in some environments, so stick to #356
- in unsupported environments (Greasemonkey, IITC-Button) nothing changed - previous injection-method is used
- whole `scriptMetaStr` is added to plugin_info, to be able to analyse more properties (#242)

- [x] New wrapper is checked with major usescript managers (Tampermonkey, Greasemonkey, Violentmonkey), and is compatible with plugins embedding old wrapper.

TODO:
- [ ] Make sure that mobile app is ready
- [ ] ~~More tests needed with misc plugins, testing different apis (such as XHR)
  https://web.archive.org/web/20141008055234/http://www.oreillynet.com/lpt/a/6257~~
- [ ] ~~IITC Button is not ready yet: https://github.com/IITC-CE/IITC-Button/issues/32~~
- [ ] May be we need some markers in code, to define exact places of different wrapper parts.
  That can be used for easy updating of embedded wrapper.
- [ ] May be we need some additional property to pass additional data from userscript manager
- [ ] May be pick another name for `plugin_info`
- [ ] It's possible to keep reference to original sandboxed `window`.
  But this also can be done manually, so probably not
- [ ] May be it's time to introduce `setup` arguments.
  We could pass iitc api as argument. And may be something other, e.g. logger object
- [ ] Some more improvements pending (see below)

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/358)


pluginwrapper.py
```diff
@@ -1,42 +1,35 @@
...
-if (typeof GM_info !== 'undefined' && GM_info && GM_info.script) info.script = { version: GM_info.script.version, name: GM_info.script.name, description: GM_info.script.description };
-script.appendChild(document.createTextNode('('+ wrapper +')('+JSON.stringify(info)+');'));
-(document.body || document.head || document.documentElement).appendChild(script);
 
+if (typeof setup !== 'function') {
+  var setup = {}; plugin_info.error = 'setup is not a function';
+}
+setup.info = plugin_info;
+(window.bootPlugins = window.bootPlugins || []).push(setup);
+if (window.iitcLoaded) { setup(); }
```
johnd0e, line 31: Perhaps better expose this: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/ea9cb98e73ec0dcc53addd955fdae7edc57b0891/core/code/boot.js#L689

Otherwise we would need special handling to see such plugins in About: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/ea9cb98e73ec0dcc53addd955fdae7edc57b0891/core/code/utils_misc.js#L3

pluginwrapper.py
```diff
@@ -1,42 +1,35 @@
...
 
-//PLUGIN AUTHORS: writing a plugin outside of the IITC build environment? if so, delete these lines!!
-//(leaving them in place might break the 'About IITC' page or break update checks)
+// PLUGIN AUTHORS: writing a plugin outside of the IITC build environment? if so, delete these lines!!
+// (leaving them in place might break the 'About IITC' page or break update checks)
 plugin_info.buildName = '@build_name@';
 plugin_info.dateTimeVersion = '@build_date@';
 plugin_info.pluginId = '@plugin_id@';
-//END PLUGIN AUTHORS NOTE
+// END PLUGIN AUTHORS NOTE
```
johnd0e, line 21: We can avoid this, if we add this info to meta block, see #242.

pluginwrapper.py
```diff
@@ -1,42 +1,35 @@
...
+# plugin wrapper code snippets
 
 # a cut-down version of GM_info is passed as a parameter to the script
 # (not the full GM_info - it contains the ENTIRE script source!)
 
 start = """
-function wrapper(plugin_info) {
-// ensure plugin framework is there, even if iitc is not yet loaded
-if(typeof window.plugin !== 'function') window.plugin = function() {};
+var plugin_info = (typeof GM_info === 'undefined') ? {} : (function (s) {
```
johnd0e: Actually we should rely on `GM_info` definition, but iOS is not ready: https://github.com/HubertZhang/IITC-Mobile/issues/1

johnd0e:
Unfortunately we still cannot avoid script injecting for Greasemonkey (and IITC-Button).
As appeared, such code is just ignored:

```js
window = ...
```

It'd be possible to replace it with local variable, but it's not enough for good compatibility with all existing 3rd-party plugins: we also need to replace [global](https://developer.mozilla.org/en-US/docs/Glossary/Global_object).

I do not know it it ever possible.
I've tried to apply `with (unsafeWindow)` statement, but then we get other issues (`Error: Permission denied to access property [...]`)

So for Greasemonkey I have to restore previous method involving script injection.

le-jeu:
>     * Direct script execution make debug easier:
>       To set breakpoints switch to Web Developer Tools' Debugger/Sources tab
>       and find related script in Tampermonkey's list of `userscript.html?id=<guid>`

I tried to use `sourceURL` to name script and make them appear in the source tree
```js
// ...
script.appendChild(document.createTextNode('('+ wrapper +')('+JSON.stringify(info)+');'));
if (info.script && info.script.name)
  script.appendChild(document.createTextNode(`//# sourceURL=${encodeURIComponent(info.script.name)}`));
(document.body || document.head || document.documentElement).appendChild(script);
```
I guess this is not compatible with IITC Mobile but the inserted script is correctly listed in source (from intel.ingress.com) (firefox: violent/grease monkey)

I used `info.script.name` but this can also be done by the build system
Sadly, the `//#` has to be added before the script is inserted in the DOM (wtr to iitc mobile)

**Edit:** Possible workaround for mobile
```js
// if on last IITC mobile, will be replaced by wrapper(info)
var mobile = `script.appendChild(document.createTextNode('('+ wrapper +')('+JSON.stringify(info)+');'));
(document.body || document.head || document.documentElement).appendChild(script);`;

if (mobile.startsWith('script')) {
  script.appendChild(document.createTextNode('('+ wrapper +')('+JSON.stringify(info)+');'));
  if (info.script && info.script.name)
    script.appendChild(document.createTextNode('//# sourceURL=' + encodeURIComponent(info.script.name)));
  (document.body || document.head || document.documentElement).appendChild(script);
} else {
  // mobile string
  wrapper(info);
}
```

PR #359, opened 2020-04-23
build.py: fix to support Python 3.8.2

modos189:
* It seems that python3.8 doesn't allow path like objects in runpy.run_path any more. (see https://github.com/spacemanspiff2007/HABApp/issues/111)
* ~In version 3.8 --bind argument enhanced to support IPv6. (https://docs.python.org/3/library/http.server.html#http.server.SimpleHTTPRequestHandler) this causes the server to start at http://[::1]:8000/.~

### DIFF
--- a/build.py
+++ b/build.py
@@ -17,7 +17,7 @@ def run_python(cmd):
         cmd = Path(__file__).with_name(cmd)
         if not cmd.is_file():
             raise UserWarning(f'no such file: {original}')
-    return run_path(cmd)
+    return run_path(str(cmd))
 
 
 def run_system(cmd):


build.py
```diff
@@ -17,7 +17,7 @@ def run_python(cmd):
         cmd = Path(__file__).with_name(cmd)
         if not cmd.is_file():
             raise UserWarning(f'no such file: {original}')
-    return run_path(cmd)
+    return run_path(str(cmd))
```
johnd0e, line 20: Hm, I was sure that I fixed similar issue previously..
Could you provide sample case to reproduce this bug?

web_server_local.py
```diff
@@ -28,4 +28,4 @@
 
 handler_class = partial(SimpleHTTPRequestHandler, directory=directory)  # Python 3.7+
 print(f'Update channel: {settings.build_name}')
-test(HandlerClass=handler_class, port=args.port, bind='localhost')
+test(HandlerClass=handler_class, port=args.port, bind='127.0.0.1')
```
johnd0e: What is wrong here?
I know that ip6 is default, but why it is not good for you?

build.py
modos189: Yes, just add these lines to build:
```
        'post_build': [
            'web_meta_gen.py'
        ]
```
using python3.8, it causes an error:
```
Traceback (most recent call last):
  File "build.py", line 171, in <module>
    backup_and_run()
  File "build.py", line 87, in backup_and_run
    iitc_build(source, workdir, deps_list=deps_list)
  File "build.py", line 65, in iitc_build
    run_cmds(settings.post_build, source, outdir)
  File "build.py", line 38, in run_cmds
    module = run_python(Path(cmd))
  File "build.py", line 20, in run_python
    return run_path(cmd)
  File "/usr/lib/python3.8/runpy.py", line 262, in run_path
    code, fname = _get_code_from_file(run_name, path_name)
  File "/usr/lib/python3.8/runpy.py", line 232, in _get_code_from_file
    with io.open_code(fname) as f:
TypeError: open_code() argument 'path' must be str, not PosixPath
```

web_server_local.py
modos189: I don't mind IPv6, I mind different addresses depending on version of Python. IITC-Button now uses 127.0.0.1 and requires a manual change to [::1] in settings. It's not good.

johnd0e: You can use `localhost` to cover both.
https://github.com/IITC-CE/IITC-Button/issues/27

modos189: Ok, you are right

build.py
johnd0e: Well, it's weird.
I have Python 3.8.0.
And have these lines in `localbuildsettings.py`:
```py
        'post_build': [
            'web_meta_gen.py',
```
And when I run `build.py` I see no problems:
```
IITC build: local
Generating meta for build: local
```

johnd0e:
This was broken in [3.8.1](https://docs.python.org/3/whatsnew/changelog.html#python-3-8-1-release-candidate-1), and fixed back in [3.8.3](https://docs.python.org/3/whatsnew/changelog.html#python-3-8-3-release-candidate-1) (RC).
More info is here: https://bugs.python.org/issue39517

PR #366, opened 2020-06-27
external: Update ulog to 2.0.0-beta.7

johnd0e:


johnd0e:
Despite current `ulog` version is 'beta 10' we are still sticking to 'beta 7', which is most recent before complete rewrite.
Anyway 'beta 10' is not stable enough.

P.S.
And I'm not sure that we need to update to latest ulog: it is not the same lib as it used to be. Perhaps it is more powerful but it is also bigger (`7618b` vs current `2234b`).
So, before updating we should consider alternative solutions again.

PR #369, opened 2020-06-27
external: update jQuery to 3.5.1

johnd0e:


McBen (commented):
breaks spectrum control.
see: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/381
also: https://github.com/bgrins/spectrum/issues/563  (includes workaround)

Issue #27, opened 2018-12-16
Idea: region selection utility


johnd0e:
There are A LOT of plugins (some we can see in #2) with same demand:

- choose several portals for further processing
- or just limit processing objects to some area

Instead of re-implement the same functions in each plugin it would be very convenient if IITC provides this as utility function(s).

And more: imho existing implementations are not generic enough.
E.g.:
- there is ability to limit area with polygon, but no ability to exclude some portals.
- draw-tools can be used for area selection, then one of _'point-in-polygon'_ algorithm implementations: #552
- there are many other possible tools (see https://leafletjs.com/plugins.html#areaoverlay-selection)
  + rectangle
    * https://github.com/kajic/leaflet-locationfilter/
    * https://github.com/heyman/leaflet-areaselect
  + polygon
    * https://github.com/bopen/leaflet-area-selection
  + freeform
    * https://github.com/sandropibia/Leaflet.SelectAreaFeature/
    * https://github.com/zakjan/leaflet-lasso

So, the best solution would be to make objects selection tool as plugin.

**Edit**: this is actually not new idea: https://github.com/iitc-project/ingress-intel-total-conversion/issues/736#issuecomment-32256206

johnd0e:
More links:
- poly-fly-links.user.js
  https://github.com/chrislennon/IITCPlugins -> https://github.com/Seraphli/IITCPlugins
  polygon-portals-list.user.js (unfunctional)
- https://github.com/hayeswise/iitc-portalsinpolygons
- https://gist.github.com/seattletechie/db4cdc748a4a27e655aaf6170cadbb05
  iitc-plugin-count-portals-in-poly.user.js
- https://github.com/dasyak/iitc-plugin-selected-portals-count/blob/master/iitc-plugin-selected-portals-count.user.js
- https://github.com/Jormund/bookmark_under_draw/blob/master/bookmark_under_draw.user.js
- https://github.com/Jormund/poly-counts/blob/master/poly-counts2.user.js
  Display a list of all localized portals by level and faction.
- https://git.depad.fr/ingress-res/iitc-plugins/
  carb.poly-counts.user.js
- https://github.com/Hurqalia/ap_under_draw

johnd0e:
Proposed interface: [`L.Evented`](https://leafletjs.com/reference.html#evented) object.

**Methods**
- filter(portal): Boolean
- .isEnabled(): Boolean

**Events**
- 'changed'
- 'enabled', 'disabled'

PR #371, opened 2020-06-30
Update portal counts plugin to integrate with draw tools and uniques

erikchristiansson:
Display count of unvisited and uncaptured portals if the uniques plugin is installed.

If any polygons are draw, only count portals in the polygons.

### DIFF
--- a/plugins/portal-counts.js
+++ b/plugins/portal-counts.js
@@ -1,27 +1,8 @@
-// ==UserScript==
 // @author         yenky
-// @name           IITC plugin: Portal count
+// @name           Portal count
 // @category       Info
 // @version        0.1.2
 // @description    Display a list of all localized portals by level and faction.
-// @id             portal-counts
-// @namespace      https://github.com/IITC-CE/ingress-intel-total-conversion
-// @updateURL      https://iitc.modos189.ru/build/release/plugins/portal-counts.meta.js
-// @downloadURL    https://iitc.modos189.ru/build/release/plugins/portal-counts.user.js
-// @match          https://intel.ingress.com/*
-// @grant          none
-// ==/UserScript==
-
-function wrapper(plugin_info) {
-// ensure plugin framework is there, even if iitc is not yet loaded
-if(typeof window.plugin !== 'function') window.plugin = function() {};
-
-//PLUGIN AUTHORS: writing a plugin outside of the IITC build environment? if so, delete these lines!!
-//(leaving them in place might break the 'About IITC' page or break update checks)
-plugin_info.buildName = 'release';
-plugin_info.dateTimeVersion = '2020-01-18-170317';
-plugin_info.pluginId = 'portal-counts';
-//END PLUGIN AUTHORS NOTE
 
 
 /* whatsnew
--- a/plugins/portal-counts.js
+++ b/plugins/portal-counts.js
@@ -78,7 +78,7 @@ window.plugin.portalcounts.getPortals = function (){
       var latlng = portal._latlng.lat + ',' + portal._latlng.lng;
       for(var dl in drawLayer){
         if(drawLayer[dl].type === 'polygon') {
-          if(window.plugin.portalcounts.portalinpolygon(latlng,drawLayer[dl].latLngs)) {
+          if(window.plugin.portalcounts.portalinpolygon(portal, drawLayer[dl].latLngs)) {
             portalInPolygon = true;
             break;
           }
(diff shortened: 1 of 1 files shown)


plugins/portal-counts.js
```diff
@@ -46,7 +66,28 @@ window.plugin.portalcounts.getPortals = function (){
...
+  if(localStorage['plugin-draw-tools-layer']) {
+    var drawLayer = JSON.parse(localStorage['plugin-draw-tools-layer']);
+  }
   $.each(window.portals, function(i, portal) {
+    if (drawLayer && drawLayer.length) {
+      var portalInPolygon = false;
+      var latlng = portal._latlng.lat + ',' + portal._latlng.lng;
+      for(var dl in drawLayer){
+        if(drawLayer[dl].type === 'polygon') {
+          if(window.plugin.portalcounts.portalinpolygon(latlng,drawLayer[dl].latLngs)) {
```
modos189:
```suggestion
          if(window.plugin.portalcounts.portalinpolygon(portal, drawLayer[dl].latLngs)) {
```

plugins/portal-counts.js
```diff
@@ -46,7 +66,28 @@ window.plugin.portalcounts.getPortals = function (){
...
+  self.upcP = 0;
+  self.upvP = 0;
+
+  if(localStorage['plugin-draw-tools-layer']) {
+    var drawLayer = JSON.parse(localStorage['plugin-draw-tools-layer']);
+  }
   $.each(window.portals, function(i, portal) {
+    if (drawLayer && drawLayer.length) {
+      var portalInPolygon = false;
+      var latlng = portal._latlng.lat + ',' + portal._latlng.lng;
```
modos189:
```suggestion
```

plugins/portal-counts.js
```diff
@@ -30,6 +30,26 @@ window.plugin.portalcounts = {
   CENTER_Y: 100,
 };
 
+/*********** HELPER FUNCTION ****************************************************/
+window.plugin.portalcounts.portalinpolygon = function(portal,LatLngsObjectsArray) {
+  var portalCoords = portal.split(',');
```
modos189:
```suggestion
  var portalCoords = portal.getLatLng();
```

plugins/portal-counts.js
```diff
@@ -30,6 +30,26 @@ window.plugin.portalcounts = {
...
+
+  var x = portalCoords[0], y = portalCoords[1];
```
modos189:
```suggestion
  var x = portalCoords.lat, y = portalCoords.lng;
```

plugins/portal-counts.js
```diff
@@ -30,6 +30,26 @@ window.plugin.portalcounts = {
   CENTER_Y: 100,
 };
 
+/*********** HELPER FUNCTION ****************************************************/
```
modos189:
```suggestion
// Determine if a point is inside of a polygon.
// ray-casting algorithm based on
// https://observablehq.com/@tmcw/understanding-point-in-polygon
```

modos189 (changes requested):
I suggest not to convert coordinates to string when passing a variable to function. And specify a link to description of function. That's what I did in comments to PR.

erikchristiansson:
> I suggest not to convert coordinates to string when passing a variable to function. And specify a link to description of function. That's what I did in comments to PR.

Good suggestions! All proposed changes merged.

modos189 (dismissed):

plugins/portal-counts.js
```diff
@@ -30,6 +30,28 @@ window.plugin.portalcounts = {
   CENTER_Y: 100,
 };
 
+// Determine if a point is inside of a polygon.
+// ray-casting algorithm based on
+// https://observablehq.com/@tmcw/understanding-point-in-polygon
+window.plugin.portalcounts.portalinpolygon = function(portal,LatLngsObjectsArray) {
```
johnd0e, line 36: @modos189 

1. Shouldn't we include some pip algorithm in IITC api? It is used in many plugins already. 
2. I'm going to open separate issue to discuss another opportunity: what if we provide general portal-filtering API?
For example, it'd be possible to use more convenient region selection tools, like these: #27

modos189: It would be great if common functions were included in API. We can replace this code when corresponding API appears

johnd0e:
@erikchristiansson 
Perhaps this now need to be updated to use intel portal history

PR #374, opened 2020-07-12
Fly-link update to 0.4.0

modos189:
(fixed PR #310)

+ region selection with the polygon draw tool
+ exclusion of portals with the marker draw tool
+ steady drawing (portals is not excluded from the drawing if they are
moved out of boundaries)

### DIFF
--- a/plugins/fly-links.js
+++ b/plugins/fly-links.js
@@ -1,7 +1,7 @@
 // @author         fly
 // @name           Fly Links
 // @category       Draw
-// @version        0.2.1
+// @version        0.3.0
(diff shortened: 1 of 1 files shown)


johnd0e:
> (fixed PR #310)

It's not clear what exactly have you fixed.
Wouldn't it be better to use review tools on #310 instead?

modos189:
Now I have highlighted my changes in #310 using the review tools.
I created a separate PR because it seemed that the author hadn't been in touch in a long time

johnd0e:
Applied onto original PR

PR #380, opened 2020-07-26
Fix error message on mobile startup "Cannot read property 'style' of …

McBen:
…null"

Dummy html element for old intel code.

### DIFF
--- a/core/style.css
+++ b/core/style.css
@@ -1156,7 +1156,7 @@ td + td {
   padding-top: 0.7em;
 }
 
-#play_button {
+#play_button, #loading_msg, #display_msg {
   display: none;
 }
 
--- a/core/total-conversion-build.js
+++ b/core/total-conversion-build.js
@@ -93,7 +93,9 @@ document.body.innerHTML = ''
   + '</div>'
   + '<div id="updatestatus"><div id="innerstatus"></div></div>'
   // avoid error by stock JS
-  + '<div id="play_button"></div>';
+  + '<div id="play_button"></div>'
+  + '<div id="loading_msg"></div>'
+  + '<div id="display_msg"></div>';
 
 // CONFIG OPTIONS ////////////////////////////////////////////////////
 window.REFRESH = 30; // refresh view every 30s (base time)


johnd0e:
This is looking good, but how can I reproduce original issue?

On my system I do not have issue with `loading_img`, but I have different error:
```
ark.js:formatted:7113 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at HTMLScriptElement.<anonymous> (https://storage.googleapis.com/nianticweb-ark/ark/ark.js:116:3030)
```

Related code:
```js
                    var u = document.createElement("script");
                    u.charset = "utf-8",
                    u.async = !0,
                    u.addEventListener("error", function() {
                        window.removeEventListener("error", c),
                        o(Error("Error loading " + e + (n ? " from " + n : "")))
                    }),
                    u.addEventListener("load", function() {
                        window.removeEventListener("error", c),
                        document.head.removeChild(u),
                        i ? o(i) : r(t.getRegister())
                    }),
```

McBen:
> how can I reproduce original issue?  

Happens every time I start mobile.  The function triggering the error is called multiple times on startup. Maybe on your system it's finished before iitc clears the body content? IDK. I didn't reversed the rest of intel.

> On my system I do not have issue with loading_img, but I have different error:  

Maybe move from `// @run-at document-end` to `// @run-at document-start` to prevent intel doing anything. But I guess that's not an easy task 'cause we need to fill in all missing intel stuff and can nolonger use the sledgehammer method: `documemt.body = ""`.

johnd0e:
After this PR I observe another similar case on IITC-Button boot (test-builds).
It seems we need one more dummy: `#display_msg`

johnd0e:
Oops.
This PR makes things even worse on mobile:
```
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:287 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:287 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:287 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:246 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:246 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
gen_dashboard_b9f549656fe8bca5f7669955311c38f7edd0680e.js:246 InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama
```
Check yourself in test-builds.

McBen:
I'm pretty sure I tested it long time ago.

Anyway, then don't use this PR. I guess the style exception will prevent the original intel code from doing more nasty stuff.

PR #378, opened 2020-07-21
UNIQUES - Adding support for scouted and droned portals

MysticJay:
- Two new checkboxes allow to indicate if a portal was scouted or if has been visited by the players drone.
- Portal-list allows to sort on either of these.
- Intel-Comm is analyzed to automatically identify portals for UPV, UPC and Scout.
- new highlighters for "Scouted" and "Droned"
- replaced the columns titels:
V - UPV (Unique Portals Visited)
C - UPC (Unique Portals Captured) 
S - User is scout for this portal 
D - User visited this portal with his drone (no related COMM messages available, thus only manual)

Please be aware that Comm detection/analysis is not reliable due to not all data delivered from intel servers.

PR #383, opened 2020-08-02
add drone visits and scout controller

AlterTobi:
add drone visits and scout controller to uniques plugin


MysticJay:
There is already a ready-to-merge PR #378. This one is duplicate

PR #385, opened 2020-08-19
log source code instead of function pointers on error

McBen:
Function pointers doesn't give much hints where they came from.
Use 'toString' to see the source code of the bad function.

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -685,14 +685,14 @@ function prepPluginsToLoad() {
 
   // executes setup function of plugin
   // and collects info for About IITC
-  function safeSetup (setup) {
-    if (!setup) {
+  function safeSetup(setup) {
+    if (typeof setup !== 'function') {
       log.warn('plugin must provide setup function');
       return;
     }
     var info = setup.info;
     if (typeof info !== 'object') {
-      log.warn('plugin does not have proper wrapper:',setup);
+      log.warn('plugin does not have proper wrapper:', { function: setup, info: setup.info, source: setup.toString() });
       info = {};
     }
     try {
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -74,7 +74,7 @@ window.runHooks = function(event, data) {
         return false;  //break from $.each
       }
     } catch (e) {
-      log.error('error running hook '+event+', error: '+e);
+      log.error('error running hook', { event: event, error: e, source: callback.toString(), data: data });
     }
   });
   return !interrupted;
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -64,10 +64,10 @@
 window._hooks = {}
 window.VALID_HOOKS = []; // stub for compatibility
 
-window.runHooks = function(event, data) {
-  if(!_hooks[event]) return true;
+window.runHooks = function (event, data) {
+  if (!_hooks[event]) return true;
   var interrupted = false;
-  $.each(_hooks[event], function(ind, callback) {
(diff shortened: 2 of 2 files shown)


Aradiv:
why don't you just use breakpoints and the debugger?

just printing the entire source also isn't very helpful as we already have the plugin name and can just look into the plugin source.
If you really want to have the code limit the output to only the line + surroundings of the line where the error occurred

McBen:
The current message gives no hint which plugin is raising the error.
grafik
(source link goes to the exception handler 'runHook')

> why don't you just use breakpoints and the debugger?

Sometimes a error occures for unknown reason. Like a plugin failed when a special (unkown) portal is loaded in link-zoom.

> just printing the entire source also isn't very helpful as we already have the plugin nam

as said, I don't see any plugin name here.  It only logs the code of error function.

McBen:
The current message gives no hint which plugin is raising the error. 
 
FF: (with PR)
grafik

/edit..ah, I see chrome does a better job here. Firefox lacks the call stack of the exception.
grafik


> why don't you just use breakpoints and the debugger? 
 
Sometimes an error occures for unknown reason. Like a plugin failed when a special (unkown) portal is loaded in link-zoom. 
 
> just printing the entire source also isn't very helpful as we already have the plugin nam 
 
as said, I don't see any plugin name here.  It only logs the code of error function.

Aradiv:
Okay so logging the plugin name can actually be helpfull but i still don't see how logging the entire source is helpful.

Especially those unknown reasons are the point where the debugger is very helpful as you can inspect the entire callstack when the error occurs. Just enable "pause on exceptions" and you can go to the entire callstack with all the values.

McBen:
> logging the plugin name can actually be helpfull 

Logging the plugin and function name would be great. But I've no idea how to archieve this.
So this is my approach.

johnd0e:
> But I've no idea how to archieve this.

We have to change plugins api in order to achieve that.

core/code/boot.js
```diff
@@ -685,14 +685,14 @@ function prepPluginsToLoad() {
...
-    if (!setup) {
+  function safeSetup(setup) {
+    if (typeof setup !== 'function') {
       log.warn('plugin must provide setup function');
       return;
     }
     var info = setup.info;
     if (typeof info !== 'object') {
-      log.warn('plugin does not have proper wrapper:',setup);
+      log.warn('plugin does not have proper wrapper:', { function: setup, info: setup.info, source: setup.toString() });
```
johnd0e, line 695: BTW, this is not needed in Chrome, as it already shows function test, and allows to jump to `setup` definition.

core/code/hooks.js
```diff
@@ -74,7 +74,7 @@ window.runHooks = function(event, data) {
         return false;  //break from $.each
       }
     } catch (e) {
-      log.error('error running hook '+event+', error: '+e);
+      log.error('error running hook', { event: event, error: e, source: callback.toString(), data: data });
```
johnd0e: Have you tried that in IITCm?
I suspect that it is unable to show objects in console.

core/code/boot.js
McBen: My main system is Firefox so I didn't recognized that before.

see https://github.com/IITC-CE/ingress-intel-total-conversion/pull/385#issuecomment-676552427

core/code/hooks.js
McBen: It will work in IITCm. It won't work in IITCm console window.

Issue #217, opened 2019-07-07
Adopt Leaflet.Evented instead of current hooks system.


johnd0e:
IITC currently is pretty integrated with Leaflet.
And Leaflet event system is powerfull, and definitely more comprehent than IITC custom homebrew hooks.

But then we need some additional compatibility layer to support old plugins.

johnd0e:
1. Currently we already have entities represented as leaflet objects : `portal`, `link`, `field`.
   They all are inherited from `L.Evented`, and able to fire/process events:
   - either themselves
     

     <details>
     <summary>Details</summary>

     Events: `add`, `remove`, `move`, and also tooltip/popup events.
     Arbitrary extra events can be added for iitc-specific purposes (`select`,`detailsLoad`,`detailsUpdate`)

     </details>

     (Note: ATM individual portals are not consistestent: #325)
   - or propagate events to parents (`map` object, or any `L.FeatureGroup`)
     <details><summary>Details</summary>
      Events: `layeradd`, `layerremove`.
     </details>

2. We also need `iitc` global object to provide general events.
   This object also may become a new base for whole iitc API (instead of polluting `window`)
   See https://github.com/IITC-CE/ingress-intel-total-conversion/issues/108#issuecomment-509034195

johnd0e:
Leaflet event system may be not directly applicable:
- There is no arguments control, so such errors will not be reported:
  - Callback must be a function
  - Callback wasn\'t registered for this event.
- When firing event - handler call is not wrapped in `try/catch`

But as other part of our code also heavily use Leaflet events - perhaps we could to make changes to `L.Evented`.

johnd0e:
> * There is no arguments control, so such errors will not be reported:

Can be fixed in https://github.com/Leaflet/Leaflet/pull/7518

> * When firing event - handler call is not wrapped in `try/catch`

We need some ideas how to refactor [`fire`](https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/core/Events.js#L170-L203) to be able to add `try`/`catch` there in compatible way.

johnd0e:
One more thing that is not supported by Leaflet: interruption of `runHooks`:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/c9d6b98f4cd62b64b1eeea85233fc59f34dcf954/core/code/hooks.js#L67-L81

[...]

PR #387, opened 2020-08-25
Safe removal of hooks

McBen:
### Problem:
if a hook is removed in an event handler the 'runHook' loop get out of sync.

Example error code:
```
addHook("portalDetailsUpdated",  myHook);  
addHook("portalDetailsUpdated",  anotherHook); // this one will be skipped and runHook will try to call "undefined"

myHook = (e) => {
  removeHook("portalDetailsUpdated",  myHook);
}
```

### (possible) Solution
Create a copy of the hook-array.
The old runHook loop will stick with the old array.
Show a warning for developers because this will be still a problem in old iitc versions.

### DIFF
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -63,11 +63,13 @@
 
 window._hooks = {}
 window.VALID_HOOKS = []; // stub for compatibility
+var runningHook;
 
 window.runHooks = function(event, data) {
   if(!_hooks[event]) return true;
   var interrupted = false;
-  $.each(_hooks[event], function(ind, callback) {
+  runningHook = event;
+  $.each(_hooks[event], function (ind, callback) {
     try {
       if (callback(data) === false) {
         interrupted = true;
@@ -77,6 +79,8 @@ window.runHooks = function(event, data) {
       log.error('error running hook '+event+', error: '+e);
     }
   });
+
+  runningHook = undefined;
   return !interrupted;
 }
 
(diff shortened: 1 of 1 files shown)


johnd0e:
> Create a copy of the hook-array.

Seems reasonable.

And in future we are going to deprecate old homebrew hooks implementation in favor of Leaflet events. **Edit**: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/217#issuecomment-800317584

cloudkucooland:
@johnd0e Thank god.

johnd0e:
> in favor of Leaflet events.

Leaflet event system may be not directly applicable:
- There is no arguments control, so such errors will not be reported:
  - Callback must be a function
  - Callback wasn\'t registered for this event.
- When firing event - handler call is not wrapped in `try/catch`

johnd0e:
> Create a copy of the hook-array.
> The old runHook loop will stick with the old array.

I've inspected [similar functionality of Leaflet](https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/core/Events.js#L147-L168), and found that it is slightly reliable, as it covers 2 more cases:
- `runHook` may be nested
- it's safer to suppress removed hook handler with noop to prevent it's calling in some edge cases.

See: #462.

> Show a warning for developers because this will be still a problem in old iitc versions.

Well, I do not feel that anybody should care old iitc versions.

PR #388, opened 2020-08-26
Uniques 0.4.0

MysticJay:
New Features
- Uniques with "droned" and "scouted"
- Im-/Export (Backup/Restore) to json-file
- PortalList to manage manually
- highlighters for scouted and droned
- Comm and Alerts analyzed for messages indicating unique actions
- Backlog for portals mentioned in COMM
- Using "cache viewed portals on map" to reduce backlog.


MysticJay:
Please bump version to 0.4.0 on pull

plugins/uniques.js
```diff
@@ -13,14 +12,18 @@ window.plugin.uniques.SYNC_DELAY = 5000;
 
 // maps the JS property names to localStorage keys
 window.plugin.uniques.FIELDS = {
-	'uniques': 'plugin-uniques-data',
-	'updateQueue': 'plugin-uniques-data-queue',
-	'updatingQueue': 'plugin-uniques-data-updating-queue',
+  'uniques':       'plugin-uniques-data',
```
modos189: Please do not change indents in old code that does not affect your PR

plugins/uniques.js
```diff
@@ -30,543 +33,878 @@ window.plugin.uniques.contentHTML = null;
...
-		&& (markup[2][1].plain == ' neutralized by ' || markup[2][1].plain == ' is under attack by ')
-		&& markup[3][0] == 'PLAYER') {
-		    // search for "Your Portal x neutralized by y"
-		    // search for "Your Portal x is under attack by y"
-			var portal = markup[1][1];
-			var guid = window.findPortalGuidByPositionE6(portal.latE6, portal.lngE6);
-			if(guid) plugin.uniques.setPortalVisited(guid);
-		}
-	});
+  var nick = window.PLAYER.nickname;
```
modos189: Because of the indentation changes here, I can't see where the old code is and where the lines you changed are. In interface it looks like one huge change for whole function

modos189: The same applies to almost the entire file.

modos189 (changes requested):

plugins/uniques.js
MysticJay: I was previously asked to take care of 2-spaces indentation and not use tabs. So the code was re-formated in all sections/functions that received new or changed code.

plugins/uniques.js
MysticJay: I was previously asked to take care of 2-spaces indentation and not use tabs. So the code was re-formated in all sections/functions that received new or changed code.

The atomic commits should help to follow up on the changes.
Shall I provide a DIFF that indicate the changes?

MysticJay: Nothing to change

plugins/cache-portals-on-map.js
```diff
@@ -29,38 +29,37 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
     }
   }
   return guid;
-}
+};
+
+window.cachePortalDetailsOnMap.searchInit = function () {
+    window.cachePortalDetailsOnMap.findPortalGuidByPositionE6old = window.findPortalGuidByPositionE6;
+    window.findPortalGuidByPositionE6 = window.cachePortalDetailsOnMap.findPortalGuidByCacheE6;
+};
```
johnd0e: Why do we ever need separate function for 2-lines code?

plugins/cache-portals-on-map.js
```diff
@@ -29,38 +29,37 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
-    window.plugin.cachePortalDetailsOnMap.findPortalGuidByPositionE6old = window.findPortalGuidByPositionE6;
-    window.findPortalGuidByPositionE6 = window.plugin.cachePortalDetailsOnMap.findPortalGuidByCacheE6;
-    console.log ("findPortalGuidByCacheE6 code injected");
-}
-/****************************************************************************************/
 window.plugin.cachePortalDetailsOnMap.setup  = function() {
 
   window.plugin.cachePortalDetailsOnMap.cache = {};
-  window.plugin.cachePortalDetailsOnMap.searchInit ();
+  window.plugin.cachePortaldetailsOnMap.searchInit;
```
johnd0e: Seriously?

plugins/cache-portals-on-map.js
```diff
@@ -29,38 +29,37 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
       }
     }
   }
   return guid;
-}
+};
+
+window.cachePortalDetailsOnMap.searchInit = function () {
+    window.cachePortalDetailsOnMap.findPortalGuidByPositionE6old = window.findPortalGuidByPositionE6;
+    window.findPortalGuidByPositionE6 = window.cachePortalDetailsOnMap.findPortalGuidByCacheE6;
```
johnd0e: Such hijacking is what we should avoid, especially in official plugins.

If you still need that badly - consider providing such possibility in some clean way (hooks, etc)

core/code/hooks.js
```diff
@@ -50,8 +50,8 @@
 //              displayed. The data hash contains both the unprocessed
 //              raw ajax response as well as the processed chat data
 //              that is going to be used for display.
-//     WARNING: this hook was disabled in earlier versions
-//              sideeffects possible but unknown. Use with care!
+//              This hook was implemented in earlier versions
+//              but commented out and never used before.
```
johnd0e: You better remove this comment completely. It is not helpful at all.

In general, providing new API (such as hook) is a big responsibility: you must to make sure that you are providing some functionality in best (and complete) way. Because you will not be able to fix this after it is already used in dozens of plugins.
So are you sure? If you are - good. If not - you should not uncomment that part prematurely.

plugins/cache-portals-on-map.js
MysticJay: Hooks cause all registered functions to execute. 
The idea here is to find a guid. The processing should stop as soon as ANY of the additional plugins was successful.

CachePortalsOnMap used to replace the core function, only extending it.
Offle-Plugin did the same but there was no way to get the benefit of both.
If the core function would change the plugins would both need to adjust.

If there is a better way, which does not cause useless CPU usage please commit.

plugins/cache-portals-on-map.js
MysticJay: Certainly not. Honestly I do not know how that got lost, my local backup has that brackets. 
Apart from that  I have no idea why this commit is in this PR. It has already been merged

core/code/hooks.js
MysticJay: We need that hook learn about the MSGs in ALERScomm.
It has been set up in the same way as for the ALLcomm.
This comment was already changed on request of @modos189.
Please agree on what you prefer and edit accordingly.

plugins/cache-portals-on-map.js
MysticJay: What is the problem?
There are way more functions that only consist of ONE line.
But I see that most plugins have their init-routines separated from setup().

plugins/cache-portals-on-map.js
MysticJay: After an hour of debugging with @McBen (thank you) we have likely identified the cause.
The file was added two different branches and submitted in two PRs.
_This_ PR has the corrected version (see commit "0.3.9beta1" ad88cca).
It seems, that after the first PR was merged the file has a more recent timestamp than the file in _this_ PR.

An other thing puzzles us: 
grafik
This PR tells there is only one file with changes, but down here it offers two to review.
also this file is marked as outdated:
grafik

Whatever goes wrong here, but do not include "cache-portals-on-map.js" in the merge.

McBen: the bug is on master - not here
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/3102c5c5cc47f9fd2e7933e633a99ed22f7c4117/plugins/cache-portals-on-map.js#L62

plugins/uniques.js
johnd0e: I suggest you to preserve existing whitespace style otherwise it gets messed and it's hard to review the actual changes.
If we need to change whitespace - we will do it at once in the whole file.

plugins/cache-portals-on-map.js
johnd0e: > What is the problem?

It's hard to read such spaghetti code. 

> There are way more functions that only consist of ONE line.
> But I see that most plugins have their init-routines separated from setup().

You see legacy code, so no surprise that there are a lot of bad patterns.

You should not do like this, in general.

plugins/cache-portals-on-map.js
johnd0e: @modos189 
Perhaps last commit should be reverted https://github.com/IITC-CE/ingress-intel-total-conversion/pull/388/commits/412fac2959f86cd789941f7354c2d4ff5d26f81b

plugins/cache-portals-on-map.js
johnd0e: > The idea here is to find a guid. The processing should stop as soon as ANY of the additional plugins was successful.

Whatever. The API should be clean.
So even in this case I prefer hook instead of hijacking, difference in CPU will be miserable anyway.

But if you care than we can extend hooks api here.

core/code/hooks.js
johnd0e: > We need that hook learn about the MSGs in ALERScomm

No problem - implement whatever you need. Just take enough time to make sure that it is implemented in best way, as already said.
And after you will be sure - those stupid comment will not be needed.

modos189:
Reverted my commit with a merge from master

plugins/uniques.js
MysticJay: As the nesting was changed the code unindented. Do the comparison with "ignore white space" and you get what you want.

johnd0e: Well, I do not see any indent change in final diff, so nothing to discuss anymore

MysticJay:
The complete branch has been rebuild.
conflicting files were removed.
and indentation reverted.

Closing this PR to avoid further conflicts

PR #389, opened 2020-08-26
Minor fixes

### LINKED ISSUE #334: Sync's method for G-authorize causes conflicts
(Forwarded from TG)
Original reporter: CharlieNasa, deviousness
The sync plugin was changed from using gapi.auth2.authorize() to gapi.auth2.init(), which pollutes the namespace and makes it so that it conflicts with any other plugin using gapi.auth2.* ; Had it stayed with gapi.auth2.authorize() it wouldn't conflict. 

https://developers.google.com/identity/sign-in/web/reference#gapiauth2authorizeparams_callback

"Warning: do not use this method alongside the recommended gapi.auth2.init and signIn flow. These are two distinct behaviors (Authorization for gapi.auth2.authorize vs Authentication for gapi.auth2.init/signIn) and will have unexpected issues if used within the same application."

Does anyone recall why this change was made? Does anyone have a working solution for multiple plugins using google auth?

MysticJay:
Small fixes and extentions to several plugins and core modules.
see commits for details

### DIFF
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -45,6 +45,13 @@
 //              displayed. The data hash contains both the unprocessed
 //              raw ajax response as well as the processed chat data
 //              that is going to be used for display.
+// alertsChatDataAvailable: this hook runs after data for the alerts
+//              chat has been received and processed, but not yet been
+//              displayed. The data hash contains both the unprocessed
+//              raw ajax response as well as the processed chat data
+//              that is going to be used for display.
+//     WARNING: this hook was disabled in earlier versions
+//              sideeffects possible but unknown. Use with care!
 // requestFinished: DEPRECATED: best to use mapDataRefreshEnd instead
 //              called after each map data request finished. Argument is
 //              {success: boolean} indicated the request success or fail.
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -50,8 +50,8 @@
 //              displayed. The data hash contains both the unprocessed
 //              raw ajax response as well as the processed chat data
 //              that is going to be used for display.
-//     WARNING: this hook was disabled in earlier versions
-//              sideeffects possible but unknown. Use with care!
+//              This hook was implemented in earlier versions
+//              but commented out and never used before.
 // requestFinished: DEPRECATED: best to use mapDataRefreshEnd instead
 //              called after each map data request finished. Argument is
 //              {success: boolean} indicated the request success or fail.
--- a/plugins/cache-portals-on-map.js
+++ b/plugins/cache-portals-on-map.js
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
   }
   data.callback(ents);
 };
+/****************************************************************************************/
+/* extend portal's GUID lookup to search in cached portals as well                      */
 
+window.cachePortalDetailsOnMap.findPortalGuidByCacheE6 = function(latE6, lngE6) {
+  let guid = window.cachePortalDetailsOnMap.findPortalGuidByPositionE6old(latE6, lngE6);
+  //PLUGIN.log("findPortalGuidByPositionE6old %s %s -> %s", latE6, lngE6, guid);
+
+  // PLUGIN: cache > cache
+  if (guid == null) {
+    const cache = window.plugin.cachePortalDetailsOnMap;
+    if (cache && cache.cache) {
+      for (let g in cache.cache) {
+        let p = cache.cache[g];
+        if (latE6 == p.ent[2][2] && lngE6 == p.ent[2][3]) {
+          PLUGIN.log("newFindPortalGuidByPositionE6 matched in cache! %s %s -> [%s] = %o", latE6, lngE6, g, p);
+          guid = g;
+          break;
+        }
+      }
+    }
+  }
+/*  // PLUGIN: (external) > offle
+  if (guid == null) {
+    const offle = window.plugin.offle;
+    if (offle && offle.portalDb) {
+      let lat = parseInt(latE6) / 1000000;
+      let lng = parseInt(lngE6) / 1000000;
+      //PLUGIN.log("lat/lng parsed as", lat, lng);
+      for (let g in offle.portalDb) {
+        let p = offle.portalDb[g];
+        if (p.lat == lat && p.lng == lng) {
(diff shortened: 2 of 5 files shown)


core/code/hooks.js
```diff
@@ -45,6 +45,13 @@
 //              displayed. The data hash contains both the unprocessed
 //              raw ajax response as well as the processed chat data
 //              that is going to be used for display.
+// alertsChatDataAvailable: this hook runs after data for the alerts
+//              chat has been received and processed, but not yet been
+//              displayed. The data hash contains both the unprocessed
+//              raw ajax response as well as the processed chat data
+//              that is going to be used for display.
+//     WARNING: this hook was disabled in earlier versions
```
modos189: This hook was originally turned off, so this is a new hook
https://github.com/MysticJay/ingress-intel-total-conversion/commit/0dcada2c3bd93bce9946bb34df35f114fdb98808#diff-f8a66be5793d55e0c4f7b7ad9b96e90b70848372de4a887f45874d7317eed77cR268

plugins/cache-portals-on-map.js
```diff
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
   }
   data.callback(ents);
 };
+/****************************************************************************************/
+/* extend portal's GUID lookup to search in cached portals as well                      */
```
modos189:
```suggestion
// extend portal's GUID lookup to search in cached portals as well
```

plugins/cache-portals-on-map.js
```diff
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
 
+window.cachePortalDetailsOnMap.findPortalGuidByCacheE6 = function(latE6, lngE6) {
+  let guid = window.cachePortalDetailsOnMap.findPortalGuidByPositionE6old(latE6, lngE6);
+  //PLUGIN.log("findPortalGuidByPositionE6old %s %s -> %s", latE6, lngE6, guid);
```
modos189: Let's delete this line, which is no longer needed

plugins/cache-portals-on-map.js
```diff
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
+        }
+      }
+    }
+*/
+  }
+  return guid;
+}
+
+window.cachePortalDetailsOnMap.searchInit = function () {
+//  if (!window.cachePortalDetailsOnMap.findPortalGuidByCacheE6) {
```
modos189: Do we still need this comment? Maybe it's better to delete it?

plugins/cache-portals-on-map.js
```diff
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
+      for (let g in offle.portalDb) {
+        let p = offle.portalDb[g];
+        if (p.lat == lat && p.lng == lng) {
+          PLUGIN.log("newFindPortalGuidByPositionE6 matched in offle! %s %s -> [%s] = %o", latE6, lngE6, g, p);
+          guid = g;
+          break;
+        }
+      }
+    }
+*/
```
modos189: Please explain what this code block is for. Will it be needed in the future? Or it's better to delete it?

plugins/cache-portals-on-map.js
```diff
@@ -29,11 +29,60 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
+  //PLUGIN.log("findPortalGuidByPositionE6old %s %s -> %s", latE6, lngE6, guid);
+
+  // PLUGIN: cache > cache
+  if (guid == null) {
+    const cache = window.plugin.cachePortalDetailsOnMap;
+    if (cache && cache.cache) {
+      for (let g in cache.cache) {
+        let p = cache.cache[g];
+        if (latE6 == p.ent[2][2] && lngE6 == p.ent[2][3]) {
+          PLUGIN.log("newFindPortalGuidByPositionE6 matched in cache! %s %s -> [%s] = %o", latE6, lngE6, g, p);
```
modos189: » error: ReferenceError: PLUGIN is not defined
🤔
Anyway, it seems that this code will be called very often with a large number of portals?
I suggest replacing it with
`// console.log("newFindPortalGuidByPositionE6 matched in cache! %s %s -> [%s] = %o", latE6, lngE6, g, p);`

modos189 (changes requested):
Hi. Sent some questions and suggestions in comments

core/code/hooks.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
MysticJay: CR implemented

plugins/cache-portals-on-map.js
```diff
@@ -30,10 +30,36 @@ window.plugin.cachePortalDetailsOnMap.entityInject = function(data) {
...
+        if (latE6 == p.ent[2][2] && lngE6 == p.ent[2][3]) {
+//        console.log("newFindPortalGuidByPositionE6 matched in cache! %s %s -> [%s] = %o", latE6, lngE6, g, p);
+          guid = g;
+          break;
+        }
+      }
+    }
+  }
+  return guid;
+};
```
johnd0e, line 52: This improvement has a very little sense, because "Cache viewed portals on map" is extremely limited and never keeps much data in it's cache. Also, `findPortalGuidByPositionE6` hijacking looks like hack: just imaging - there are other caching plugins in the wild, and what if every of them would hijack that single function.

If we really need this functionality - than we should better implement via some clean hook.

PR #393, opened 2020-09-13
Refactor linkDetails a bit

### LINKED ISSUE #391: Unable to share portal
I've discovered that I'm not able to share some portals, while others work just fine.

I've found a couple and developed a theory since I found a thing they had in common. The ones that don't work contain the special letter "å" or "Å" (in caps). The letter is used in the Danish language, others are "æ/Æ" and "ø/Ø", however those work just fine. 

When I try to share the portal location, using the normal button from the portal info screen, nothing happens. I can see iitc register me clicking on the button, and normally there would open a menu on how or where to I would share that portal. Well with these, nothing happens. I can click the share button again, I've tried to wait and give it a minutes, but nothing. If I go back to the map and click on another portal, try and share that, it works as normal. If I then go back to the first portal again, still nothing.

This is one of those I've found, and containing the special letter which I suspect might be causing the problem, or at least have some influence on the problem:

https://intel.ingress.com/intel?ll=55.3366,10.331674&z=16&pll=55.33672,10.330928

johnd0e:
Fix #391

### DIFF
--- a/core/code/portal_detail_display.js
+++ b/core/code/portal_detail_display.js
@@ -78,27 +78,30 @@ window.renderPortalDetails = function(guid) {
   levelDetails = "Level " + levelDetails;
 
 
-  var linkDetails = [];
+  var linkDetails = $('<div>', { class: 'linkdetails' });
 
-  var posOnClick = 'window.showPortalPosLinks('+lat+','+lng+',\''+escapeJavascriptString(title)+'\')';
-  var permalinkUrl = window.makePermalink([lat,lng]);
+  var posOnClick = window.showPortalPosLinks.bind(this,lat,lng,title);
 
   if (typeof android !== 'undefined' && android && android.intentPosLink) {
     // android devices. one share link option - and the android app provides an interface to share the URL,
     // share as a geo: intent (navigation via google maps), etc
 
-    var shareLink = $('<div>').html( $('<a>').attr({onclick:posOnClick}).text('Share portal') ).html();
-    linkDetails.push('<aside>'+shareLink+'</aside>');
+    var shareLink = $('<a>').text('Share portal').click(posOnClick);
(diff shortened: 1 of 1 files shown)


core/code/portal_detail_display.js
```diff
@@ -78,27 +78,30 @@ window.renderPortalDetails = function(guid) {
...
-    var shareLink = $('<div>').html( $('<a>').attr({onclick:posOnClick}).text('Share portal') ).html();
-    linkDetails.push('<aside>'+shareLink+'</aside>');
+    var shareLink = $('<a>').text('Share portal').click(posOnClick);
+    linkDetails.append($('<aside>').append($('<div>').append(shareLink)));
 
   } else {
     // non-android - a permalink for the portal
-    var permaHtml = $('<div>').html( $('<a>').attr({href:permalinkUrl, title:'Create a URL link to this portal'}).text('Portal link') ).html();
-    linkDetails.push ( '<aside>'+permaHtml+'</aside>' );
+    var permaHtml = $('<a>').attr({
```
McBen, line 94: Note: you can even skip the "attr" here, like:
`$("<a>", {href: ...; title: "..."; text: "..."; click: ...})`

johnd0e: > Note: you can even skip the "attr" here, like: [...]

Yeah, but (imho) in that case readability would be worse.

McBen:
```suggestion
    var permaHtml = $('<a>', {
             text: 'Portal link',
             title: 'Create a URL link to this portal'},
             href: window.makePermalink([lat,lng])
         });
```

maybe

johnd0e: Yeah, may be. Matter of taste.

modos189 (approved):

PR #397, opened 2020-10-05
plugins: use Date/toLocaleString instead of own functions

le-jeu:
- `score-cycle-times`: for cycle times
  Settings:
  + `window.plugin.scoreCycleTimes.locale`
  + `window.plugin.scoreCycleTimes.dateTimeFormat`
- `player-activity-tracker`: time in search
  Settings:
  + `window.plugin.playerTracker.locale`
  + `window.plugin.playerTracker.dateTimeFormat`

Note: started here: https://github.com/IITC-CE/ingress-intel-total-conversion/commit/e92dd96334bc490a759876ed6b2aed739f8f365a#commitcomment-42962123

### DIFF
--- a/plugins/score-cycle-times.js
+++ b/plugins/score-cycle-times.js
@@ -10,7 +10,7 @@ window.plugin.scoreCycleTimes = function() {};
 
 window.plugin.scoreCycleTimes.CHECKPOINT = 5*60*60; //5 hours per checkpoint
 window.plugin.scoreCycleTimes.CYCLE = 7*25*60*60; //7 25 hour 'days' per cycle
-
+window.plugin.scoreCycleTimes.localeTime = 'default';
 
 window.plugin.scoreCycleTimes.setup  = function() {
 
@@ -40,11 +40,13 @@ window.plugin.scoreCycleTimes.update = function() {
   var checkpointStart = Math.floor(now / (window.plugin.scoreCycleTimes.CHECKPOINT*1000)) * (window.plugin.scoreCycleTimes.CHECKPOINT*1000);
   var checkpointEnd = checkpointStart + window.plugin.scoreCycleTimes.CHECKPOINT*1000;
 
+  var o = new Intl.DateTimeFormat(window.plugin.scoreCycleTimes.localeTime, {
+    year: 'numeric', month: 'numeric', day: 'numeric',
+    hour: 'numeric',
+  });
 
   var formatRow = function(label,time) {
-    var timeStr = unixTimeToString(time,true);
-    timeStr = timeStr.replace(/:00$/,''); //FIXME: doesn't remove seconds from AM/PM formatted dates
-
+    var timeStr = o.format(new Date(time));
     return '<tr><td>'+label+'</td><td>'+timeStr+'</td></tr>';
   };
 
--- a/plugins/score-cycle-times.js
+++ b/plugins/score-cycle-times.js
@@ -42,8 +42,8 @@ window.plugin.scoreCycleTimes.update = function() {
 
 
   var formatRow = function(label,time) {
-    var timeStr = unixTimeToDateTimeString(time,false);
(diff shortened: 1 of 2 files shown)


johnd0e:
Well, in fact there are more differences:

Compare:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/3f4867a4a4ab5dd86d093fcaa9829f3267f70e9b/core/code/utils_misc.js#L184-L186
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/3f4867a4a4ab5dd86d093fcaa9829f3267f70e9b/core/code/utils_misc.js#L199-L201

le-jeu:
in this context, the date is always present `full=true`, so both are returning something like `2020-10-05 <time with seconds>`
First returns `2020-10-05 7:00:00 AM` or `2020-10-05 07:00:00` depending on locale
Second always returns `2020-10-05 07:00:00`, thus removing seconds always works

johnd0e:
Have you closed this intentionally?

le-jeu:
nope, I was cleaning up u_u

johnd0e:
Tempting solution, but we should adopt some consistent approach.
Related: #458

johnd0e:
I believe that we should use it in the way that allow to apply global options overriding.
E.g. with `window.localeOptions`:
```js
  window.locale = navigator.languages; // can be overriden
  window.localeOptions = {
    timeZone: 'Etc/GMT+3',
    hourCycle: 'h24',
  };
// ...........
  var options = L.extend({}, window.localeOptions, {
    year: 'numeric', month: 'numeric', day: 'numeric',
    hour: 'numeric',
  });
  var o = new Intl.DateTimeFormat(window.locale, options);
```

plugins/score-cycle-times.js
```diff
@@ -10,7 +10,7 @@ window.plugin.scoreCycleTimes = function() {};
 
 window.plugin.scoreCycleTimes.CHECKPOINT = 5*60*60; //5 hours per checkpoint
 window.plugin.scoreCycleTimes.CYCLE = 7*25*60*60; //7 25 hour 'days' per cycle
-
+window.plugin.scoreCycleTimes.localeTime = 'default';
```
johnd0e: With default value I get such format:
```
3/17/2021, 5 AM
3/18/2021, 4 PM
3/18/2021, 9 PM
3/24/2021, 12 PM
```
That is not good, as I have different settings in my OS.
And if I force value that to custom I still get hours only (may be correct but still unexpected):
```
17.03.2021, 05
18.03.2021, 16
18.03.2021, 21
24.03.2021, 12
```

P.S.
You can also see that AM/PM format looks not aligned (both time and date),

plugins/score-cycle-times.js
```diff
@@ -40,11 +40,13 @@ window.plugin.scoreCycleTimes.update = function() {
   var checkpointStart = Math.floor(now / (window.plugin.scoreCycleTimes.CHECKPOINT*1000)) * (window.plugin.scoreCycleTimes.CHECKPOINT*1000);
   var checkpointEnd = checkpointStart + window.plugin.scoreCycleTimes.CHECKPOINT*1000;
 
+  var o = new Intl.DateTimeFormat(window.plugin.scoreCycleTimes.localeTime, {
+    year: 'numeric', month: 'numeric', day: 'numeric',
+    hour: 'numeric',
```
johnd0e:
```suggestion
    year: '2-digit', month: '2-digit', day: '2-digit',
    hour: '2-digit', minute: '2-digit'
```

plugins/score-cycle-times.js
le-jeu: my point about user tastes, dev tastes and locales. Locales add complexity, illustrated by this small PR...
hour only is expected, we don't need minutes in case of cycle time... halfhour timezone... right I forgot about them

modos189 (approved):

modos189 (approved):

PR #400, opened 2020-10-26
region_scoreboard: fix daylightsaving hour

### LINKED ISSUE #399: Daylight-Saving in regionscore
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/3f4867a4a4ab5dd86d093fcaa9829f3267f70e9b/core/code/region_scoreboard.js#L124-L128

Here we skip daylight saving.

In a cycle which includes a clock-change the CP times (after clock-change) will be wrong.

McBen:
Fix #399


johnd0e:
Is it single place in IITC where such issue may occur?

PR #404, opened 2020-11-23
Redirect to https if user still uses old http

McBen:
Force HTTPS protocol.

There are still users out there using the outdated http page (or maybe scripts generating old links?)
Additional Nia removed the redirect from http to https awhile ago.

### DIFF
--- a/buildsettings.py
+++ b/buildsettings.py
@@ -22,7 +22,7 @@ defaults = {  # common for all build targets
 
     # these settings should not be touched unless intel url(s) changed
     'url_intel_base': 'https://intel.ingress.com/',
-    'match': 'https://intel.ingress.com/*',
+    'match': '*://intel.ingress.com/*',
 
     'plugin_wrapper': 'pluginwrapper', # use wrapper from pluginwrapper.py
 
--- a/core/total-conversion-build.js
+++ b/core/total-conversion-build.js
@@ -13,6 +13,11 @@ if (document.documentElement.getAttribute('itemscope') !== null) {
 }
 window.iitcBuildDate = '@build_date@';
 
+// force https
+if (location.protocol !== 'https:') {
+  location.replace('https:' + location.href.substring(location.protocol.length));
+}
+
 // disable vanilla JS
 window.onload = function() {};
 document.body.onload = function() {};


MysticJay:
@McBen Didn't we try that already and we ended up in mess when loging in to google
(redirect to wrong place)?

IMHO as the @match is only allowing https://intel.ingress.com/* IITC should simply not launch in that case anyway.

McBen:
> @McBen Didn't we try that already and we ended up in mess when loging in to google
> (redirect to wrong place)?

Did we? I only found this one: fe2d240eb77bf7f05de45a73c93cd33e29ff2073
But Niantic still (or better: again) allows http access.

> IMHO as the @match is only allowing https://intel.ingress.com/* IITC should simply not launch in that case anyway.

this is included in this PR


The problem is that links like "http://intel.ingress.com/intel?pll=52.522756,13.402208" are still available or shared.
The truth is: I don't like the redirect as well but I prefer this workaround before a customer crying "my iitc won't work anymore"
In the end it's up to you to decide if this is a valid feature or not.

MysticJay:
@McBen "allows http (again)" or "redirects http itself"?

But isn't the root cause legacy tools and plugins that still have not been updated (for years)? Where do we draw the line? This one looks easy, but the next could be way more complex having side effects. 

We can not move forward if we keep looking back.
IMHO users should rather complain to those maintainers/authors or find new ones to take responsabilty.

McBen:
It's not about a legacy or dropped technology. It's about a valid intel address not supported by iitc-ce.

This workaround should help people accidently got stucked on the other "not so good" page. The other page iitc-ce currently ignores but niantic still provides.

buildsettings.py
```diff
@@ -22,7 +22,7 @@
 
     # these settings should not be touched unless intel url(s) changed
     'url_intel_base': 'https://intel.ingress.com/',
-    'match': 'https://intel.ingress.com/*',
+    'match': '*://intel.ingress.com/*',
```
johnd0e, line 25: If I understand correctly - we need this change only for main script.

McBen: Sure. But ATM there are no separated main & plugin settings and changing all doesn't hurt.

johnd0e: > changing all doesn't hurt.

Well, we know that standard plugins serve as samples and templates for 3rd-party, that's why I believe that it'd better to have single form - https.

Also I'm not sure that we should put that workaround into core. Maybe we should resurrect "force http" plugin instead.

And see https://github.com/IITC-CE/ingress-intel-total-conversion/commit/fe2d240eb77bf7f05de45a73c93cd33e29ff2073 - previously there was separate routine in mobile app, which also may be needed.

McBen: > Maybe we should resurrect "force http" plugin instead.

This would only help these people who might know they run into problems. And these people know how to fix it.

> And see [fe2d240](https://github.com/IITC-CE/ingress-intel-total-conversion/commit/fe2d240eb77bf7f05de45a73c93cd33e29ff2073) - previously there was separate routine in mobile app,

IDK how mobile reacts on http links. I thought it would be able to handle all registered web-links.

McBen: > Well, we know that standard plugins serve as samples and templates for 3rd-party, that's why I believe that it'd better to have single form - https.

why is it BAD to support http too? Niantic supports both.

johnd0e: > why is it BAD to support http too? Niantic supports both.

For example because:
1) it would lead us to old "bug" with inconsistent plugin data, because localStorage is not shared between different protocols.
2) http is vulnerable, and there is no single reason to prefer it in IITC
3) there is pretty big chance that tomorrow Niantic will forbid http again.

So we just shouldn't encourage "http" because nobody gain from it.

But anyway I agree that we should reimplement http->https redirection on our side.
Question is whether put it into core or plugin.

McBen: as said: as plugin you won't help anybody.

my PR was for people not knowing they use the wrong page.

PR #406, opened 2020-12-07
Draw tools merge

MysticJay:
Merge draws into one project


plugins/draw-tools.js
```diff
@@ -1,8 +1,8 @@
 // @author         breunigs
 // @name           Draw tools
 // @category       Draw
-// @version        0.8.1
-// @description    Allow drawing things onto the current map so you may plan your next move.
+// @version        0.8.0
```
johnd0e: Perhaps here again, last commit (merge) was not careful enough

plugins/draw-tools.js
```diff
@@ -365,7 +375,7 @@ window.plugin.drawTools.optCopy = function() {
       if (stockWarnings.unknown) stockWarnTexts.push('Warning: UNKNOWN ITEM TYPE');
 
       var html = '<p><a onclick="$(\'.ui-dialog-drawtoolsSet-copy textarea\').select();">Select all</a> and press CTRL+C to copy it.</p>'
-                +'<textarea readonly onclick="$(\'.ui-dialog-drawtoolsSet-copy textarea\').select();">'+localStorage[window.plugin.drawTools.KEY_STORAGE]+'</textarea>'
+                +'<textarea readonly onclick="$(\'.ui-dialog-drawtoolsSet-copy textarea\').select();">'+window.localStorage[window.plugin.drawTools.KEY_STORAGE]+'</textarea>'
```
johnd0e: I have such concerns about original code:
1. Too looong line.
2. Crazy nesting of quotes
3. Using `onclick` attribute instead of attaching listener in js.

Latter is not so easy to fix though.
But in my opinion, that should be done anyway to make draw tools menu extendable by other plugins.
As you remember, DrawToolsPlus used to rewrite major parts of original code just to add a couple items to that menu.

johnd0e: Btw, I see no sense in changing `localStorage` to `window.localStorage`. Former was easier to read.

plugins/draw-tools.js
```diff
@@ -383,12 +393,15 @@ window.plugin.drawTools.optCopy = function() {
 }
 
 window.plugin.drawTools.optExport = function() {
-  var data = localStorage[window.plugin.drawTools.KEY_STORAGE];
-  window.saveFile(data, 'IITC-drawn-items.json', 'application/json');
+  if (window.localStorage[window.plugin.drawTools.KEY_STORAGE] === '' ||
+      window.localStorage[window.plugin.drawTools.KEY_STORAGE] === undefined)
```
johnd0e: Again, all this mess is hard to read.
Why not write like this:
```js
  var drawTools = window.plugin.drawTools;
  if (localStorage[drawTools.KEY_STORAGE] === '' ||
      localStorage[drawTools.KEY_STORAGE] === undefined)
```

Or better this:
```js
  var drawTools = window.plugin.drawTools;
  if (!localStorage[drawTools.KEY_STORAGE])
```

plugins/draw-tools.js
```diff
@@ -502,7 +515,7 @@ window.plugin.drawTools.optImport = function() {
 window.plugin.drawTools.optReset = function() {
   var promptAction = confirm('All drawn items will be deleted. Are you sure?', '');
   if(promptAction) {
-    delete localStorage[window.plugin.drawTools.KEY_STORAGE];
+    localStorage[window.plugin.drawTools.KEY_STORAGE] = '[]';
```
johnd0e: Why?

modos189: Just like in upstream [draw-tools](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/plugins/draw-tools.js#L493).

This PR only replaces [`679939c` (#406)](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/406/commits/679939c461b61b2fc9701c7279de8a2df9d4c232) the key with a constant

johnd0e: Ok, that let's continue in origin: https://github.com/IITC-CE/ingress-intel-total-conversion/commit/940ff669aaf08e1bdac2b5c4cbcdd9e06f3cc1d1#r45236089

plugins/draw-tools.js
MysticJay: On the one hand you ask me not to change lines that do not belong to the commit, OTOH you ask me to change the way the code was written until now.

I try to follow they way the plugin was written until now. If we want to remove "window." and "window.plugin" we have to do that in one single commit, but not every now and then.

As KEY_STORAGE can be set to '' (empty string), the short variant is insufficient.

plugins/draw-tools.js
MysticJay: right and not.
the provided PRs are trying to be independant to keep them "one per feature".
As mentioned several times in the paste these two lines need to be updated manually depending on the selected order of merging.

plugins/draw-tools.js
johnd0e: > On the one hand you ask me not to change lines that do not belong to the commit,

Right. As I specified earlier - those concerns are about original code.
Just to make sure that we both (and all others) are sharing similar view.

As for changing this bad code: definitely there is no sense to change just 3 lines leaving the rest kilobytes in old state.
So yes, if you make some minor change - better not mix it with refactoring.
Still, this legacy requires refactoring. Just keep it separately (separate commits, or separate PR).

plugins/draw-tools.js
MysticJay: 1) add a line break
2) agree, line was not changed in that perspective
3) see 2)

My task was NOT code optimization, but to get rid of DTplus. So just adding it's code where it belongs.

"window.": just trying to follow the way the rest of the code was done. IMHO if you try to understand the code  and suddenly this is dropped you ask yourself "why did the DEV do so?" See my next comment below.

johnd0e: > "window.": just trying to follow the way the rest of the code was done.

1. But changing that line was not actually in need for your DTPlus, right?
2. Make all the code follow one style is good goal. Just keep it separately. And choose best patterns (here: `localStorage` without `window`)

johnd0e: > IMHO if you try to understand the code and suddenly this is dropped you ask yourself "why did the DEV do so?"

In fact if one tries to understand iitc code currently - he should prepare to real quest with all those legacy patterns everywhere.
Yeah, in future we should try to make it follow some common (and good!!) style.

plugins/draw-tools.js
johnd0e: > As KEY_STORAGE can be set to '' (empty string), the short variant is insufficient.

Why should we allow empty string?

plugins/draw-tools.js
MysticJay: The task is NOT to delete the KEY_STORAGE, but to provide an empty JSON-String (nothing drawn yet).
The value of KEY_STORAGE is changing depending on the selected MPE-Project. 
To be able select the active dataset transparently for the plugin, i.e. without the orginal plugin to know, which storage is currently in use. Only MPE is aware of the project names, not the plugin.

plugins/draw-tools.js
MysticJay: After checking the old code. These two lines are legacy code. I only added a linebreak for readability.
The newline itself was added in an earlier PR (likely MPE).

PR #407, opened 2020-12-07
Sync50a

### LINKED ISSUE #394: SYNC - Collision handling
SYNC has a general design issue:
let's assume we have two clients. 

Every 3 Minutes each client checks, if the file on the drive has changed (SYNC stores a UUID of the client that did the last update in the file). If so it will replace local data of the registered plugin (yes, there is no delta handling).

This is fine if you prepare your data at home and then go out to play and use IITCm (usecase).

The trouble comes, when you use both clients at the same time (within that 3 minutes window): 
every time a registered plugin has changed its local data it will trigger SYNC to update the file.
SYNC will simply replace it without prior checking anything.

So if you change one element on first client, then an other element on the second client, the second change will survive.

Many agents have lived with that problem without even noticing it, so we could just ignore it...
IMHO not a good idea, now that we know the issue and the cause.

Should be easy to fix you would guess, but it is not:

- SYNC does not know about the structure of the plugin's data. UNIQUES data is pretty flat, one key per portal visited, 10.000 keys in the file. Bookmarks is hierarchical with exactly two keys on top level: MAPS and PORTALS. Below we find folder-objects and only then the entry-objects. Moving a bookmark from one folder to an other is a complex action, causing many changes. Thus the easiest is to copy the whole data.
- The different clients might not be online or not connected to drive while the user changes entries.
- A client might get killed, reloaded or simply pushed to background and nothing will sync at all.

Making a professional solution with locks, semaphores, databases, delta handling, checks et.al might be a good practice for a programmer but also create a very complex and expensive (CPU) solution.

The current proposal is:
Step 1: we are able to identify a collision, so we will ask the user "remote, local or abort?"
Step 2: the plugin can provide a collisionHandler-function. As it knows the data structure it would be up to the plugin's author to implement what ever he thinks.

The proposal will keep 
SYNC easy, without the need to know anything about the data it handles (transparency).
the responsibility with the user (step 1) or
with the author of the plugin, while providing an alternative (step 2)

The change will not break any 3rd-party plugin, data or structure.

MysticJay:
SYNC collision handling

Summary: multiple active clients accessing the drive might cause IITC to "forget" recently done changes when polling the drive.

Added a collision detection that will ask the user with which data-set to continue.

Clean copy of PR #403
addressing Issue #394
Status: RC1

### DIFF
--- a/plugins/sync.js
+++ b/plugins/sync.js
@@ -98,7 +98,12 @@ window.plugin.sync.RegisteredMap = function(options) {
   this.initializing = false;
   this.initialized = false;
   this.failed = false;
-
+  this.doSync = 'remote';                                   // normal 'sync'
+                                                            // 'no' sync needed
+                                                            // force 'remote' (1st sync)
+                                                            // force 'local'
+                                                            // sync is on 'stop'
+  
   this.initialize = this.initialize.bind(this);
   this.loadDocument = this.loadDocument.bind(this);
 };
@@ -122,11 +127,6 @@ window.plugin.sync.RegisteredMap.prototype.updateMap = function(keyArray) {
   } finally {
     _this.dataStorage.saveFile(_this.prepareFileData());    // save data and UUID to drive
   }
-};
-
-window.plugin.sync.RegisteredMap.prototype.isUpdatedByOthers = function() {
-  var remoteUUID = this.lastUpdateUUID.toString();
-  return (remoteUUID !== '') && (remoteUUID !== this.uuid);
 };
 
 // concatenate a the filename for the drive "pluginName[fieldName]"
@@ -195,41 +195,97 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
     setTimeout(function() {_this.loadDocument();}, window.plugin.sync.checkInterval);
   };
 
-  // this function called when the document is loaded
-  // update local data if the document is updated by other (overwrite!)
-  // and adding a timer to further check for updates
+  // this function gets called when the document is loaded
   onFileLoaded = function(data) {
-    _this.map = data['map'];
-    _this.lastUpdateUUID = data['last-update-uuid'];
-
+    
     if (!_this.intervalID) {                        // if the DL-interval is not set
       _this.intervalID = setInterval(function() {   // create a new timer to load
         _this.loadDocument();                       // the file from drive
       }, window.plugin.sync.checkInterval);         // every <checkInterval> ms
     }
 
-    // Replace local value if data is changed by others
-    if(_this.isUpdatedByO\thers()) {
-      plugin.sync.logger.log(_this.getFileName(), 'Updated by others, replacing content from ';
+    /* Sync only checks the drive every 3 mins. But changes to local data get pushed
+       immediately. Two or more clients that do changes might might store data, that the 
+       other is not aware of. We are trying to identify crititcal changes. Following 
+       situations have been identified and should be coverd by the logic:
+       
+       Status-table:
+       0: init, first start, always load from remote
+       1: lastUUID and remoteUUID show the same UUID, load from remote
+       2: lastUUID and remoteUUID show local UUID, no load needed
+       3: lastUUID and remoteUUID show different, not local UUIDs, load remote
+       4: lastUUID is local and remoteUIID is different, CONFLICT! Stop sync and ask user
+    */
+    _this.remoteUUID = data['last-update-uuid'];    // get the UUID of the data on drive 
+// -----------------------------------------------------------------------------------------    
+    // the initial value for doSync is 'remote', so initially loading from drive 
+    if (_this.doSync == 'sync') {
+      let conflict = true;                                   // 
+      // check for conflict
+      if (_this.remoteUUID == _this.lastUpdateUUID) {       // last Update was done on drive
+        conflict = false;                                   // so there is not a conflict
+      }; 
+      if (_this.remoteUUID != _this.lastUpdateUUID &&       // 3rd client modified the drive
+          _this.lastUpdateUUID != _this.uuid) {
+        // maybe issue a warning
+        console.log ('SYNC detected a 3rd client that modified the drive');
+        conflict = false;                                   // but not a conflict here
+      };
+      
+      if (conflict) {                                       // some other client changed the
+        _this.doSync = 'stop';                              // data on the drive, after we
+                                                            // did a change. So we need
+        _this.conflictHandler();                            // to stop sync and ask the user
+        plugin.sync.logger.log(_this.getFileName(), 'SYNC detected a conflict caused by '+ _this.lastRemoteUUID); 
+      } else {                                              // ELSE no conflict found: 
+        if (_this.remoteUUID == _this.uuid) {               // local data still is on drive?
+          _this.doSync = 'sync';                            // no need to do anything
+        } else {                                            // otherwise
+          _this.doSync = 'remote';                          // load remote data 
+        };                                                  // (it might have changed)
+      };      
+    };
+// -----------------------------------------------------------------------------------------    
+    if (_this.doSync == 'remote') {
+      // load the remote data to the plugin's dataobject
+      _this.lastUpdateUUID = _this.remoteUUID;      // remember last client on drive
+      _this.map = data['map'];                      // map from the loaded data
+
       window.plugin[_this.pluginName][_this.fieldName] = {};// clear the plugin's dataobject
-      $.each(_this.map, function(key, value) {          // fill the dataobject with data from drive
+      $.each(_this.map, function(key, value) {      // fill the dataobject with data from drive
         window.plugin[_this.pluginName][_this.fieldName][key] = _this.map[key]; 
       });
-      if(_this.callback) _this.callback(                // execute the callback functions with
-       _this.pluginName,                                //    pluginName
-       _this.fieldName,                                 //    fieldName
-       null,                                            //    null (for backwards compatibility)
-       true                                             //    true = full file load from Drive
+
+      // execute the callback functions             
+      if(_this.callback) _this.callback(            // execute the callback functions with
+        _this.pluginName,                           //    pluginName
+        _this.fieldName,                            //    fieldName
+        null,                                       //    null (for backwards compatibility)
+        true                                        //    true = full file load from Drive
       );
-    }
-
-    _this.initialized = true;                       // connection to drive is "initalized"
-    _this.initializing = false;                     // remove semaphore
-    plugin.sync.logger.log(_this.getFileName(), 'Data loaded from: '); 
-                                                            // execute the callback functions
-
-    if(_this.callback) _this.callback();
-    if(_this.initializedCallback) _this.initializedCallback(_this.pluginName, _this.fieldName);
+    
+      if (_this.initializing) {
+        _this.initialized = true;                   // connection to drive is "initalized"
+        _this.initializing = false;                 // remove semaphore
+        if(_this.initializedCallback) _this.initializedCallback( 
+          _this.pluginName, 
+          _this.fieldName
+        );
+      }
+      plugin.sync.logger.log(_this.getFileName(), 'Data loaded from: '+ _this.lastUpdateUUID); 
+      _this.doSync = 'sync';                        // return to normal sync
+    };
+// -----------------------------------------------------------------------------------------    
+    if (_this.doSync == 'local') {                  // User decided for 'local' data.
+      _this.lastUpdateUUID = _this.uuid;            // 
+      _this.updateMap({});                          // push local data to drive
+      _this.doSync = 'sync';                        // resume normal sync
+    };
+// -----------------------------------------------------------------------------------------     
+    if (_this.doSync == 'stop') {
+      // do nothing
+      console.log('SYNC stopped synching data for plugin '+_this.pluginName);
+    };
   };
 
   // Stop the sync if any error occur and try to re-authorize
@@ -256,6 +312,50 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
   // finally execute readFile
   this.dataStorage.readFile(initializeFile, onFileLoaded, handleError);
 };
+
+// conflict handler to decide between remote or local data
+//   results: 'remote', 'local', 'stop'
+window.plugin.sync.RegisteredMap.prototype.conflictHandler = function () {
+  var _this;
+  _this = this;
+  
+  window.dialog ({
+    title: "SYNC Conflict Handling",
+    html: "<div>SYNC found a conflict. The remote data stored on the drive "
+        + "was changed while you did local changes . "
+        + "Please decide which data shall be used. <br>"
+        + "Your decission will overwrite the other changes. "
+        + "It is impossible to merge the data!</div><br>"
+        + "",
+    buttons: [
+      { text: "remote",
+        click: function() {
+          _this.doSync = 'remote';
+          $( this ).dialog( "close" );
+        }
+      },
+      { text: "local",
+        click: function() {
+          _this.doSync = 'local';
+          $( this ).dialog( "close" );
+        }
+      },
+      { text: "STOP SYNC",
+        click: function() {
+          _this.doSync = 'stop';
+          $( this ).dialog( "close" );
+        }
+      }
+    ],
+    closeCallback: function() {
+       _this.loadDocument(); 
+    },
+    maxHeight: 300
+  });
+  
+};
+
+
 //// end RegisteredMap
 
 
@@ -771,7 +871,7 @@ window.plugin.sync.loadLocal = function(mapping) {
                           : obj;
 };
 
-//load a previously generated UUUID (or generate a new one)
+//load a previously generated UUID (or generate a new one)
 window.plugin.sync.loadUUID = function() {
   plugin.sync.loadLocal(plugin.sync.KEY_UUID);
   if(!plugin.sync.uuid) {
--- a/plugins/sync.js
+++ b/plugins/sync.js
@@ -202,25 +202,32 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
     _this.map = data['map'];
     _this.lastUpdateUUID = data['last-update-uuid'];
 
-    if (!_this.intervalID) {
-      _this.intervalID = setInterval(function() {
-        _this.loadDocument();
-      }, window.plugin.sync.checkInterval);
+    if (!_this.intervalID) {                        // if the DL-interval is not set
+      _this.intervalID = setInterval(function() {   // create a new timer to load
+        _this.loadDocument();                       // the file from drive
+      }, window.plugin.sync.checkInterval);         // every <checkInterval> ms
     }
 
     // Replace local value if data is changed by others
-    if(_this.isUpdatedByOthers()) {
-      plugin.sync.logger.log(_this.getFileName(), 'Updated by others, replacing content');
-      window.plugin[_this.pluginName][_this.fieldName] = {};
-      $.each(_this.map, function(key, value) {
-        window.plugin[_this.pluginName][_this.fieldName][key] = _this.map[key];
+    if(_this.isUpdatedByO\thers()) {
+      plugin.sync.logger.log(_this.getFileName(), 'Updated by others, replacing content from ';
+      window.plugin[_this.pluginName][_this.fieldName] = {};// clear the plugin's dataobject
+      $.each(_this.map, function(key, value) {          // fill the dataobject with data from drive
+        window.plugin[_this.pluginName][_this.fieldName][key] = _this.map[key]; 
       });
-      if(_this.callback) _this.callback(_this.pluginName, _this.fieldName, null, true);
+      if(_this.callback) _this.callback(                // execute the callback functions with
+       _this.pluginName,                                //    pluginName
+       _this.fieldName,                                 //    fieldName
+       null,                                            //    null (for backwards compatibility)
+       true                                             //    true = full file load from Drive
+      );
     }
 
-    _this.initialized = true;
-    _this.initializing = false;
-    plugin.sync.logger.log(_this.getFileName(), 'Data loaded');
+    _this.initialized = true;                       // connection to drive is "initalized"
+    _this.initializing = false;                     // remove semaphore
+    plugin.sync.logger.log(_this.getFileName(), 'Data loaded from: '); 
+                                                            // execute the callback functions
+
     if(_this.callback) _this.callback();
     if(_this.initializedCallback) _this.initializedCallback(_this.pluginName, _this.fieldName);
   };
--- a/plugins/sync.js
+++ b/plugins/sync.js
@@ -1,7 +1,7 @@
 // @author         xelio
 // @name           Sync
 // @category       Misc
-// @version        0.4.1
+// @version        0.4.2
 // @description    Sync data between clients via Google Drive API. Only syncs data from specific plugins (currently: Keys, Bookmarks, Uniques). Sign in via the 'Sync' link. Data is synchronized every 3 minutes.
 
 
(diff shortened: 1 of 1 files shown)


McBen:
sry to say but the "comments" stuff is annoying.
I don't say they are bad at all. But please reduce them to a minimum. Only write a comment that is worth to read or when the code need explanation. In most cases a varialbe rename or an extra variable or moving stuff to an extra function can do a far better job.

code is read by programmers 

grafik
stuff like this doesn't help anybody. But makes the code messy and outdated comments can lead to wrong assumpations.

MysticJay:
@McBen It took (me) hours to understand SYNC as it was hardly commented at all. Yes code is read by programmers, but some are more skilled than others and some "see" the symantics from long constructs like x= a ? b : c; esp if a and b are implemented as functions.

when I read code I endorse evrey single bit of comment the programmer added as this will explain his intention and makes debugging way more easy (for me).

You may call me a bad programmer, but if we want new DEVs the current uncommented code is tough stuff.

plugins/sync.js
```diff
@@ -35,7 +35,7 @@ window.plugin.sync.authorizer = null;
 window.plugin.sync.registeredPluginsFields = null;
```
McBen, line 35: Why not comment the real important stuff like this?
What is it for? What's the structure of that variable?

plugins/sync.js
```diff
@@ -58,21 +58,22 @@ window.plugin.sync.updateMap = function(pluginName, fieldName, keyArray) {
...
                'fieldName': fieldName,
-               'callback': callback,
+               'callback': updateCallback,
                'initializedCallback': initializedCallback,
                'authorizer': plugin.sync.authorizer,
                'uuid': plugin.sync.uuid};
-  registeredMap = new plugin.sync.RegisteredMap(options);
-  plugin.sync.registeredPluginsFields.add(registeredMap);
+  registeredMap = new plugin.sync.RegisteredMap(options);   // create a new datasructure 
+                                                            // for the calling plugin and
```
McBen, line 70: double line comments

some one might (accidently) add a code line between them and split the comment

plugins/sync.js
```diff
@@ -58,21 +58,22 @@ window.plugin.sync.updateMap = function(pluginName, fieldName, keyArray) {
...
+  plugin.sync.registeredPluginsFields.add(registeredMap);   // add it to the main structure
```
McBen, line 71: as said above. would a explanlation of "sync.registeredPluginFields" wouldn't tell much more ?

plugins/sync.js
```diff
@@ -58,21 +58,22 @@ window.plugin.sync.updateMap = function(pluginName, fieldName, keyArray) {
...
 };
 
 
 
-//// RegisteredMap
+//// RegisteredMap *************************************************************************
```
McBen, line 76: note: comments like "*************"  usually mean: split this here in 2 files

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
   this.initializing = false;
   this.initialized = false;
   this.failed = false;
-
+  this.doSync = 'remote';                                   // normal 'sync'
+                                                            // 'no' sync needed
+                                                            // force 'remote' (1st sync)
+                                                            // force 'local'
+                                                            // sync is on 'stop'
+
```
McBen, line 106: I don't understand the comment. I assume "sync","no","remote","local","stop"  are valid states of `doSync` ?

maybe better something like - just my guess what they could mean:
```suggestion
 /** 
 *  doSync : state of current sync
 *   = "sync" : normal sync (what ever this means)
 *   = "no" : sync needed (or "no sync running, start delay" ?)
 *   = "stop" : sync stopped by user
 *   = "remote" : get values from remote
 *   = "local : use local values instead of remote
 */
  this.doSync = 'remote';
```

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
   this.callback = options['callback'];
   this.initializedCallback = options['initializedCallback'];
   this.authorizer = options['authorizer'];
-  this.uuid = options['uuid'];
+
+  this.uuid = options['uuid'];                              // local UUID
+  this.lastUpdateUUID = null;                               // the UUID that we think
+                                                            // did last Update (can be us)
```
McBen, line 90:
```suggestion
  this.lastUpdateUUID = null; // the UUID we assume did the last update
```

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
-        _this.map[key] = value;
-      } else {
-        delete _this.map[key];
+    this.lastUpdateUUID = this.uuid;                        // use this clients uuid
+    $.each(keyArray, function(ind, key) {                   // find all changed elements from the array
+      var value = window.plugin[_this.pluginName][_this.fieldName][key]; 
+      if(typeof(value) !== 'undefined') {                  
+        _this.map[key] = value;                             // replace element in the array
+      } else {                                              // or
+        delete _this.map[key];                              // remove it from the array
```
McBen, line 124:
```suggestion
        _this.map[key] = value; 
      } else {
        delete _this.map[key];
```

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
       }
     });
   } finally {
-    _this.dataStorage.saveFile(_this.prepareFileData());
+    _this.dataStorage.saveFile(_this.prepareFileData());    // save data and UUID to drive
```
McBen, line 128:
```suggestion
    _this.dataStorage.saveFile(_this.prepareFileData());
```

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
 window.plugin.sync.RegisteredMap.prototype.getFileName = function() {
   return this.pluginName + '[' + this.fieldName + ']'
 };
 
+// initialize access to the sync-file on drive
+//   callback (optional) - to be performed on success
 window.plugin.sync.RegisteredMap.prototype.initFile = function(callback) {
   var assignIdCallback, failedCallback, _this;
-  _this = this;
+  _this = this;                                             // have _this point to this Object
```
McBen, line 141:
```suggestion
  _this = this;
```
or replace "_this" with "self" ? ...this might help people to see the difference more clearly.

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
 
-  assignIdCallback = function(id) {
+  // file access granted, perform callback-function
+  assignIdCallback = function(id) {                         // where does this id come from?
     _this.forceFileSearch = false;
-    if(callback) callback();
+    if(callback) callback();                                // do the callback passed to .initfile
```
McBen, line 146:
```suggestion
    if (callback) callback();
```
maybe rename "callback" to "onSuccessCallback" ?

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
   failedCallback = function(resp) {
     _this.initializing = false;
     _this.failed = true;
     plugin.sync.logger.log(_this.getFileName(), 'Could not create file. If this problem persist, delete this file in IITC-SYNC-DATA-V3 in your Google Drive and try again.');
   };
 
-  this.dataStorage = new plugin.sync.DataManager({'fileName': this.getFileName(),
-                                                    'description': 'IITC plugin data for ' + this.getFileName()});
-  this.dataStorage.initialize(this.forceFileSearch, assignIdCallback, failedCallback);
+  this.dataStorage = new plugin.sync.DataManager(           // create a new dataStorage with
```
McBen, line 156:
```suggestion
  this.dataStorage = new plugin.sync.DataManager(
```
The interesting part is "what is a DataManager" - and this will hopefully answered later.

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
+    { 
+      'fileName': this.getFileName(),                       
+      'description': 'IITC plugin data for ' + this.getFileName()
+    }
+  );
+  
+  this.dataStorage.initialize(                              // initalize the dataStorage to gain 
+    this.forceFileSearch,                                   // find the related file on drive
+    assignIdCallback,                                       // onSuccess:
```
McBen, line 165: why is it named "assignIdCallback" ? it only triggers the callback? isn't it?

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
+    assignIdCallback,                                       // onSuccess: 
+    failedCallback                                          // onFailed: 
+  );
 };
 
 window.plugin.sync.RegisteredMap.prototype.initialize = function(callback) {
   this.initFile(this.loadDocument);
 };
-
+// prepend UUID to data
```
McBen, line 173:
```suggestion

// prepend UUID to data
```
space

plugins/sync.js
```diff
@@ -83,73 +84,93 @@ window.plugin.sync.RegisteredMap = function(options) {
...
-  this.dataStorage.initialize(this.forceFileSearch, assignIdCallback, failedCallback);
+  this.dataStorage = new plugin.sync.DataManager(           // create a new dataStorage with
+    { 
+      'fileName': this.getFileName(),                       
+      'description': 'IITC plugin data for ' + this.getFileName()
+    }
+  );
+  
+  this.dataStorage.initialize(                              // initalize the dataStorage to gain 
+    this.forceFileSearch,                                   // find the related file on drive
```
McBen, line 164:
```suggestion
    this.forceFileSearch,                                   // if true: find the related file on drive
```

But the "initialze" function should explain what the parameters are for. Not the function call.
If using a javadoc style comments most IDE's will display this informations.

plugins/sync.js
```diff
@@ -162,46 +183,109 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
   // this function called when the document is created first time
   // and the JSON file is populated with data in plugin field
   initializeFile = function() {
-    _this.map = {};
+    _this.map = {};                                         // create an empty map
```
McBen, line 186:
```suggestion
    _this.map = {};
```

plugins/sync.js
```diff
@@ -162,46 +183,109 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
...
 
-    // Init the map values if this map is first created
+    // copy all elements from the plugin's data to the new map
     $.each(window.plugin[_this.pluginName][_this.fieldName], function(key, val) {
       _this.map[key] = val;
     });
 
-    _this.dataStorage.saveFile(_this.prepareFileData());
+    _this.dataStorage.saveFile(_this.prepareFileData());    // save the map to drive
```
McBen, line 193:
```suggestion
    _this.dataStorage.saveFile(_this.prepareFileData());
```

McBen (commented):
just started to add some comments to the comments :)
Didn't had time for all..that was simply too much.

johnd0e:
In general agree with @McBen.
Also: initial message is not enough: "Clean PR fro SYNC50".
What is it all about? Please add some context.

See also https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Contributing-Code about commit messages style.

MysticJay:
@johnd0e @McBen  
> In general agree with @McBen.

Commenting SYNC is "work in progress" and some comments might even need to be revised.
I prefer comments that start at the same column (if possible) and do not exceed certain boundary. 
IMHO this helps to "understand" the code without the need to constantly jump between code and comments.
The above remarks you made will be taken forward the one or the other way.

> Also: initial message is not enough: "Clean PR fro SYNC50".
> What is it all about? Please add some context.
> See also https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Contributing-Code about commit messages style.

fixed

McBen:
> The proper use of comments is to compensate for our failure to express ourself in code. Note that I used the word failure. I meant it. Comments are always failures. We must have them because we cannot always figure out how to express ourselft without them, but their use is not a cause of celebration. 
_(Robert C. Martin / Clean Code)_

Some comments are required. Some comments might by helpful. But they always cost time to write, read and maintain. They can be boring and in worst case be missleading or wrong.

**Style**
> [..] without the need to constantly jump between code and comments.

right now you're jumping left & right. 
Even worse: when reviewing code on a smaller screen (or as side-by-side) you've messed up code or have to scroll left & right

This style of comments are fine when teaching someone to code. You've the pure code on the left and the documention on the right. This way you can focus and explain the code without the distraction of comments. 

While developing this is terrible wrong. While developing code & comments should be a unit. I expect a comment is important, has a valid reason to be there and I've to read it. (like `function feedGremlin() {  // never call it after midnight`)

MysticJay:
> This style of comments are fine when teaching someone to code. You've the pure code on the left and the documention on the right. This way you can focus and explain the code without the distraction of comments.

Still I prefer that way of explaining the code and if Xelio had done so in 2013 it wouldn't have cost us/me weeks to understand the code and someone else could have taken the job to refurbish it way earlier. Up to now all we received from other DEVs was "SYNC is a mess, I do not understand it, I will not touch it, I use my own code."

SYNC is very "difficult" code and digging out the ideas @xelio had in 2013 is time consuming, but unavoidable if we want to keep SYNC alive and compatible. 

For the time being the comment style will be "Code-Teaching". When the process is done, we can think about dropping, merging or rewriting comments.

johnd0e:
> SYNC is very "difficult" code

So may be it's worth some refactoring to make it easier?

Better spent time simplifying things than explaining wh things are so complicated.

MysticJay:
@johnd0e I explained my intention, the plan, the future development and that the remarks are valuable and will be taken forward into the next steps.

If you do not like my changed comments but still feel the issue is valid you are free to cherry-pick the dev-part only or ignore the PR.

plugins/sync.js
```diff
@@ -195,41 +195,97 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
     setTimeout(function() {_this.loadDocument();}, window.plugin.sync.checkInterval);
   };
 
-  // this function called when the document is loaded
-  // update local data if the document is updated by other (overwrite!)
-  // and adding a timer to further check for updates
+  // this function gets called when the document is loaded
```
McBen, line 198: 'document' usually reference to `window.document` . Here we talk about a g-drive file

plugins/sync.js
```diff
@@ -195,41 +195,97 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
...
-
-    if(_this.callback) _this.callback();
-    if(_this.initializedCallback) _this.initializedCallback(_this.pluginName, _this.fieldName);
+      // execute the callback functions             
+      if(_this.callback) _this.callback(            // execute the callback functions with
+        _this.pluginName,                           //    pluginName
+        _this.fieldName,                            //    fieldName
+        null,                                       //    null (for backwards compatibility)
+        true                                        //    true = full file load from Drive
+      );
```
McBen, line 265:
```suggestion
      // execute the callback functions             
      if(_this.callback) _this.callback(
        _this.pluginName,
        _this.fieldName,
        null, // obsolete, for backwards compatibility
        true // load complete file
      );
```

plugins/sync.js
```diff
@@ -216,30 +237,31 @@ window.plugin.sync.RegisteredMap.prototype.loadDocument = function(callback) {
...
 
 
 
 
-//// RegisteredPluginsFields
+//// RegisteredPluginsFields ***************************************************************
 // Store RegisteredMap and handle initialization of RegisteredMap
 window.plugin.sync.RegisteredPluginsFields = function(options) {
-  this.authorizer = options['authorizer'];
+  this.authorizer = options['authorizer'];                  //
```
McBen, line 367: empty comment
```suggestion
  this.authorizer = options['authorizer']; 
```

plugins/sync.js
```diff
@@ -259,62 +281,71 @@ window.plugin.sync.RegisteredPluginsFields.prototype.add = function(registeredMa
...
-  if(Object.keys(this.waitingInitialize).length > 0) {
-    this.timer = setTimeout(function() {_this.initializeWorker()}, 10000);
+  this.cleanWaitingInitialize();                            // clean up the initialization queue
+  plugin.sync.toggleDialogLink();                           // set the color of the menu entry
+  this.initializeRegistered();                              // try initialize remaining entries 
+                                                            //   in the queue
+  clearTimeout(this.timer);                                 // reset the timer
+  if(Object.keys(this.waitingInitialize).length > 0) {      // if queue still not empty
+    this.timer = setTimeout(
+      function() {_this.initializeWorker()}, 10000          // retry in 10 sec
```
McBen, line 436: move time constant to extra variable
```suggestion
      function() {_this.initializeWorker()}, RETRY_TIME
```

NB: also comment in 318 references this

plugins/sync.js
```diff
@@ -259,62 +281,71 @@ window.plugin.sync.RegisteredPluginsFields.prototype.add = function(registeredMa
...
+    );
   }
 };
 //// end RegisteredPluginsFields
 
 
 
 
-//// DataManager
+//// DataManager ***************************************************************************
```
McBen, line 445: *****************************
is a sign for "split this file here"

plugins/sync.js
```diff
@@ -457,37 +488,63 @@ window.plugin.sync.DataManager.prototype.createFile = function(callback) {
...
-      params: { uploadType: 'media' },
-      body: JSON.stringify(data)
-    }).execute();
-  });
+  
+  gapi.client.load('drive', 'v3').then(
+    function () {        
+      gapi.client.request({
+        path: '/upload/drive/v3/files/'+_this.fileId,
+        method: 'PATCH',                                    // why patch?
```
McBen, line 648: don't ask me

plugins/sync.js
```diff
@@ -457,37 +488,63 @@ window.plugin.sync.DataManager.prototype.createFile = function(callback) {
...
+  gapi.client.load('drive', 'v3').then(
+    function () {        
+      gapi.client.request({
+        path: '/upload/drive/v3/files/'+_this.fileId,
+        method: 'PATCH',                                    // why patch? 
+        params: { uploadType: 'media' },
+        body: JSON.stringify(data)
+      }).execute();                                         // write (patch) the file on drive
+    }
+                                                            // no errorhandling here?!
```
McBen, line 653: again: don't ask me

if you think this needs rework/review add a "TODO" or "FIXME" in comment.
(and make sure to handle these TODOs really soon)

plugins/sync.js
```diff
@@ -533,22 +593,27 @@ window.plugin.sync.Authorizer = function(options) {
   this.isAuthorizing = this.isAuthorizing.bind(this);
   this.authorize = this.authorize.bind(this);
 };
-
+// values that are static at the moment
```
McBen, line 703: if constant move to file-top

plugins/sync.js
```diff
@@ -533,22 +593,27 @@ window.plugin.sync.Authorizer = function(options) {
...
 window.plugin.sync.Authorizer.prototype.CLIENT_ID = '1099227387115-osrmhfh1i6dto7v7npk4dcpog1cnljtb.apps.googleusercontent.com';
 window.plugin.sync.Authorizer.prototype.SCOPES = 'https://www.googleapis.com/auth/drive.file';
 
+// method to tell if SYNC is authorized
```
McBen, line 707: I would prefer javadoc style 
```suggestion
/**
 * aks if SYNC is authorized
 * @return true if authorized
 **/
```

plugins/sync.js
```diff
@@ -533,22 +593,27 @@ window.plugin.sync.Authorizer = function(options) {
...
 window.plugin.sync.Authorizer.prototype.isAuthed = function() {
   return this.authorized;
 };
 
+// method to tell if authorization is in progress
 window.plugin.sync.Authorizer.prototype.isAuthorizing = function() {
   return this.authorizing;
 };
+
+// (external) method to add an other authCallback function
```
McBen, line 717: what is "external" ?

plugins/sync.js
```diff
@@ -533,22 +593,27 @@ window.plugin.sync.Authorizer = function(options) {
...
   return this.authorizing;
 };
+
+// (external) method to add an other authCallback function 
 window.plugin.sync.Authorizer.prototype.addAuthCallback = function(callback) {
   if(typeof(this.authCallback) === 'function') this.authCallback = [this.authCallback];
   this.authCallback.push(callback);
 };
 
+// (internal) function that will execute all enregistered authCallback functions
```
McBen, line 723: what is "internal" ? private? ... then maybe mark it the good old way: "_authComplete"

plugins/sync.js
```diff
@@ -561,6 +626,9 @@ window.plugin.sync.Authorizer.prototype.authComplete = function() {
   }
 };
 
+// (external) function that will initiate the authorization
+// redirect: (type: boolean)
```
McBen, line 737: redirect ? where is it mentioned?

plugins/sync.js
```diff
@@ -702,10 +773,12 @@ window.plugin.sync.loadUUID = function() {
   }
 };
 
+
 window.plugin.sync.updateLog = function(messages) {
   $('#sync-log').html(messages);
 };
 
+// AuthButon that shows IN the dialog
```
McBen, line 888:
```suggestion
// Toggle the state of the AuthButton in sync dialog
```

plugins/sync.js
```diff
@@ -717,6 +790,7 @@ window.plugin.sync.toggleAuthButton = function() {
   $('#sync-authButton').toggleClass('sync-authButton-dimmed', authed || authorizing);
 };
 
+// Sync menuitem in the toolbox
```
McBen, line 900: "toggle" and "sync" sounds really differnt..one of it might be wrong

plugins/sync.js
```diff
@@ -774,23 +851,32 @@ window.plugin.sync.setupCSS = function() {
   .appendTo("head");
 };
 
+////////////////////////////////////////////////////
+//// Main **********************************************************************************
 var setup = function() {
-  window.plugin.sync.logger = new plugin.sync.Logger({'logLimit':10, 'logUpdateCallback': plugin.sync.updateLog});
+  window.plugin.sync.logger = new plugin.sync.Logger(
+    {'logLimit':10, 'logUpdateCallback': plugin.sync.updateLog}
+  );
```
McBen, line 966: go the full way
```suggestion
  window.plugin.sync.logger = new plugin.sync.Logger({
      logLimit: 10, 
      logUpdateCallback: plugin.sync.updateLog
  });
```

plugins/sync.js
```diff
@@ -774,23 +851,32 @@ window.plugin.sync.setupCSS = function() {
...
   window.plugin.sync.loadUUID();
   window.plugin.sync.setupCSS();
   window.plugin.sync.setupDialog();
-
-  window.plugin.sync.authorizer = new window.plugin.sync.Authorizer({
-    'authCallback': [plugin.sync.toggleAuthButton, plugin.sync.toggleDialogLink]
-  });
+// create an authorizer structure
```
McBen, line 970: keep the empty line as separator

plugins/sync.js
```diff
@@ -774,23 +851,32 @@ window.plugin.sync.setupCSS = function() {
...
+  window.plugin.sync.authorizer = new window.plugin.sync.Authorizer(
+    {
+    'authCallback': [plugin.sync.toggleAuthButton,          // functions to execute after 
+                     plugin.sync.toggleDialogLink           // successfull authorization
```
McBen, line 974: another good example why this comment format is bad,
if you add another callback this comment get's messed up ... and it's currently hard to read if you're just looking for "toggleDialogLink"

McBen (commented):
some more comments

plugins/sync.js
le-jeu: try with `PUT` https://developers.google.com/drive/api/v2/reference/files/update
Google is likely to translate `PATCH` by `PUT` on the route `https://www.googleapis.com/upload/drive/v2/files/{fileId}`

Suburbanno:
@MysticJay Do you intend to continue this contribution? If not, could you close it?

PR #409, opened 2020-12-07
Fix #364

### LINKED ISSUE #364: genPostData should use a different timestamp from storageHash.oldestTimestamp when getOlderMessages =  True.
Currently IITC uses the stored oldest timestamp of received messages as the `maxTimestampMs` in the `data` for retrieving messages when `getOlderMessages` is `true`.

```javascript
data = $.extend(data, {maxTimestampMs: storageHash.oldestTimestamp});
```
However, where a single request results in 50+ (or a large value, unknown specific) messages all at the same time, say when a heavily layered field set is destroyed, this means that the first request sets the oldestTimestamp to that time, but every subsequent request keeps requesting the same time period.

Since the chat never receives older messages, it logjams at this timestamp and won't display any earlier entries.

**Suggested Solution:**

Use the timestamp of the oldestTimestamp minus one millisecond.  Since Niantic uses end inclusive timestamps in their request, you need to shift the end timestamp back by one millisecond or you will continue to receive the same data over and over.

```javascript
data = $.extend(data, {maxTimestampMs: storageHash.oldestTimestamp - 1});
```

le-jeu:
_Originally posted by @Looka13 in https://github.com/IITC-CE/ingress-intel-total-conversion/issues/364#issuecomment-740100570_

Use `plextContinuationGuid` to fix #364 like stock intel

I didn't test against a virus yet


core/code/chat.js
```diff
@@ -106,7 +106,10 @@ window.chat.genPostData = function(channel, storageHash, getOlderMsgs) {
   if(getOlderMsgs) {
     // ask for older chat when scrolling up
     // "-1" to avoid looping at the same timestamp if many msgs are returned
```
McBen: this comment is now obsolete and missleading

core/code/chat.js
```diff
@@ -302,6 +308,30 @@ window.chat.nicknameClicked = function(event, nickname) {
 }
 
 window.chat.writeDataToHash = function(newData, storageHash, isPublicChannel, isOlderMsgs) {
+
+  if (newData.result.length > 0) {
+    //track oldest + newest timestamps/GUID
+    let first = {
```
McBen:
```suggestion
    var first = {
```

iitc uses ES5 - no `let`

core/code/chat.js
```diff
@@ -302,6 +308,30 @@ window.chat.nicknameClicked = function(event, nickname) {
...
+      guid: newData.result[0][0],
+      time: newData.result[0][1]
+    };
+    let last = {
```
McBen:
```suggestion
    var last = {
```
ES5

core/code/chat.js
```diff
@@ -302,6 +308,30 @@ window.chat.nicknameClicked = function(event, nickname) {
...
+      guid: newData.result[newData.result.length-1][0],
+      time: newData.result[newData.result.length-1][1]
+    };
+    if (storageHash.oldestTimestamp === -1 || storageHash.oldestTimestamp >= last.time) {
+      if (isOlderMsgs || storageHash.oldestTimestamp != last.time) {
+        storageHash.oldestTimestamp = last.time;
+        storageHash.oldestGUID = last.guid;
+      }
+    }
+    if (storageHash.newestTimestamp === -1 || storageHash.newestTimestamp <= first.time) {
```
McBen, line 334: shouldn't this be `last` if `ascendingTimestampOrder: true` ? 

but I see this information isn't available here.

core/code/chat.js
```diff
@@ -302,6 +308,30 @@ window.chat.nicknameClicked = function(event, nickname) {
...
+      if (!isOlderMsgs || storageHash.newestTimestamp != first.time) {
+        storageHash.newestTimestamp = first.time;
+        storageHash.newestGUID = first.guid;
+      }
+    }
+  }
```
McBen: I would recommend moving this to a new function for better code style.
`writeDataToHash` is already a bloated function.

le-jeu: what about moving markup rendering to `chat.renderMsg` ?

PR #414, opened 2020-12-16
Multi project extension: fixes

MysticJay:
1) As font awesome plugin was removed from the merge, some icons need an alternative string to display (using first three 
    letters of the pluginname now)
2) redrawing markers when switching between different Bookmark Projects with MPE fixed
3) bookmarks: moved `setupCSS` to the end to improve readability of built script
4) minor refactoring

### DIFF
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -993,6 +993,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
         window.plugin.bookmarks.refreshBkmrks();
         window.plugin.bookmarks.resetAllStars();
         window.runHooks('pluginBkmrksSyncEnd', {"target": "all", "action": "sync"});
+        console.log('BOOKMARKS: synchronized all from drive after offline');
         return;
       }
 
@@ -1000,6 +1001,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
       if(e.isLocal) {
         // Update pushed successfully, remove it from updatingQueue
         delete window.plugin.bookmarks.updatingQueue[e.property];
+        console.log('BOOKMARKS: synchronized to drive');
       } else {
         // Remote update
         delete window.plugin.bookmarks.updateQueue[e.property];
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -992,7 +992,8 @@ window.plugin.bookmarks.loadStorageBox = function() {
       if(fullUpdated) {
         window.plugin.bookmarks.refreshBkmrks();
         window.plugin.bookmarks.resetAllStars();
-        console.log('BOOKMARKS: synchronized all after offline');
+        window.runHooks('pluginBkmrksSyncEnd', {"target": "all", "action": "sync"});
+        console.log('BOOKMARKS: synchronized all from drive after offline');
         return;
       }
 
@@ -1000,6 +1001,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
       if(e.isLocal) {
         // Update pushed successfully, remove it from updatingQueue
         delete window.plugin.bookmarks.updatingQueue[e.property];
+        console.log('BOOKMARKS: synchronized to drive');
       } else {
         // Remote update
         delete window.plugin.bookmarks.updateQueue[e.property];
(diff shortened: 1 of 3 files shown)


johnd0e:
> 1. As font awesome plugin was removed from the merge, some icons need an alternative string to display (using first three letters of the pluginname now)

The font loading was removed, but I see the rest of plugin left opinionated: plugin still tries to use fa-icons itself, fontawesome just need to be loaded by some other plugin.

1) That is not bad itself (especially if sample code for such plugin would be provided).
But look: current implementation relies on outdated fa version! That is NOT GOOD. Ref: https://github.com/IITC-CE/ingress-intel-total-conversion/pull/221#discussion_r311423667
2) May be we should consider some alternative icons implementation instead of fontawesome 

Related: #236

MysticJay:
@johnd0e fa plugin will be refurbished with an svg based alternative.
that is on my roadmap, including using a younger version.

Removing the use of FA in this moment would disturb the existing UX for existing MPE users.

I think the current implementation of MPE is an acceptable compromise.

ATM most of ZasoItems use FA
So FA based icons are my first choice.

plugins/bookmarks.js
```diff
@@ -992,14 +992,16 @@ window.plugin.bookmarks.loadStorageBox = function() {
       if(fullUpdated) {
         window.plugin.bookmarks.refreshBkmrks();
         window.plugin.bookmarks.resetAllStars();
-        console.log('BOOKMARKS: synchronized all after offline');
+        window.runHooks('pluginBkmrksSyncEnd', {"target": "all", "action": "sync"});
+        console.log('BOOKMARKS: synchronized all from drive after offline');
```
johnd0e, line 996: Do we need this?

plugins/bookmarks.js
```diff
@@ -992,14 +992,16 @@ window.plugin.bookmarks.loadStorageBox = function() {
...
         return;
       }
 
       if(!e) return;
       if(e.isLocal) {
         // Update pushed successfully, remove it from updatingQueue
         delete window.plugin.bookmarks.updatingQueue[e.property];
+        console.log('BOOKMARKS: synchronized to drive');
```
johnd0e, line 1004: And this

johnd0e (changes requested):
Not sure that these should be exposed:

```js
window.plugin.drawTools.mpe = {};
window.plugin.drawTools.mpe.ui = {};
window.plugin.drawTools.mpe.boot //...
window.plugin.drawTools.mpe.initMPE //...
```
(I've checked only draw-tools atm)

plugins/bookmarks.js
MysticJay: yes, otherwise Book-Markers would not update on map
console.log: me I like it , you do not I guess.

plugins/bookmarks.js
MysticJay: console.log: me I like it , you do not I guess.

johnd0e: I like it, when it needed.
But I do not like when I see a mess of console.log (from dozens sources) and unable to find that one I need.
Just install old iitc-me and see what I mean.
And I've put much effort to make log useful again in IITC-CE.

johnd0e:
> 1. some icons need an alternative string to display (using first three
>    letters of the pluginname now)

@MysticJay 
May be introduce something like `window.plugin.mpe.data.getTitleShort(PJ);` (or `getSymbol` etc)

MysticJay:
@johnd0e This is only a workaround to fix missing font-awsome. I do not want to manifest that in a function. An icon is the goal.

plugins/multi-projects-extension.js
```diff
@@ -506,12 +510,10 @@ window.plugin.mpe.setupCSS = function(){
 
 
 var setup = function(){
-  window.pluginCreateHook('mpe');
```
modos189, line 509: Is the `mpe` hook no longer needed?

modos189 (approved):

plugins/multi-projects-extension.js
johnd0e: Hook is needed, but there is no need to pre-register it.

PR #420, opened 2020-12-26
[IITCm] allow popups to fix some Google login cases from plugins

### LINKED ISSUE #362: IITCm for Android: gapi.auth2.authorize limitation
`gapi.auth2.authorize` looks ok to be used in IITC, using either the google ingress account or another google account.
But, on android device, the webview seems to break the gapi auth process as long as we ask the user for a prompt (`consent` or `select_account`).
Apparently, the google popup is forbidden to close itself nor to send data back to the original window.

image

I don't know if this is some absolute limitation of android + gapi or if a solution exists. (android code
 or javascript). I explored javascript side without finding answer.

Example of authentication script (ask for a token)
```js
// @author         jaiperdu
// @name           Authentication test
// @category       Misc
// @version        0.1.0
// @description    Simple authentication test


window.plugin.authTest = function() {};

window.plugin.authTest.dialogHTML = null;

// web app client id with auth screen allowing https://intel.ingress.com as origin
window.plugin.authTest.CLIENT_ID = 'CLIENTID.apps.googleusercontent.com';

window.plugin.authTest.SCOPES = 'openid profile email';
window.plugin.authTest.TYPE = 'id_token permission';

window.plugin.authTest.authorize = function() {
  const options = {
    prompt: 'select_account', // or 'consent' (use 'none' for implicit auth if possible)
    client_id: this.CLIENT_ID,
    scope: this.SCOPES,
    response_type: this.TYPE
  };

  window.gapi.auth2.authorize(options, response => {
    if (response.error) {
      const err = `error: ${response.error}: ${response.error_subtype}`;
      alert(err);
      console.log(err);
      return;
    }

    alert(response.access_token);
    console.log(response.id_token, response.access_token);
  });
};

window.plugin.authTest.setupDialog = function() {
  $('#toolbox').append('<a id="authTest-authorize" onclick="window.plugin.authTest.authorize();">AuthTest</a> ');
};

var setup = function() {
  if (typeof window.gapi !== "undefined") {
    window.gapi.load("auth2", () => {
      window.gapi.auth2.enableDebugLogs(true);
    });
  }
  else {
    const script = document.createElement("script");
    script.type = "text/javascript";
    script.async = true;
    script.defer = true;
    script.src = "https://apis.google.com/js/platform.js";
    script.onload = () => {
      window.gapi.load("auth2", () => {
        window.gapi.auth2.enableDebugLogs(true);
      });
    };
    (document.body || document.head || document.documentElement).appendChild(
      script
    );
  }
  window.plugin.authTest.setupDialog();
};
```

related #277 #334

le-jeu:
allow popup inside iitcm as long as it is in the whitelist  by

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/4bdfd4cb767ff8b9aa94c3baac66e2fcd2c35d3a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewClient.java#L268-L270

fix #362 

> I found a workaround for android source.
> From https://stackoverflow.com/questions/23308601/android-open-pop-up-window-in-my-webview and https://stackoverflow.com/questions/23211722/window-close-not-working-in-android-webview (and some other I lost the track)
> 
> This is a PoC and aims simplicity (with the cost of safety). I almost have no idea what I am doing, but it works so far.

https://github.com/IITC-CE/ingress-intel-total-conversion/issues/362#issuecomment-636477585

here is a plugin to play with google auth : https://le-jeu.github.io/dist/auth-test.user.js


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebChromeClient.java
```diff
@@ -5,6 +5,7 @@
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Message;
```
johnd0e, line 9: I do not see that be used anywhere

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebChromeClient.java
```diff
@@ -208,4 +209,23 @@ public boolean onJsPrompt(final WebView view, final String url, final String mes
...
+        transport.setWebView(newWebView);
+        resultMsg.sendToTarget();
+
+        return true;
+    }
+
+    public void onCloseWindow(WebView w){
+        if (mIitc.getWebView() != w)
+            mIitc.getWebView().removeView(w);
+        Log.d("onCloseWindow");
```
johnd0e: Is that really useful?

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebChromeClient.java
```diff
@@ -208,4 +209,23 @@ public boolean onJsPrompt(final WebView view, final String url, final String mes
...
+        view.addView(newWebView);
+        WebView.WebViewTransport transport = (WebView.WebViewTransport) resultMsg.obj;
+        transport.setWebView(newWebView);
+        resultMsg.sendToTarget();
+
+        return true;
+    }
+
+    public void onCloseWindow(WebView w){
+        if (mIitc.getWebView() != w)
```
johnd0e: Please use some meaningful name instead of `w`.

johnd0e (dismissed):

johnd0e:
> allow popup inside iitcm as long as it is in the whitelist

Please point me where is that whitelist in code

johnd0e:
Also I'd suggest you to choose PR titles more carefully, as we using them directly in changelog.

E.g.: [IITCm] allow popups to fix some Google login cases

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebChromeClient.java
le-jeu: because `onCreateWindow` last argument

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebChromeClient.java
le-jeu: afair, It was, but it looks ok without atm

le-jeu: nvm, It is useful, without, the popup can't close itself

johnd0e: Actually I am asking about logging only.

le-jeu:
Commit 4b0ff58a7b2e300277ea4d64598b4555470a3e3a removed the account management from the device leading the `IITC_DeviceAccountLogin` not beeing use if I understand the code correctly.

Using the device to select the account may make thing more Android-like but I have no idea how it works(ed)

johnd0e:
> Commit [4b0ff58](https://github.com/IITC-CE/ingress-intel-total-conversion/commit/4b0ff58a7b2e300277ea4d64598b4555470a3e3a) removed the account management from the device leading the `IITC_DeviceAccountLogin` not beeing use if I understand the code correctly.

I suppose it was removed because it (never?) was working.

> Using the device to select the account may make thing more Android-like but I have no idea how it works(ed)

Yes, that would be great. But I'm even not sure that it is ever possible to implement, because intel auth code is out of our control.

le-jeu:
This feature was supposed to use with the `x-auto-login` HTTP header from server, but I don't find any documentation about it in google auth docs.

I tried re-enabling it without seeing any changes and removing the class `IITC_DeviceAccountLogin` looks ok so far.

modos189:
Can you please provide examples of plugins where this PR helps?

le-jeu:
Sync, Wasabee, or any other plugins that uses google auth and allow the user to select their login account.

McBen:
Looks like the last commit 205635b won't work for me.  I'm still investigating. But rest look good.

McBen:
what I see sofar:
`window.open("https://www.google.com");` 
will no longer work on IITCm.  ( because of 205635b )

le-jeu:
it should be working now

McBen:
runs partially
grafik

And please don't commented out code.

McBen:
On a first glace it looks good to me. Will do some more testing soon.

I still get an exception when closing the popup. But nothing critical and maybe related to my clicks.
```
D/iitcm: close popup window
W/cr_AwAutofillManager: Application attempted to call on a destroyed AutofillManagerWrapper
    java.lang.Throwable
        at gV.a(chromium-TrichromeWebViewGoogle.aab-stable-428010133:2)
        at org.chromium.components.autofill.AutofillProvider.setNativeAutofillProvider(chromium-TrichromeWebViewGoogle.aab-stable-428010133:5)
        at J.N.MQGusfGb(Native Method)
        at xK.run(chromium-TrichromeWebViewGoogle.aab-stable-428010133:2)
        at DM.handleMessage(chromium-TrichromeWebViewGoogle.aab-stable-428010133:10)
        at FM.a(chromium-TrichromeWebViewGoogle.aab-stable-428010133:3)
        at org.chromium.android_webview.AwContents.g(chromium-TrichromeWebViewGoogle.aab-stable-428010133:9)
        at oK.run(chromium-TrichromeWebViewGoogle.aab-stable-428010133:1)
        at android.os.MessageQueue.nativePollOnce(Native Method)
        at android.os.MessageQueue.next(MessageQueue.java:335)
        at android.os.Looper.loop(Looper.java:183)
        at android.app.ActivityThread.main(ActivityThread.java:7660)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

le-jeu:
Scenario I was not able to solve is about the destruction of the webview object. I use the chrome://inspect as tool to know the living webviews.

I wanted to use a dialog since it looks good as UI for the login purpose.
When I dismiss the dialog, I expect the attached webview to be automatically destroyed, but it is not automatic.
It does on some occasion but wont on other. Basically, when I open a external link, the dialog should be dismiss, but it looks the webview is not attached yet, so it's not destroyed. (that's a guess)
This also happens if the user closes the dialog (using the back button).

The current behavior is dirty (either leaving webviews, or raising exception) and I don't know how to fix it.

The choice of a dialog looks bad so far, using an other activity (with a fixed webview from resources for instance) might be cleaner.

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
```diff
@@ -0,0 +1,143 @@
...
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT)
+            setWebContentsDebuggingEnabled(true);
+
+        setWebChromeClient(new WebChromeClient() {
+            @Override
+            public void onCloseWindow(WebView view) {
+                Log.d("close popup window");
+                mDialog.dismiss();
+                view.destroy();
```
McBen: is "destroy" here required?
and why is the superclass function skipped?

...BTW I'm a totally noob with android+webviews

le-jeu: Not adding it leads to some webviews not beeing destroyed on my phone

I'm also a noob with those two...

modos189 (approved):

PR #423, opened 2020-12-28
Script controlled login pages

McBen:
Manage the list of hardcoded pathes by the main-script (or plugin) instead of the mobile code.

PR #420 is required to allow interactions with these pages.
(Based on PR #420 - Only the last commit 5f679db (or followers) is subject of this PR.)

in Detail:
It moves the 
```
   if (uriHost.endsWith("facebook.com")  && (uriPath.contains("oauth") || .....
```
from the mobile-code into the control of core/total-conversion-build.js
```
  android.registerLoginPage("^facebook\\.com\\/(.*oauth|login.php$|checkpoint\/$)");
```

Gives more control about what pages are open in the app itself or on a external view. For custom 3rd-party login stuff. 
This way you create various login plugins (like "sign in with Apple", "sign in with Twitter", ... ) without even touching IITC-Core code.


McBen:
ATM marked as draft 'cause I didn't tested latest build.

johnd0e:
@McBen
Please explain in details what exact issue you are trying to solve here.

FesterCluck:
This would be very useful. One must remember IITC started as a server-side implementation which one often had to log in to, in addition to Ingress. Many of those still exist to this day.

PR #427, opened 2020-12-31
privacy-plugin: fix privacy control in expanded chat

### LINKED ISSUE #426: Privacy view on Intel: Controls do not behave as expected
The behavior of the controls for the COMM-pane if "Privacy view on Intel" is broken.

image

The left field should either show "Privacy active" or "Privacy inactive". Also the COMM can be minimized to the top left, which might be intended, but then two arrow down show up.

Tested with 
IITC Beta (0.31.1.20201230.231144)
IITC plugin: Privacy view on Intel (1.1.0.20201230.231144)

le-jeu:
fix #426 

also, use #chatcontrol class and css to control the chat toggle shape without changing the toggle class.

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -521,19 +521,17 @@ window.chat.tabToChannel = function(tab) {
 window.chat.toggle = function() {
   var c = $('#chat, #chatcontrols');
   if(c.hasClass('expand')) {
-    $('#chatcontrols a:first').html('<span class="toggle expand"></span>');
     c.removeClass('expand');
     var div = $('#chat > div:visible');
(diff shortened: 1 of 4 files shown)


johnd0e (approved):

MysticJay (changes requested):
The changes return old functionality.
But shouldn't the arrow point down after maximizing?

le-jeu:
> The changes return old functionality.
> But shouldn't the arrow point down after maximizing?

image

MysticJay (approved):

modos189 (approved):

PR #428, opened 2021-01-03
Draw-Functions

MysticJay:
Draw-Tools:  

(Migrating additional functions from DrawTools plus.)
These functions add a convenient and easy to use way to directly create a draw, using IITCs standard parameter and settings from a dialog or script. An optional color-parameter can be used, otherwise the currently active color will be used. Passing a color will not change the selected color.

  window.plugin.drawTools.drawPolyline(arrCoordArr[, color]);
  window.plugin.drawTools.drawPolygon(arrCoordArr[, color]);
  window.plugin.drawTools.drawCircle(coord, radius[, color]);
  window.plugin.drawTools.drawMarker(coord[, color]);
  
  coord := [LatE6,LonE6]
  arrCoordArr : [coord, coord]  e.g.  `var arrCoordArr = [[42.447585,13.285419],[41.692902,13.418577]];`

Plugins using these functions: Destroyed Links Simulator, Link Prolongation

### DIFF
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -90,7 +90,7 @@ window.plugin.drawTools.fireDraw = function(layer, layerType) {
 /**
  * Create a polyline
  * @param arrCoordArr Array of coordinates
- * @param {string=] color (optional)
+ * @param {string=} color (optional)
  */
 window.plugin.drawTools.drawPolyline = function(arrCoordArr, color) {
   if (color !== undefined) {
@@ -113,7 +113,7 @@ window.plugin.drawTools.drawPolyline = function(arrCoordArr, color) {
 /**
  * Create a polygon
  * @param arrCoordArr Array of coordinates
- * @param {string=] color (optional)
+ * @param {string=} color (optional)
  */
 window.plugin.drawTools.drawPolygon = function(arrCoordArr, color) {
   if (color !== undefined) {
@@ -137,7 +137,7 @@ window.plugin.drawTools.drawPolygon = function(arrCoordArr, color) {
  * Create a marker
  * @param coord coordinates [LatE6, LonE6]
  * @param radius (in meters)
- * @param {string=] color (optional)
+ * @param {string=} color (optional)
  */
 window.plugin.drawTools.drawCircle = function(coord, radius, color) {
   if (color !== undefined) {
@@ -160,7 +160,7 @@ window.plugin.drawTools.drawCircle = function(coord, radius, color) {
 /**
  * Create a marker
  * @param coord coordinates (LatE6, LonE6)
- * @param {string=] color (optional)
+ * @param {string=} color (optional)
  */
 window.plugin.drawTools.drawMarker = function(coord, color) {
   if (color !== undefined) {
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -78,14 +78,8 @@ window.plugin.drawTools.setDrawColor = function(color) {
 }
 
 // Easy create a draw from an IITC-plugin, allow passing an optional color parameter
-  /*
-  // "color" is optional
-  window.plugin.drawTools.drawPolyline(arrCoordArr, color);
-  window.plugin.drawTools.drawPolygon(arrCoordArr, color);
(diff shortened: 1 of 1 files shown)


johnd0e:
Not sure that we should reinvent the wheel here.
Why we need any special function when we have all the [Leaflet](https://leafletjs.com/reference-1.7.1.html) and [Leaflet.draw](
https://leaflet.github.io/Leaflet.draw/docs/leaflet-draw-latest.html) directly?

plugins/draw-tools.js
```diff
@@ -77,6 +77,109 @@ window.plugin.drawTools.setDrawColor = function(color) {
   });
 }
 
+// Easy create a draw from an IITC-plugin, allow passing an optional color parameter
+
+// common function used by folloing functions to trigger display on new drawn item
+window.plugin.drawTools.fireDraw = function(layer, layerType) {
```
johnd0e, line 83: This wrapper is excessive.
I'd prefer to use original code (it is even more expressive).

plugins/draw-tools.js
```diff
@@ -77,6 +77,109 @@ window.plugin.drawTools.setDrawColor = function(color) {
...
+    layerType: layerType
+  });
+}
+
+/**
+ * Create a polyline
+ * @param arrCoordArr Array of coordinates
+ * @param {string=} color (optional)
+ */
+window.plugin.drawTools.drawPolyline = function(arrCoordArr, color) {
```
johnd0e, line 95: All following functions have the same structure, and perform the same action: temporary override currentcolor, and fire draw with params which are default for corresponding shape.

It may be better to refactor into single function.
But I tend to thing that something is wrong in the approach itself - this overriding is ugly.

So let's start from the very beginning: why we ever need to override current color? Why not draw with specified color directly?

MysticJay: e.g. if a script detects a Crosslink from a draw on a draw you might want to automatically highlight that by drawing the link in a different color, without changing the chosen drawcolor. 
_It's "put a red line here, whatever current color is."_ 
Link prolongation uses the functions too.

Putting that into a single function would not be nice either, as each has different parms. (Coords, arrCordsArr, Radius)

MysticJay:
> 
> 
> Not sure that we should reinvent the wheel here.
> Why we need any special function when we have all the [Leaflet](https://leafletjs.com/reference-1.7.1.html) and [Leaflet.draw](https://leaflet.github.io/Leaflet.draw/docs/leaflet-draw-latest.html) directly?

Initially I only wanted to migrate the remaining functions from DTplus to DrawTools only. But by now I found that it can be pretty handy to have those wrapper functions without the need to learn/understand leaflet first.

If many plugins use a leaflet function and leaflet changes, it would mean that all these plugins break until they have adopted to the new leaflet. inserting an intermediate layer helps to unify the interface to what WE need. This reduces the work to adopt to new leaflet to this one interface and all plugins can rely on it.

Reinventing a wheel would be to recreate leaflet. And to be honest, IITC is adding a lot of functions that could be done by their atomic code.

plugins/draw-tools.js
johnd0e: > e.g. if a script detects a Crosslink from a draw on a draw you might want to automatically highlight that by drawing the link in a different color, without changing the chosen drawcolor.

I understand why one may want to **use** default color, or current color, or any specific.
But I do not understand for what purpose all these functions **change** default color (temporally).

johnd0e:
> without the need to learn/understand leaflet first.

I'd strongly suggest you to learn Leaflet. This lets you to avoid reinventing many different wheels.

And again about these wrappers. Well, they could be useful, may be. Or may be not.
It's really hard to judge, because code of draw-tools is ugly and overcomplicated.

So I am hardly against adding anything to it in its current state.
I'd rather suggest you to refactor current code first, and then it will be more clear what api should we add.

Current iitc suffers from over-exposing all its internals. I believe that we should start opposite - remove all unrelated functions from api.

johnd0e:
> If many plugins use a leaflet function and leaflet changes, it would mean that all these plugins break until they have adopted to the new leaflet.

Nonsense.
Leaflet API is stable, by definition.
And even if some things will change in 2.0 - a couple of wrappers will not save you.

PR #432, opened 2021-01-27
fix gps orientation side effect

le-jeu:
when gps orientation is enabled in iitc android configuration but user location layer is disabled (from layer menu), iitc rotates all html elements with classname `container`.

### DIFF
--- a/mobile/plugins/user-location.js
+++ b/mobile/plugins/user-location.js
@@ -132,6 +132,8 @@ window.plugin.userLocation.onOrientationChange = function(direction) {
 
   window.plugin.userLocation.user.direction = direction;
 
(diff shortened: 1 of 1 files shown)


modos189:
In theory this is the right change, but do you have an example where some plugin uses the `container` classname so it causes a problem?

le-jeu:
the wasabee plugin and a recent plugin I made (not on purpose)

This issue could have been produced for more than a year. But nobody disables the userlocation from the layer chooser but uses the setting instead.

While it should be fixed, there is no hurry

image

modos189:
Wow looks like a fun redesign :)

Issue #325, opened 2020-02-12
Permanent portal entities


johnd0e:
Currently portal entity (represented by CircleMarker) is not permanent: every time we get some data with newer timestamp - portal marker is recreated:
<details><summary><code>createPortalEntity()</code></summary>

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/36a9b69e55c409a091523401422f532e628c3b1d/core/code/map_data_render.js#L292-L307
</details>

So all additional data (potentially attached by plugins) is **discarded**, including event listeners and other leaflet-binded things.
In such case two hooks are run consequentially:
<details>

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/36a9b69e55c409a091523401422f532e628c3b1d/core/code/map_data_render.js#L225
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/36a9b69e55c409a091523401422f532e628c3b1d/core/code/map_data_render.js#L351
</details>

So data is discarded, but not completely lost: plugin may restore some data from `previousData` property.
But as you see it's not so obvious and in many cases it may be tricky or cumbersome - to keep continuity between separated portal entities.

I propose to change this approach to some more simple and consistent: keep same portal marker, and only update it's data.
Examples of potential benefits:
- Use leaflet event system for portals.
  See also #217.
  In particular, this may help to implement **item 1** of https://github.com/IITC-CE/ingress-intel-total-conversion/issues/197#issue-443106932 (and all similar).
- Use other portal-binded objects, like tooltips (#195).
- Keeping permanent (but updating) entity data may help with issues like #324 
- With some extent, it is important step towards https://github.com/IITC-CE/ingress-intel-total-conversion/issues/121#issuecomment-585100660
  Another necessary change here - keep markers removed from map in some cache.
  Ether as leaflet objects, or just raw data (I'm not sure yet).

johnd0e:
> I propose to change this approach to some more simple and consistent: keep same portal marker, and only update it's data.

But we must consider compatibility, so all current hooks should stay.
1. Instead, we could extend hook data with additional property `update` (both hooks)
2. Alternatively, we may leave old hooks apart (keep as they are), and implement this as completely independent event system #217.

#107 is another important improvement (though not directly related to subj).

MysticJay:
Related problem?
Hook `portalSelected` gets executed two or three times when a portal is selected.
code to check:
```
window.addHook("portalSelected", someFunction);
[...]

Issue #142, opened 2019-03-27
Portals levels in layer chooser


johnd0e:
We rarely need to fine-select exact portals levels to show. And those portals checkboxes take too much space in layer chooser.

What we can do:
* use custom hierarchical layer chooser, like here:
  - https://github.com/McBen/ingress-intel-total-conversion
  - https://github.com/bambrikii/leaflet-layer-tree-plugin
  - https://jjimenezshaw.github.io/Leaflet.Control.Layers.Tree/examples/
* use completely different control, like ingress stock.
* use simple dialog (or even plain text inputbox)

PR #436, opened 2021-02-05
Support portal history

### LINKED ISSUE #434: Portal history: Ingress's newly-released layers for explorer and pioneer activity
The generic Ingress Intel map now shows a player's visited (explorer) and captured (pioneer) portals using two filters: https://community.ingress.com/en/discussion/13451/portal-history-now-available-on-intel-map

Currently, this feature is not working in IITC. Could you add a feature for these layers? Thanks!

MysticJay:
[Portal history](https://community.ingress.com/en/discussion/13451/portal-history-now-available-on-intel-map) provided by intel is now supported by IITC-CE!

- History values are now displayed in portal details (visited | captured | scout controlled)
- New `highlight-portal-history` plugin, with 4 basic highlighters:
  - visited/captured
  - scout controlled
  - **not** visited/captured
  - **not** scout controlled
- API: `history` field is now in portal data.


Aradiv:
@MysticJay I would suggest. Having 2 highlighters 1 for visited 1 for captured

MysticJay:
@Aradiv most users have been happy with yellow and red til now. Still some changed it to completely different settings,

In general I do not like the idea to add several history highlighters. I'd prefer one "highlight history" and when that get activated a small dialog, where you can select the values

plugins/highlight-portal-history.js
```diff
@@ -0,0 +1,33 @@
...
+  if (visited !== undefined) {
+    if (visited === 1) {
+      var opacity = .6;
+      data.portal.setStyle(
+        {fillColor: 'yellow' 
+        }
+      );
+    }
+    if (visited === 2 || visited ===3) {
+      var opacity = .6;
```
lachlan-harris: `opacity` defined twice, maybe `const` instead

lachlan-harris:
It would be beneficial to be able to toggle the opacity of these Portals on/off entirely, as part of a potential dialogue box.

core/code/entity_decode.js
```diff
@@ -73,7 +73,10 @@ function corePortalData(a) {
   }
 };
 
-var SUMMARY_PORTAL_DATA_LENGTH = 14;
+/* 
+history: bit_0 = visited, bit_1 = captured, bit_2 = scouted
+*/
+var SUMMARY_PORTAL_DATA_LENGTH = 19;
```
jujaga: May be worth defining this as a const as it doesn't appear to be re-assigned anywhere else.

plugins/highlight-portal-history.js
jujaga: The way the visited variable is being used, could consider rewriting this section as a switch/case instead.

core/code/entity_decode.js
MysticJay: The code will be converted to a new ES version, but that is a different ISSUE.
I don't think we should start to mix in the source for the core.

plugins/highlight-portal-history.js
MysticJay: see below comment

core/code/entity_decode.js
```diff
@@ -85,7 +88,8 @@ function summaryPortalData(a) {
     mission:       a[10],
     mission50plus: a[11],
     artifactBrief: parseArtifactBrief(a[12]),
-    timestamp:     a[13]
+    timestamp:     a[13],
+    history:       a[18]
```
lachlan-harris, line 92: It looks like this is actually an optional field, like `mods` or `resonators`. From what I can see, it would be better to define this as part of the detailed portal data below.

core/code/entity_decode.js
MysticJay: THIS IS WRONG! do not use!

anydot:
Alternative to highlighter: add a filters to show/hide visited/captured/scouted portals - similar to how users can choose to show/hide portals by their levels and if it's enl/res portal etc. At least for the captures/visited the major (if not the only one) scenario will be to see what are the uncaptured/unvisited portals around "me"

MysticJay:
What would be better in hiding the portals from the map by an additional Layer, while a highlighter can do the same?

anydot:
1/ Filters are more intuitive to use - it's what everyone is used to use
2/ You can only have a single highlighter on at a time
3/ It's confusing to have highlighter to effectively hide stuff (say visited portals)
4/ With filters you can separately choose between showing/hiding the captured and visited portals - is that something what can be done with a single highlighter?

le-jeu:
We need to improve this part
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/996e2949f73c479840dcefa5f54dce902a0e4f7d/core/code/map_data_render.js#L286-L316

Scenario:
 - User select a portal at zoom < 15, Intel response does not contain history info
 - User zoom on the portal at zoom >= 15, intel response contains history info but IITC skip the info since the data are up to date

johnd0e:
Related: #325

le-jeu:
imo since `history` is send by default (unless undefined), it should be in portalSummary
the Intel protocol is no longer : 
 - placeholder: `core`
 - portal from map entities `core,summary`
 - portal details : `core,summary,details`

but
 - portal from map entities `core,summary` OR `core,summary,null details,history`

MysticJay:
I checked the sizes of the array that `decodeArray.portalDetail` has to handle.
a.length can be 14,18, or 19
10 core
4 summary
4 details
1 extended

so I see 
`core`+ `summary` (14)
`core`+ `summary` + [null,null,null,null] + `extended` (19)
`core`+ `summary` + `details` (18)

Where the last is only available when asking portalDetails

core/code/portal_detail_display_tools.js
```diff
@@ -2,6 +2,19 @@
 // hand any of these functions the details-hash of a portal, and they
 // will return pretty, displayable HTML or parts thereof.
 
+window.getPortalHistoryDetails = function (d) {
+  let visited = ((d.history & 1) === 1) || ((d.history & 2) === 2);
+  let captured = ((d.history & 2) === 2);
+  let scouted = ((d.histroy & 4) === 4);
```
johnd0e: typo
And we still use `var`

johnd0e:
I'd suggest to change format of provided structure to include parsed values, e.g.:
```js
history: {
  _raw: 3,
  visited: true,
  captured: true,
  scouted: false,
  // ...
}
```
Motivation:
1. Parsed values are simple to use, and easy to understand for anyone, without digging in details.
The only reason I can imagine to prefer unparsed - packing data for transmission
2. `_raw` value can be kept as well, here, or we just rely on raw `ent`.
In case if more states will be added - we can easily add them as new properties.
We can even do it in some overridable function, like `window._parsePortalHistory`.
3. This way is much more safe than using bitflags directly: stock code may change bit order.
In this case the only change would be needed in single place - `window._parsePortalHistory`.

core/code/portal_detail_display_tools.js
MysticJay: fixed

MysticJay:
I prefer to leave the interpretation to a higher level. NIA will likely use the other BITs as well. This would mean that IITC has to be rebuild, while 3rd party plugins would simply be adjusted by the owners. A possible use could be "droned" (drone visited).

Also in this special case the returned data from NIA is still "under development": Following the mechanics captured means visited too. So the value should be "3", but currently they provide "2" (visited: no; captured: yes). Not sure if the same applies for visited and scouted. 

Now how should IITC interprete the raw data? 
- force visited to be true if captured to fix the bug
- leave the BITs as they are and handle correction at a higher level?

Finally IITC-CE shows a significant delay in addressing changes on intel and is having an unclear release strategy. This causes weird fixes and quick hacks as we see them now.

IMHO IITC should only provide the data as is. interpretation has to happen on usage (either in plugin or core module)

johnd0e:
> This would mean that IITC has to be rebuild, while 3rd party plugins would simply be adjusted by the owners.

You've missed point 2 - overridable function will solve it without rebuilding the iitc core.

> Now how should IITC interprete the raw data?

Fix it on our side.
Leave 'bits' as is but fix our own properties.

> IMHO IITC should only provide the data as is.

I do not agree.
We are parsing raw data instead of providing raw `ent`.

core/code/entity_decode.js
```diff
@@ -60,10 +60,20 @@ function parseArtifactDetail(arr) {
   };
 }
 
+function parseHistoryDetail(bitarray) {
+  if (bitarray == null) { return null; }
+  return {
+    _raw: bitarray,
+    visited:  !!(bitarray & (1|2|4)),
```
perringaiden: Visited should just be:
```suggestion
    visited:  !!(bitarray & 1),
```
There is no guarantee that scanned = visited for historical data, and the bit will always be set when appropriate.

core/code/entity_decode.js
```diff
@@ -60,10 +60,20 @@ function parseArtifactDetail(arr) {
...
+    captured: !!(bitarray & 2),
+    scanned:  !!(bitarray & 4),
```
perringaiden:
```suggestion
    scoutControlled:  !!(bitarray & 4),
```
The bit flag here does not indicate that a portal has been scanned.  It indicates that the portal has been held as Scout Controller.  Using "scanned" is misleading.

core/code/portal_detail_display_tools.js
```diff
@@ -2,6 +2,23 @@
...
 // will return pretty, displayable HTML or parts thereof.
 
+window.getPortalHistoryDetails = function (d) {
+  if (!d.history) {
+    return ('<div style="text-align: center; color: #ffce00">History missing</div>')
+  }
+
+  var visited = d.history.visited;
+  var captured = d.history.captured;
+  var scouted = d.history.scanned;
```
perringaiden:
```suggestion
  var scoutControlled = d.history.scoutControlled;
```

core/code/portal_detail_display_tools.js
```diff
@@ -2,6 +2,23 @@
...
+  var colors = {true:'#03fe03',false:'#ff4a4a'};
+
+  return ('<div style="text-align: center; color: #ffce00">History:'
+  + ' <span id="history_visited" style="color:'+ colors[visited] + '">visited</span> |'
+  + ' <span id="history_captured" style="color:'+ colors[captured] + '">captured</span> |'
+  + ' <span id="history_scanned" style="color:' + colors[scouted]  + '">scouted</span>'
```
perringaiden:
```suggestion
  + ' <span id="history_scoutControlled" style="color:' + colors[scoutControlled]  + '">scout controlled</span>'
```

core/code/entity_decode.js
johnd0e: > There is no guarantee that scanned = visited for historical data

Could you explain what may be different (and why) for historical data?

johnd0e: P.S.
In either case, even if we decide to fix (supposed) intel bug and make correction on our side, let's do that explicitly.
Like
```js
var parsed = {
  visited:  !!(bitarray & 1),
  // ...
};
if (parsed.captured || parsed.scoutControlled) {
  // fix (supposed) intel bug on our side
  parsed.visited = true;
}

File: core/code/map_data_render.js [hunk 9]
```diff
@@ -288,12 +288,16 @@ window.Render.prototype.createPortalEntity = function(ent) {
...
+
   // check if entity already exists
   if (ent[0] in window.portals) {
     // yes. now check to see if the entity data we have is newer than that in place
     var p = window.portals[ent[0]];
 
-    if (p.options.timestamp >= ent[1]) return; // this data is identical or older - abort processing
+    if (!data.history || p.options.data.history === data.history)
+      if (p.options.timestamp >= ent[1])
+        return; // this data is identical or older - abort processing
```
johnd0e:
```suggestion
      if (p.options.timestamp >= ent[1]) {
        return; // this data is identical or older - abort processing
      }
```

core/code/map_data_render.js
```diff
@@ -303,25 +307,27 @@ window.Render.prototype.createPortalEntity = function(ent) {
 
     previousData = p.options.data;
 
+    // preserve history
+    if (!data.history)
+      data.history = previousData.history;
```
johnd0e:
```suggestion
    if (!data.history) {
      data.history = previousData.history;
    }
```

core/code/map_data_render.js
```diff
@@ -303,25 +307,27 @@ window.Render.prototype.createPortalEntity = function(ent) {
...
+  var latlng = L.latLng(data.latE6/1E6, data.lngE6/1E6);
 
   var dataOptions = {
     level: portalLevel,
     team: team,
     ent: ent,  // LEGACY - TO BE REMOVED AT SOME POINT! use .guid, .timestamp and .data instead
     guid: ent[0],
     timestamp: ent[1],
-    data: data
+    data: data,
```
johnd0e: Unintentional?

core/code/portal_detail_display.js
```diff
@@ -225,7 +225,7 @@ window.getPortalMiscDetails = function(guid,d) {
 
     if (d.artifactBrief && d.artifactBrief.target && Object.keys(d.artifactBrief.target).length > 0) {
       var targets = Object.keys(d.artifactBrief.target);
-//currently (2015-07-10) we no longer know the team each target portal is for - so we'll just show the artifact type(s) 
+//currently (2015-07-10) we no longer know the team each target portal is for - so we'll just show the artifact type(s)
```
johnd0e: Either do not touch trailing space, or fix indent as well.

core/code/entity_decode.js
MysticJay: Attention!
At the moment 'scanned' layer is wrongly titled as 'Scout Controlled' in stock intel!

"Scout Controller" is a volatile status, that the agent looses sooner or later on interaction of other agents with the same portal.
"scanned" is a permanent status, set when the agent submits the first 3D-Scan of a portal.

core/code/entity_decode.js
MysticJay: ATM we decided to fix the value for `visited`
A captured portal is always a visited portal.
And `scanned` portals do the same. While this might not be true for for some early scans. The situation where this causes a wrong value for "visited" will be very rare. Agents usually do not SCAN a portal but "forget" to interact with it.
A plugin can still revert to using `ent`

core/code/portal_detail_display_tools.js
MysticJay: Attention!
At the moment 'scanned' layer is wrongly titled as 'Scout Controlled' in stock intel!
"Scout Controller" is a volatile status, that the agent looses sooner or later on interaction of other agents with the same portal.
"scanned" is a permanent status, set when the agent submits the first 3D-Scan of a portal.
@MysticJay

core/code/portal_detail_display_tools.js
MysticJay: Attention!
At the moment 'scanned' layer is wrongly titled as 'Scout Controlled' in stock intel!
"Scout Controller" is a volatile status, that the agent looses sooner or later on interaction of other agents with the same portal.
"scanned" is a permanent status, set when the agent submits the first 3D-Scan of a portal.
@MysticJay

plugins/highlight-portal-history.js
MysticJay: Old status

johnd0e (changes requested):

core/code/portal_detail_display_tools.js
johnd0e: Better have this as comment in source code

core/code/map_data_render.js
MysticJay: @le-jeu ?

core/code/portal_detail_display_tools.js
le-jeu: from field data, the data `scanned` or `scoutControlled` is set to 1 if the player gains the scout controlled at one point.
The fact that one gains scout controlled is permanent and counted in the agent stats for the "Scout Controller" medal.

This follows the same principle as `captured`, one looses the own of a portal but the capture stays

core/code/entity_decode.js
MysticJay: I think we better leave that as an operation on the bitarray for speed reasons, but add a comment.

perringaiden: The point is you're making an unnecessary decision for the purposes of what?  Intentionally making mistakes?

If a scout controlled portal is always visited, the first bit will always be set.
If a captured portal is always visited, the first bit will always be set.

You will never need a "catchup method" by setting it when the others are true, and all you're doing is introducing a bug, then claiming it's irrelevant.

core/code/portal_detail_display_tools.js
perringaiden: image

Scout Controller is a volatile status.
Scout Controlled is a historical statement.

Captured does not mean you *currently* hold the portal.  Visited does not mean you're *at* the portal.  Scout Controlled means that at some time you have been Scout Controller.

Scanning directly is irrelevant to the historical status.  Scanning a portal does not award you Scout Controller, or the Scout Controlled historical status.  Exceeding the number of scans of another person does.

core/code/portal_detail_display.js
```diff
@@ -37,8 +37,6 @@ window.renderPortalDetails = function(guid) {
   var data = portal.options.data;
   var details = portalDetail.get(guid);
 
-  var historyDetails = getPortalHistoryDetails(data);
-
   // details and data can get out of sync. if we have details, construct a matching 'data'
   if (details) {
     data = getPortalSummaryData(details);
```
le-jeu: `data` is overwrited here
and portalDetail.get doesn't give history yet

core/code/map_data_request.js
```diff
@@ -74,7 +74,7 @@ window.MapDataRequest = function() {
   var _this = this;
   addHook('portalDetailLoaded', function(data){
     if(data.success) {
-      _this.render.processGameEntities([data.ent]);
+      _this.render.processGameEntities([data.ent], 'detailed');
```
le-jeu: should we call `createPortalEntity(data.ent, 'detailed)` instead ?

core/code/map_data_render.js
```diff
@@ -106,7 +106,7 @@ window.Render.prototype.processDeletedGameEntityGuids = function(deleted) {
 
 }
 
-window.Render.prototype.processGameEntities = function(entities) {
+window.Render.prototype.processGameEntities = function(entities, details) {
```
le-jeu: `processGameEntities` could be call only when we get entites (`getEntities` or `mapInject` hook) and remove the `details` parameters

core/code/map_data_render.js
```diff
@@ -131,7 +131,7 @@ window.Render.prototype.processGameEntities = function(entities) {
     var ent = entities[i];
 
     if (ent[2][0] == 'p' && !(ent[0] in this.deletedGuid)) {
-      this.createPortalEntity(ent);
+      this.createPortalEntity(ent, details);
```
le-jeu: following previous comment: replace `details` by `'extended'`

core/code/map_data_render.js
le-jeu:
```suggestion
window.Render.prototype.processGameEntities = function(entities) {
```

core/code/map_data_render.js
le-jeu:
```suggestion
      this.createPortalEntity(ent, 'extended');
```

core/code/map_data_request.js
le-jeu:
```suggestion
      _this.render.createPortalEntity(data.ent, 'detailed');
```

johnd0e: Good point!

core/code/entity_decode.js
```diff
@@ -60,10 +60,21 @@ function parseArtifactDetail(arr) {
   };
 }
 
+function parseHistoryDetail(bitarray, required) {
+  if (bitarray == null && !required) return null;
+  if (bitarray == null) bitarray = 0;
```
le-jeu: This is handled at call

```suggestion
function parseHistoryDetail(bitarray) {
```

core/code/entity_decode.js
```diff
@@ -89,53 +102,64 @@ function summaryPortalData(a) {
...
   }
 
-  //TODO look at the array values, make a better guess as to which index the mods start at, rather than using the hard-coded SUMMARY_PORTAL_DATA_LENGTH constant
+  var data = corePortalData(a);
 
+  if (a.length >= SUMMARY_PORTAL_DATA_LENGTH) {
+    $.extend(data, summaryPortalData(a));
+  }
+  if (a.length >= DETAILED_PORTAL_DATA_LENGTH) {
+    if (details !== 'extended' || a[SUMMARY_PORTAL_DATA_LENGTH]) {
```
le-jeu: why not parse it whenever there is data to parse (not `null`)?
```suggestion
    if (a[SUMMARY_PORTAL_DATA_LENGTH]) {
```

core/code/portal_detail.js
```diff
@@ -33,7 +33,7 @@ var handleResponse = function(deferred, guid, data, success) {
 
   if (success) {
 
-    var dict = decodeArray.portalDetail(data.result);
+    var dict = decodeArray.portal(data.result, 'detailed');
```
le-jeu, line 36: note: `data.result` is parsed a second time with 
```window.runHooks ('portalDetailLoaded', {guid:guid, success:success, details:dict, ent:ent});```

File: core/code/portal_detail_display_tools.js [hunk 20]
```diff
@@ -2,16 +2,33 @@
...
+  var visited = d.history.visited;
+  var captured = d.history.captured;
+  var scoutControlled = d.history.scoutControlled;
+  var colors = {true:'#03fe03',false:'#ff4a4a'};
+
+  return ('<div style="text-align: center; color: #ffce00">History:'
+  + ' <span id="history_visited" style="color:'+ colors[visited] + '">visited</span> |'
+  + ' <span id="history_captured" style="color:'+ colors[captured] + '">captured</span> |'
+  + ' <span id="history_scout_controlled" style="color:' + colors[scoutControlled]  + '">scout controlled</span>'
+  + '</div>');
```
le-jeu: todo: move style to CSS

core/code/entity_decode.js
johnd0e: It is opposite: parsing is forced even if there is no data.
Reason: throw error to make us know about weird data case.

johnd0e: Alternatively, we could use `console.warn` here.
@le-jeu?

core/code/portal_detail.js
johnd0e: Should we add some remark to sources?

core/code/portal_detail_display_tools.js
johnd0e: Are you going to enhance this PR?
Or this note is just a task for future?

le-jeu: I can do a quick sub PR for this

le-jeu: https://github.com/MysticJay/ingress-intel-total-conversion/pull/7

core/code/portal_detail.js
le-jeu: I suggest to reverse the order of the code here:
creating an entity first and then , use the the parsed data to populate `portalDetails`'s cache
like in this commit https://github.com/le-jeu/ingress-intel-total-conversion/commit/bf916d18c6aa3406cd4409fce58432215e772c5e (forgetting the specificity of this branch)

this can be in this PR or in a later one

core/code/entity_decode.js
le-jeu: hm, if `a[SUMMARY_PORTAL_DATA_LENGTH]` is null the code is likely to raise an exception  due to `parseDetails` assumming it is an Array

core/code/portal_detail.js
johnd0e: I prefer later.

core/code/entity_decode.js
johnd0e: Exactly.
So question: is it acceptable (if intel changes data - we adapt).
Or better avoid exception, but send just warning.

Exception is better because it's harder to miss.
But warning is better because app is not breaking for everyone.

core/code/entity_decode.js
```diff
@@ -89,53 +102,64 @@ function summaryPortalData(a) {
...
+      $.extend(data, detailsPortalData(a));
+    }
+  }
```
le-jeu:
```suggestion
    if (a[SUMMARY_PORTAL_DATA_LENGTH]) {
      $.extend(data, detailsPortalData(a));
    } else if (details !== 'extended') {
      log.warn('Portal details are missing  - portal details likely broken!');
    }
  }
```

core/code/entity_decode.js
le-jeu: better warn and don't parse than parse and crash in the middle of something
https://github.com/IITC-CE/ingress-intel-total-conversion/pull/436/files/d4834db670a9177471a7834ee18785c48558705a#r574442530

core/code/portal_detail.js
MysticJay: the duplicate parsing has been analyzed by me and @McBen in November.
We have been looking at the complete function stack and found that part of the routines get executed twice.

renderPortalDetails is the culprit.
it starts to analyze, requests portalDetails, which executes runhooks('portalDetailsLoaded')
then renderPortalDetails continues  to analyze the data and finishes with an other runhooks('portalDetailsLoaded').

The effect is that on each portal that gets clicked (selected) we get two hook calls with:
1) oldGUID, newGUID
2) newGUID, newGUID

This also causes trouble in several 3rd party plugins, that now have to filter calls of the second type.

This is the call stack (read reverse). the first 6 lines have already been executed where the log is indented:
```
someFunction (source-server0.conn3.child37/source28869#20)
runHooks (source-server0.conn3.child37/source28774#20656)
each (source-server0.conn3.child37/source28774#18713)
runHooks (source-server0.conn3.child37/source28774#20654)
selectPortal (source-server0.conn3.child37/source28774#23179)
renderPortalDetails (source-server0.conn3.child37/source28774#22885)
handleResponse (source-server0.conn3.child37/source28774#22829)
doRequest (source-server0.conn3.child37/source28774#22849)
onSuccess (source-server0.conn3.child37/source28774#25469)
c (source-server0.conn3.child37/source28774#18713)
fireWith (source-server0.conn3.child37/source28774#18713)
l (source-server0.conn3.child37/source28774#18713)
o (source-server0.conn3.child37/source28774#18713)
send (source-server0.conn3.child37/source28774#18713)
ajax (source-server0.conn3.child37/source28774#18713)
postAjax (source-server0.conn3.child37/source28774#25490)
doRequest (source-server0.conn3.child37/source28774#22848)
request (source-server0.conn3.child37/source28774#22860)
        someFunction (source-server0.conn3.child37/source28869#20)
        runHooks (source-server0.conn3.child37/source28774#20656)
        each (source-server0.conn3.child37/source28774#18713)
        runHooks (source-server0.conn3.child37/source28774#20654)
        selectPortal (source-server0.conn3.child37/source28774#23179)
renderPortalDetails (source-server0.conn3.child37/source28774#22892)
handler_portal_click (source-server0.conn3.child37/source28774#21471)
fire (source-server0.conn3.child37/source28774#4001)
_fireDOMEvent (source-server0.conn3.child37/source28774#7816)
_fireEvent (source-server0.conn3.child37/source28774#15911)
_onClick (source-server0.conn3.child37/source28774#15853)```

MysticJay (commented):
I feel not entitled to review the changes done regarding .portalDetail -> .portal

johnd0e (approved):

ryantheleach:
Layers and Highlighters are quickly becoming not fit for purpose with all the data people would like to see at once.

We need to seriously consider some sort of highlighting / filtering engine overhaul, that allows users to customize it how it looks for them.

E.g. A user may want to style portals that are available for capturing (e.g. of the opposite faction or neutral) and not use the highlighter. They may want it to be a decoration, a highlight, or a complete style change.

It's possible that they also want to see something else at the same time, to the point where it's impossible for the developers to predict. Having some sort of basic rules / customization to let users view portals in a variety of methods would be best.

MysticJay:
@ryantheleach while the request is not new, we face some prerequisites:
- IITC is still based on NIA's intel and what the intel servers deliver.
portal status 
- the portal markers are organized in 16 different layers based on team and level.
- on a change the marker is recreated in a different layer
game related status 
- NIA might add additional information to each portal that is not related to the basic gameplay: Ornaments (anomalies, Scouting volatiles, shards). These ornaments add an additional layer to the map. removing an ornament from its layer, when hiding the portal is easy. To hide a portal based on the ornaments requires to search all 16 basic layers for the portal to be removed. And this has to be done for each portal on the map. re-adding the portal would mean to reload the map. This would slow down IITC and thus cause additional load on the INTEL-servers. IITC might become unusable in anomalies.
user related status 
- new history information is generated by NIA. This information is static, but OTOH different for each user. Many users looking at the same place can not be served with the same (cached) data. They all need to be handled individually. While this does not bother IITC in the first place, thefeature might get turned off, when the servers are stressed (e.g. anomaly)
IITC or plugin generated status information
- portal health, player name, innoculation, etc. are also possible filters that could be used
- 3rd party plugins rely on the current design, so a change must be compatible!

So what are the ways to decorate?
- modify portal marker (circle-properties) used by highlighters. combining existing highlighters has already been shown by @JaiPerduLeJeu 
- additional layers for ornaments/shapes we also have examples for that, not at least with volatile ornaments.

We need to consider server-load and IITCs speed and carefully think about the use-cases (as operator, field planner or UPC/UPV hunter). A combination of these roles might happen, but should be very rare.

MysticJay:
@MysticJay May I also consider layer-profiles plugin in ZasoItems. This allows switch between different sets of layers quickly.
https://github.com/MysticJay/ZasoItems.CE/blob/master/layers-profiles.user.js
check here for the other plugins:
https://github.com/MysticJay/ZasoItems.CE

johnd0e:
This is closed PR. Please open new Issue.
Related: #446, #142

Issue #195, opened 2019-05-09
Portal tooltip


johnd0e:
![image](https://user-images.githubusercontent.com/1838643/57460341-9ce14c00-727d-11e9-8f73-77addfd0e79f.png)

Here is concept:
<details>

```portal-tooltip.user.js
// ==UserScript==
// @id             portal-tooltip
// @name           IITC plugin: Portal tooltip
// @category       Info
// @version        0.1
// @namespace      https://github.com/IITC-CE/ingress-intel-total-conversion
// @updateURL      none
// @downloadURL    none
// @description    Show tooltip with portal name on mouse hover 
// @include        https://intel.ingress.com/*
// @grant          none
// ==/UserScript==

function wrapper(plugin_info) {
// ensure plugin framework is there, even if iitc is not yet loaded
if(typeof window.plugin !== 'function') window.plugin = function() {};

// PLUGIN START ////////////////////////////////////////////////////////


function tooltip (data) {
  var title = data.portal.options.data.title;
  if (title) { data.portal.bindTooltip(title); }
}

function setup () {
  window.addHook('portalAdded', tooltip);
  window.addHook('portalDetailsUpdated', tooltip);
}

// PLUGIN END //////////////////////////////////////////////////////////


setup.info = plugin_info; //add the script info data to the function as a property
if(!window.bootPlugins) window.bootPlugins = [];
window.bootPlugins.push(setup);
// if IITC has already booted, immediately run the 'setup' function
if(window.iitcLoaded && typeof setup === 'function') setup();
} // wrapper end
// inject code into site context
var script = document.createElement('script');
var info = {};
if (typeof GM_info !== 'undefined' && GM_info && GM_info.script) info.script = { version: GM_info.script.version, name: GM_info.script.name, description: GM_info.script.description };
script.appendChild(document.createTextNode('('+ wrapper +')('+JSON.stringify(info)+');'));
(document.body || document.head || document.documentElement).appendChild(script);
```
</details>

Questions are:
1. Should we implement something similar in core? (IMHO plugin is enough)
2. How better style portal tooltips? (propose your css)
3. What if style'd depend on faction (or anything else)?

Edit: consider using https://github.com/makinacorpus/Leaflet.OverIntent

Tursaanpoika:
Works well as plain tooltip, using it just as you posted it above. Styling is also ok for my use - but in case you go for styling by faction, make it simple - color the text maybe?

PR #437, opened 2021-02-06
Permanent portal entities

le-jeu:
Address  #325 

# Changes
 
Extend the CircleMarker class to PortalMarker class following the same idea of #251

Portal entities are only deleted when removed from rendering:
 - portal deleted according to getEtentities data (does this really occurs?)
 - no selected and absent from getEntities data (zoom <15, not in view, not a endpoint)
 
On any other situtation, the portal marker is the same and only receive updates until removed from rendering.

## class `L.PortalMarker`
  Methods:
* `setStyle(style)` (deprecated, for compatibility with old highlighters)
      don't apply a style, but save it for later when calling `setMarkerStyle`
* `setMarkerStyle(style)`
    * compute base portal style according to details and level
    * use the given style on top of it
    * call the highlighter to update further the style
    * if selected, force the color
    * apply the style
* `setSelected(details)`
    * change selected intern state and call setMarkerStyle()
* `willUpdate(details)`
    * given details (from decodeArray.portal), return true if the details are more recent or more detailled than the known details 
* `updateDetails(details)`
    * apply new details
    * maintain permanent data (history)
    * handle partial update with identical timestamp (mods, resonators...)
    * update options (compatibility)
    * update the marker on the map
    * if selected, update the side bar
* `getDetails()`
    * return portal details
* `hasFullDetails()`
    * return true if the marker store mods/resonators/... details 
* `renderDetails()`
    * call `renderPortalDetails` (even if unselected) while avoiding a call to `selectPortal` 

## Hooks `portalSelected`

Add a new field `event` that specified the origin of the event.
This is mainly to distinguish a selection that comes from a call to `renderPortalDetails` and a click/touch event on the marker from the user.

### Before

on `portalDetail.request`, the data is parsed, store, used for render portal details in the side bar if the portal is selected, then the `portalDetailLoaded` hook trigger the creation of a new portal entity.

portal entity only stores partial data (summary only but full raw)

### Now
on `portalDetail.request`, we create or update a portal entity,  this part renders the portal details if needed, store the portal details in cache, run the `portalDetailLoaded` hook.

portal detail cache is used for incomplete portal entity upon creation

**Gain:** the portal entity always has the most recent data, details are parsed once, the portal entity is sure to exist for the hook.

~**Missing:** if the portal changed before details in cache expires, the code won't fetch new details unless a direct call to `portalDetail.request` (is it true before also?)~ see a587775

## Placeholder entities

Placeholder portal entities are introduced for link/field endpoints.
Fields and links have an unused timestamp.
The PR uses the timestamp from link/field to handle portal updates as any other portal entity.

## Sidenotes

Portal timestamp change on portal edit (title, location, portal main picture).

## deprecated:
  * `window.createMarker`
  * `window.setMarkerStyle`

From #251
## related functions (next candidates for refactoring/deprecation)
  * `window.setPortalIndicators`
  * `window.selectPortal`
  * `window.portalMarkerScale`
  * `window.getMarkerStyleOptions`
  * `window.renderPortalDetails`

### DIFF
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -345,6 +345,13 @@ window.Render.prototype.createPortalEntity = function(ent, details) { // details
   } else {
     marker = createMarker(latlng, data);
 
+    // in case of incomplete data while having fresh details in cache, update the portal with those details
+    if (portalDetail.isFresh(guid)) {
+      var oldDetails = portalDetail.get(guid);
+      if (marker.willUpdate(oldDetails))
+        marker.updateDetails(oldDetails);
+    }
+
     window.runHooks('portalAdded', {portal: marker});
 
     window.portals[data.guid] = marker;
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -286,34 +286,6 @@ window.Render.prototype.createPortalEntity = function(ent, details) { // details
   var data = decodeArray.portal(ent[2], details);
   var guid = ent[0];
 
-  // check if entity already exists
-  var oldPortal = guid in window.portals;
-
-  if (oldPortal) {
-    // yes. now check to see if the entity data we have is newer than that in place
-    var p = window.portals[guid];
-
-    if (!data.history || p.options.data.history === data.history) {
-      if (p.options.timestamp > ent[1]) {
-        return p; // this data is older - abort processing
-      }
-
-      if (p.options.timestamp == ent[1] && p.hasFullDetails()) // this data is identical - abort processing
-        return p;
-    }
-
-    // the data we have is newer. many data changes require re-rendering of the portal
-    // (e.g. level changed, so size is different, or stats changed so highlighter is different)
-
-    // remember the old details, for the callback
-    previousData = p.getDetails();
-
-    // preserve history
-    if (!data.history) {
-      data.history = previousData.history;
-    }
-  }
-
   // add missing fields
   data.guid = guid;
   if (!data.timestamp)
@@ -321,6 +293,25 @@ window.Render.prototype.createPortalEntity = function(ent, details) { // details
 
   // LEGACY - TO BE REMOVED AT SOME POINT! use .guid, .timestamp and .data instead
   data.ent = ent;
+
+  // check if entity already exists
+  var oldPortal = guid in window.portals;
+
+  if (oldPortal) {
+    // yes. now check to see if the entity data we have is newer than that in place
+    var p = window.portals[guid];
+
+    if (!p.willUpdate(data)) {
+      // this data doesn't bring new detail - abort processing
+      return p;
+    }
+
+    // the data we have is newer. many data changes require re-rendering of the portal
+    // (e.g. level changed, so size is different, or stats changed so highlighter is different)
+
+    // remember the old details, for the callback
+    previousData = $.extend(true, {}, p.getDetails());
+  }
 
   var latlng = L.latLng(data.latE6/1E6, data.lngE6/1E6);
 
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -291,7 +291,9 @@ window.Render.prototype.createPortalEntity = function(ent, details) { // details
   var data = decodeArray.portal(ent[2], details);
 
   // check if entity already exists
-  if (ent[0] in window.portals) {
+  var oldPortal = ent[0] in window.portals;
+
+  if (oldPortal) {
     // yes. now check to see if the entity data we have is newer than that in place
     var p = window.portals[ent[0]];
 
@@ -302,18 +304,14 @@ window.Render.prototype.createPortalEntity = function(ent, details) { // details
 
     // the data we have is newer. many data changes require re-rendering of the portal
     // (e.g. level changed, so size is different, or stats changed so highlighter is different)
-    // so to keep things simple we'll always re-create the entity in this case
 
     // remember the old details, for the callback
-
     previousData = p.options.data;
 
     // preserve history
     if (!data.history) {
       data.history = previousData.history;
     }
-
-    this.deletePortalEntity(ent[0]);
   }
 
   var portalLevel = parseInt(data.level)||0;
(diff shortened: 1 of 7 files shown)


johnd0e:
See also #251

core/code/portal_marker.js
```diff
@@ -10,14 +10,15 @@ var portalBaseStyle = {
 
 // portal hooks
 function handler_portal_click (e) {
-  e.target.select(true);
+  window.selectPortal(e.target.options.guid, e.type);
+  window.renderPortalDetails(e.target.options.guid, true)
 }
 function handler_portal_dblclick (e) {
-  e.target.select(true);
+  handler_portal_click(e);
```
johnd0e: BAD!

le-jeu: what is so bad for you to capslock ?

johnd0e:
Please consider this use case: #195
It would be great if the tooltip will be kept open even during internal data updating.

core/code/portal_marker.js
johnd0e: click/dblclicl/contexmenu - are separate actions.
Don't call one from another, even if currently it does the job.
Better make separate function for these calls:
```
  window.selectPortal(e.target.options.guid, e.type);
  window.renderPortalDetails(e.target.options.guid, true)
```  
Also I suspect that we we can refactor code to leave only one of these.

le-jeu:
It is, using standard `portal.bindPopup(content).openPopup()`

johnd0e:
> It is, using standard `portal.bindPopup(content).openPopup()`

But what will happen when placeholder will be replaced with real entity? What if it will be moved from one faction to another, etc.
(Of course I will test them myself too, when PR will be ready)

le-jeu:
placeholders are not destroyed but updated when we get real data.

> Portal entities are only deleted when removed from rendering:
>  - portal deleted according to getEtentities data (does this really occurs?)
>  - no selected and absent from getEntities data (zoom <15, not in view, not an endpoint)
> 
> On any other situtation, the portal marker is the same and only receive updates until removed from rendering.


the main feature (permanent portal entities) is ready for testing

the `portalSelected` related code (`renderPortalDetails()` and `selectPortal()`) needs better plumbing but the behavior is saner than before

le-jeu (commented):
Link/field timestamp may be higher than the portal timestamp, breaking the logic
Maybe use a timestamp minus decay time...

core/code/portal_detail_display.js
```diff
@@ -9,8 +9,8 @@ window.resetScrollOnNewPortal = function() {
   }
 };
 
-window.renderPortalDetails = function(guid) {
-  selectPortal(window.portals[guid] ? guid : null);
+window.renderPortalDetails = function(guid, dontSelect) {
```
johnd0e:
```suggestion
window.renderPortalDetails = function(guid) {
```
I haven't reviewed this PR as a whole, but still believe that here we need slightly different logic:
**do not select portal if it is already set as current** without additional arguments: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/442#issuecomment-781241495

In case if we will found that it is still required in some special case, then we better add opposite:
```suggestion
window.renderPortalDetails = function(guid, forceSelect) {
```

le-jeu: Second looks saner.
My though was to not break (unknonwn) plugins that would assume renderPortalDetails triggers a `portalSelected` event as long as the portal has its own marker.
But I bet plugins relying on this will eventually update to match the needs

johnd0e: > But I bet plugins relying on this will eventually update to match the needs

So am I.
Let's try to remove `dontSelect`, may be taht will be best for all.

core/code/map_data_render.js
```diff
@@ -267,125 +264,101 @@ window.Render.prototype.createPlaceholderPortalEntity = function(guid,latE6,lngE
...
+  var marker = undefined;
+  if (oldPortal) {
+    // update marker style/highlight and layer
+    marker = window.portals[data.guid];
+    // remove portal from its faction/level specific layer
+    this.removePortalFromMapLayer(marker);
+
+    marker.updateDetails(data);
+
+    window.runHooks('portalAdded', {portal: marker, previousData: previousData});
```
le-jeu, line 339: the hook is misleading now since the marker is the same marker. Only the data changed.
In this case, we are closer to the `portalDetailLoaded` but with eventually no details (i.e. from core to summary/extended)

johnd0e: > the hook is misleading

But that was expected, that's why `previousData` provided. I suppose we should leave it as it.
May be we should add some property to denote placeholders.

But we can revise that from scratch when implementing #217.

johnd0e:
> breaking the logic

Could you point exact places in code? (I haven't studied this PR yet).

le-jeu:
I introduced link/field timestamp as placeholder timestamp instead of the value `0`.
The value `0` was set to be sure that any complete data (from `getEntities` or `getPortalDeails`) takes priority over placeholder data (that are forged by iitc).

In many cases, the link/field timestamp is more recent that the portal timestamp. The previous rule that portal data is ignored if timestamp is older prevents the marker to update from detailed data (while getPortalDetails are always up to date to my knowledge).

I fix this following this new rule about placeholder (provided the field/link timestamp):
 - if the placeholder is more recent and has a different team, it takes priority
 - if the placeholder is older or has the same team, no update
That is, trust the placeholder team as long it's more recent that the known portal data.

Uses cases: 
 - a selected portal with detail loaded, but outside the view, this portal switches team and a link is thrown, the placeholder will update the selected portal data.
 - intel send a link while it has just been destroyed (but details are correct), the placeholder won't update the portal details

johnd0e:
> if the placeholder is more recent and has a different team, it takes priority

And what if that portal had links/fields attached?

P.S.
If we continue this way, than we could parse messages plexts digging for updates...

le-jeu:
> And what if that portal had links/fields attached?

I don't understand your question
placeholders come from links/fields, so they do have links/fields attached
that's the point of this rule, if intel says that there is a blue field at T+1 on the portal A, while the data about portal A being green at T is invalidate (and we don't get data on this portal because of request's bounding box)

The current code is also trusting the team of placeholders, whatever the timestamp of the attached field/links.

> P.S.
> If we continue this way, than we could parse messages plexts digging for updates...

A plugin could do this to update the map before getEntities...

johnd0e:
> I don't understand your question

Imaging that we have Green portal with links and fields.
And then we get fresher info where same portal (as placeholder) is already Blue.
In this moment our data become inconsistent, as Blue portal cannot have Green links and fields.
So I wonder what are you proposing for such case.

le-jeu:
> > I don't understand your question
> 
> Imaging that we have Green portal with links and fields.
> And then we get fresher info where same portal (as placeholder) is already Blue.
> In this moment our data become inconsistent, as Blue portal cannot have Green links and fields.
> So I wonder what are you proposing for such case.

Since data is fresher, the portal is updated to blue, the links may stays until the the next getEntities request

you should give a look to `portalMarker.willUpdate`

johnd0e:
(Haven't seen the code yet, but) I suppose that portals not only change color, but also become placeholder, loosing details about resos, mods (though titles and history info may be kept).

But drawn links/fields (and data in window.links/fields) become not consistent, with consequences that are hard to predict now (e.g. some plugins may rely on consistent data).

Perhaps we also could try to fix that by forcing additional requests for map data of our interest.

le-jeu:
> But drawn links/fields (and data in window.links/fields) become not consistent, with consequences that are hard to predict now (e.g. some plugins may rely on consistent data).

I doubt any plugin relies on consistency since it's not guaranteed yet. But this is manageable as an addition to upstream

johnd0e:
conflicts!

PR #440, opened 2021-02-12
Portal history - fixing color logic

MysticJay:
fixed color logic, which was broken after revamping the code


modos189:
The last commit with a fix is cherry-picked in master. Please make further changes based on master branch

MysticJay:
@modos189 nope, i specially opened a new Branch and PR for that fix.

plugins/highlight-portal-history.js
```diff
@@ -17,33 +17,35 @@ function visited (data) {
   if (!history || !(history.visited || !history.captured)) {
     return;
   }
-
   if (history.captured) {
     setStyle(data, 'red', 1);
-  } else if (history.visited) {
+    return;
+  }
+  if (history.visited) {
```
johnd0e, line 24: Explain this code change.
I do no see any difference in functionality

plugins/highlight-portal-history.js
```diff
@@ -17,33 +17,35 @@ function visited (data) {
...
   if (history.captured) {
-    setStyle(data, 'white', 0);
-  } else if (history.visited) {
+    return;
+  }
+  if (history.visited) {
     setStyle(data, 'yellow', 1);
+    return;
   }
+  setStyle(data, 'red', 1);
```
johnd0e: Again, why all this code? We only want to portals with both capture/visited in false `state`.

johnd0e:
I suppose that should be changed: https://github.com/IITC-CE/ingress-intel-total-conversion/compare/master...johnd0e:Portal-History

plugins/highlight-portal-history.js
MysticJay: We want
No history -> no change
captured -> no change, shrink (radius: 4) or even hide the portal (radius: 0) (undecided yet)
visited only -> yellow & return
all others -> red

So to be able to handle all 4 situations the code should not interleave in if-else clauses.

Also I think the code is easier to understand and even a user is able to change the styles to his personal flavor.

johnd0e: Ok then

PR #441, opened 2021-02-14
portal-history: separate styles

johnd0e:
Separate style for every included here highlighter.
Any style may be overridden without touching main logic:
- inline (`portalsHistory.styles` object)
- or in separate plugin, with code like this:
  ```js
  var setup = function () {
    var portalsHistory = window.plugin.portalHighlighterPortalsHistory;
    if (!portalsHistory || !portalsHistory.styles) {
      console.warn('highlight-portal-history plugin is required!');
      return;
    }
    portalsHistory.styles.common.radius = 10;
    portalsHistory.styles.commonOther.fillOpacity = 0.1;
  };
  ```

Known styles:
```
  captured <------------ |
  visitTarget <--------- |
  scoutControlled <----- |
  scoutControllTarget <- | <== marked ===== |
                                            | <= common
  visited <------------- | <== semiMarked = |
  captureTarget <------- |
```
```
  otherVC <------- | <= commonOther (empty by default)
  otherScout <---- |
  otherNotVC <---- |
  otherNotScout <- |
```

### DIFF
--- a/plugins/highlight-portal-history.js
+++ b/plugins/highlight-portal-history.js
@@ -10,13 +10,14 @@ var portalsHistory = {};
 window.plugin.portalHighlighterPortalsHistory = portalsHistory;
 
 portalsHistory.styles = {
-  marked: {
-    fillColor: 'red',
+  common: {
     fillOpacity: 1
   },
+  marked: {
+    fillColor: 'red'
+  },
   semiMarked: {
-    fillColor: 'yellow',
-    fillOpacity: 1
+    fillColor: 'yellow'
   }
 };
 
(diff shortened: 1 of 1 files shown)


MysticJay:
The code is complicated without need.
At two places a new function gets inserted that only call one global function.
The construction of the styles from various parts is complex and not easy to maintain.

Create direct named style-objects that hold all wanted properties
and use them directly in `data. portal.setStyle(<name of style>)`

johnd0e:
`notVisited` has a bug now: we need to check `!visited` first, then `!captured`.
Will fix tomorrow.

plugins/highlight-portal-history.js
```diff
@@ -1,73 +1,104 @@
...
+    data.portal.setStyle(s.otherNotScout);
+  }
 };
 
-var setup = function () {
-  window.addPortalHighlighter('History: visited/captured', window.plugin.portalHighlighterPortalsHistory.visited);
-  window.addPortalHighlighter('History: not visited/captured', window.plugin.portalHighlighterPortalsHistory.notVisited);
-  window.addPortalHighlighter('History: scout controlled', window.plugin.portalHighlighterPortalsHistory.scoutControlled);
-  window.addPortalHighlighter('History: not scout controlled', window.plugin.portalHighlighterPortalsHistory.notScoutControlled);
+function inherit (parentName, childNames) {
```
modos189, line 88:
```suggestion
// Creating styles based on a given template
function inherit (parentName, childNames) {
```

plugins/highlight-portal-history.js
```diff
@@ -1,73 +1,104 @@
...
+  var styles = portalsHistory.styles;
+  childNames.forEach(function (name) {
+    styles[name] = L.extend(L.Util.create(styles[parentName]), styles[name]);
```
modos189, line 92:
```suggestion
    // Extension of _styles_ with a new _name_ object, created based on _parentName_ object.
    styles[name] = L.extend(L.Util.create(styles[parentName]), styles[name]);
```

modos189 (changes requested):

modos189 (approved):

PR #444, opened 2021-02-18
API: enhance decodeArray.portal

johnd0e:
Enhance #436

- When `details` is not defined - data is checked over all known cases
- `details` arg is added to `processGameEntities`, as it may be called from different places,
  e.g. via `mapDataEntityInject` hook, like in `cache-portals-on-map` plugin

### DIFF
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -106,7 +106,7 @@ window.Render.prototype.processDeletedGameEntityGuids = function(deleted) {
 
 }
 
-window.Render.prototype.processGameEntities = function(entities) {
+window.Render.prototype.processGameEntities = function(entities, details) { // details expected in decodeArray.portal
 
   // we loop through the entities three times - for fields, links and portals separately
   // this is a reasonably efficient work-around for leafletjs limitations on svg render order
@@ -131,7 +131,7 @@ window.Render.prototype.processGameEntities = function(entities) {
     var ent = entities[i];
 
     if (ent[2][0] == 'p' && !(ent[0] in this.deletedGuid)) {
-      this.createPortalEntity(ent, 'extended');
+      this.createPortalEntity(ent, details);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


core/code/map_data_render.js
```diff
@@ -131,7 +131,7 @@ window.Render.prototype.processGameEntities = function(entities) {
     var ent = entities[i];
 
     if (ent[2][0] == 'p' && !(ent[0] in this.deletedGuid)) {
-      this.createPortalEntity(ent, 'extended');
+      this.createPortalEntity(ent, details || 'extended');
```
le-jeu: to replace by : ` this.createPortalEntity(ent, details)` for release

le-jeu (approved):

core/code/map_data_render.js
johnd0e: Do you have any 3rd party plugin triggering warning in this case?

le-jeu: my plugin about inventory uses the hook to add core data entities for all portal with keys into the view.
adding `'core'` to the callback prevent the warning (as expected)

johnd0e: So I expect that alternative inventory plugins also use that hook?
Could you show sample of data inserted (just curious).

le-jeu: https://github.com/le-jeu/ingress-intel-total-conversion/blob/ab36c073ddd4ecc44e5da5748c943cf9b1a51a23/plugins/player-inventory.js#L540-L559
```js
const injectKeys = function(data) {
  if (!plugin.isHighlighActive)
    return;

  const bounds = window.map.getBounds();
  const entities = [];
  for (const [guid, key] of plugin.inventory.keys) {
    if (bounds.contains(key.latLng)) {
      // keep known team
      const team = window.portals[guid] ? window.portals[guid].options.ent[2][1] : 'N';
      const ent = [
        guid,
        0,
        ['p', team, Math.round(key.latLng[0]*1e6), Math.round(key.latLng[1]*1e6)]
      ];
      entities.push(ent);
    }
  }
  data.callback(entities);
}
```

modos189 (approved):

PR #447, opened 2021-02-19
Refactor comm tabs code

le-jeu:
This follows an 6 y o idea from 0dcada2 by @jonatkins 
Public/Faction/Alerts tabs are the same up to the data they need and request.

This PR mainly intends to remove duplicate code.

In a beautiful world, this will also lead to an API allowing plugins to add there own comm tab (as [Log Diary](https://github.com/MysticJay/ZasoItems.CE/blob/master/logs-diary.user.js), filtered comm, external chat integration etc)

## Examples 
Chat tab (filtering player messages): https://gist.github.com/le-jeu/58c2d8502d796f5136ca6b1e002801eb
Same, but without the refactoring: https://gist.github.com/le-jeu/38eea4eee723d53d6102cd105567cc56

## Pane integration

<details>
<summary>Selection in pane list</summary>

image

</details>
<details>
<summary>Chat only pane</summary>

image

</details>


johnd0e:
Does this PR introduce any breaking change (for plugins)?
It is still Draft, so what else are you going to do?

le-jeu:
this will break plugins that add a custom comm tab.

modos189:
I've added the code
```
window.chat.renderData = function (data, element, likelyWereOldMsgs, sortedGuids) {
  return comm.renderData(data, element, likelyWereOldMsgs, sortedGuids);
};
```

And it fixed 2 plugins that work with chat (COMM Filter Tab and MUs Tab). Maybe it is possible to provide backwards compatibility with all (or most) plugins?

Also you added an API to add a tab, what if move it to `IITC` namespace?

core/code/comm.js
```diff
@@ -0,0 +1,780 @@
...
+ *
+ * @memberof window.comm
+ * @type {Object}
+ */
+comm._channels = {};
+
+/**
+ * Initialize the channel data.
+ *
+ * @function comm.initChannelData
```
modos189:
```suggestion
 * @function window.comm.initChannelData
```
The same change needs to be made for all functions. Otherwise jsdoc does not see the function

core/code/comm.js
```diff
@@ -0,0 +1,784 @@
...
+ * @memberof comm
+ * @type {Object}
+ */
+comm._channels = {};
+
+/**
+ * Initialize the channel data.
+ *
+ * @function comm.initChannelData
+ * @param {string} The channel object.
```
modos189:
```suggestion
 * @param {string} channel - The channel object.
```

modos189:
If you want to put all comm.js functions in IITC.comm, make these changes in jsdoc
```
diff --git a/core/code/comm.js b/core/code/comm.js
index fe97c8f6..9c8b78f6 100644
--- a/core/code/comm.js
+++ b/core/code/comm.js
@@ -1,5 +1,5 @@
 /**
- * @file Namespace for comm-related functionalities.
+ * Namespace for comm-related functionalities.
  *
  * @memberof IITC
  * @namespace comm
@@ -10,7 +10,7 @@ IITC.comm = comm;
 /**
  * Holds data related to each intel channel.
  *
- * @memberof comm
+ * @memberof IITC.comm
  * @type {Object}
  */
 comm._channels = {};
@@ -18,8 +18,8 @@ comm._channels = {};
 /**
  * Initialize the channel data.
  *
- * @function comm.initChannelData
- * @param {string} The channel object.
+ * @function IITC.comm.initChannelData
+ * @param {string} channel - The channel object.
  */
 comm.initChannelData = function (channel) {
   // preserve channel object
@@ -35,7 +35,7 @@ comm.initChannelData = function (channel) {
 /**
  * Updates the oldest and newest message timestamps and GUIDs in the chat storage.
  *
- * @function comm.updateOldNewHash
+ * @function IITC.comm.updateOldNewHash
  * @param {Object} newData - The new chat data received.
  * @param {Object} storageHash - The chat storage object.
  * @param {boolean} isOlderMsgs - Whether the new data contains older messages.
@@ -198,7 +198,7 @@ comm._oldBBox = null;
 /**
  * Generates post data for chat requests.
  *
- * @function comm.genPostData
+ * @function IITC.comm.genPostData
  * @param {string} channel - The chat channel.
  * @param {boolean} getOlderMsgs - Flag to determine if older messages are being requested.
  * @returns {Object} The generated post data.
  ......... etc.........
```

But surely it's worth making all these features public? I propose to make the IITC API namespace simple and abstract. As it is already done for `IITC.filters` and `IITC.toolbox`.

For example, why a public method `genPostData`? If the plugin needs to request to download messages - there is `requestChannel`. Why a public `handleChannel`? It calls hooks that plugins can use. And so on. 

What do you think about this?

modos189 (approved):
Looks good

plugins/debug-console.js
```diff
@@ -112,80 +118,44 @@ function overwriteNative() {
...
+};
 
 function setup() {
   window.plugin.debug = debugTab;
-  create();
+  debugTab.create();
   overwriteNative();
 
+  $('<style>')
+    .text('#chat #chatdebug td:nth-child(-n+2) { \n  width: 51px\n' + '}\n#chat #chatdebug td:nth-child(3) {\n  font-family: monospace\n}')
```
xscreach: Styles in separate CSS with import/include here?

plugins/debug-console.js
```diff
@@ -71,34 +76,35 @@ debugTab.console.renderLine = function(errorType, args) {
   var d = new Date();
   var ta = d.toLocaleTimeString(); // print line instead maybe?
   var tb = d.toLocaleString();
-  var t = '<time title="'+tb+'" data-timestamp="'+d.getTime()+'">'+ta+'</time>';
-  var s = 'style="color:'+color+'"';
-  var l = '<tr><td>'+t+'</td><td><mark '+s+'>'+errorType+'</mark></td><td>'+text+'</td></tr>';
-  $('#debugconsole table').prepend(l);
+  var t = '<time title="' + tb + '" data-timestamp="' + d.getTime() + '">' + ta + '</time>';
```
xscreach: Could you rather use the string block for all this?

plugins/debug-console.js
```diff
@@ -18,41 +18,46 @@ var debugTab = {};
...
 
-debugTab.console.renderLine = function(errorType, args) {
+debugTab.renderLine = function (errorType, args) {
   args = Array.prototype.slice.call(args);
   var color = '#eee';
   switch (errorType) {
-  case 'error':   color = '#FF424D'; break;
-  case 'warning': color = '#FFDE42'; break;
+    case 'error':
+      color = '#FF424D';
```
xscreach: CSS classes instead of colors? And then colors in css? 🤷

xscreach (changes requested):
Looks good but needs polishing

plugins/debug-console.js
le-jeu: I only intended to adapt debug-console but that's a good occasion to improve it a bit :+1:

plugins/debug-console.js
le-jeu: in fact, it could also be dom element instead of string

plugins/debug-console.js
le-jeu: yes, definitely

plugins/debug-console.js
xscreach: Well, anything will be better than this 🤷😂

xscreach (approved):

xscreach (approved):

PR #454, opened 2021-03-11
chat: fix public/faction not cleared on move

### LINKED ISSUE #417: comm/all shows messages when it should show little or none
Here are the steps to reproduce the issue.  Pick a busy area such as San Francisco.  Go to COMM/All.  See that COMM/All has lots of log entries.  Move the map view way off shore.  Notice that COMM/All has nothing in it but messages just to you if any.  Move the map view back to San Francisco and see that COMM/All has lots of log entries as expected.  Now choose COMM/Faction.  In San Francisco there are lots of log entries.  Move the map view way off shore.  Notice the log entries haven't changed, they are just the same as before.  Now choose COMM/All.  Notice you have very few if any log entries.  Now choose COMM/Faction.  Notice there are few if any log entries.

I expected that if one chooses COMM/Faction in a busy area then the map view changes to someplace where there is no activity then COMM/Faction logging would change accordingly, to little or none.

le-jeu:
fix #417

Force comm tab render when we need to clear the old log.

Note: I don't clear the alerts tab since alerts do not depend of map view

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -179,8 +179,10 @@ window.chat.handleFaction = function(data, olderMsgs, ascendingTimestampOrder) {
     return log.warn('faction chat error. Waiting for next auto-refresh.');
   }
 
-  var needsClearing = $('#chatfaction').data('needsClearing');
-  if(!needsClearing && data.result.length === 0) return;
+  if (!data.result.length && !$('#chatfaction').data('needsClearing')) {
+    // no new data and current data in chat._faction.data is already rendered
+    return;
+  }
 
   $('#chatfaction').data('needsClearing', null);
 
@@ -230,8 +232,10 @@ window.chat.handlePublic = function(data, olderMsgs, ascendingTimestampOrder) {
     return log.warn('public chat error. Waiting for next auto-refresh.');
   }
 
-  var needsClearing = $('#chatall').data('needsClearing');
-  if(!needsClearing && data.result.length === 0) return;
+  if (!data.result.length && !$('#chatall').data('needsClearing')) {
+    // no new data and current data in chat._public.data is already rendered
+    return;
+  }
 
   $('#chatall').data('needsClearing', null);
 
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -179,7 +179,10 @@ window.chat.handleFaction = function(data, olderMsgs, ascendingTimestampOrder) {
     return log.warn('faction chat error. Waiting for next auto-refresh.');
   }
 
-  if(data.result.length === 0) return;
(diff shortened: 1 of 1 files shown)


core/code/chat.js
```diff
@@ -179,7 +179,10 @@ window.chat.handleFaction = function(data, olderMsgs, ascendingTimestampOrder) {
     return log.warn('faction chat error. Waiting for next auto-refresh.');
   }
 
-  if(data.result.length === 0) return;
+  var needsClearing = $('#chatfaction').data('needsClearing');
+  if(!needsClearing && data.result.length === 0) return;
+
+  $('#chatfaction').data('needsClearing', null);
```
johnd0e:
```suggestion
  if (!data.result.length && !$('#chatfaction').data('needsClearing')) {
    return;
  }
  $('#chatfaction').data('needsClearing', null);
```

core/code/chat.js
```diff
@@ -227,7 +230,10 @@ window.chat.handlePublic = function(data, olderMsgs, ascendingTimestampOrder) {
     return log.warn('public chat error. Waiting for next auto-refresh.');
   }
 
-  if(data.result.length === 0) return;
+  var needsClearing = $('#chatall').data('needsClearing');
+  if(!needsClearing && data.result.length === 0) return;
+
+  $('#chatall').data('needsClearing', null);
```
johnd0e:
```suggestion
  if (!data.result.length && !$('#chatall').data('needsClearing')) {
    return;
  }
  $('#chatall').data('needsClearing', null);
```

johnd0e (changes requested):
Nice catch!
I see this in current code:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/2daad393c75f45e9f121136c4717d4dec51fca4f/core/code/chat.js#L73
But it was not used anywhere before your patch.

I have a couple of minor style suggestions, feel free to decide yourself whether to accept suggestions or not.

core/code/chat.js
le-jeu:
```suggestion
  if (!data.result.length && !$('#chatfaction').data('needsClearing')) {
    // no new data and current data in chat._faction.data is already rendered
    return;
  }

  $('#chatfaction').data('needsClearing', null);
```

core/code/chat.js
le-jeu:
```suggestion
  if (!data.result.length && !$('#chatall').data('needsClearing')) {
    // no new data and current data in chat._public.data is already rendered
    return;
  }

  $('#chatall').data('needsClearing', null);
```

johnd0e (approved):

modos189 (approved):

PR #456, opened 2021-03-12
portalslist: add scout/capture/visit filters

le-jeu:
- `All` filter is hidden on desktop, reselect the active filter to disable it 
- includes scout/capture/visit filters
- rework some portalslist functions 
- UI filters: 3 columns on desktop/landscape(mobile) and 2 columns on portrait (mobile) 
- increase description font size 
- fix V/C&S UI size 

image

image

image

~This need to be rebased before merging~

### DIFF
--- a/plugins/portals-list.css
+++ b/plugins/portals-list.css
@@ -192,7 +192,6 @@
 
 #portalslist .disclaimer {
   margin-top: 10px;
-  font-size: 10px;
 }
 
 #portalslist .history .icon {
--- a/plugins/portals-list.css
+++ b/plugins/portals-list.css
@@ -23,6 +23,7 @@
   color: white;
   padding: 3px;
   white-space: nowrap;
+  vertical-align: middle;
 }
 
 #portalslist table th {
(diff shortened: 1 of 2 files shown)


johnd0e:
Rebase?

modos189 (changes requested):
For some reason text on my header and footer are black on my mobile, which makes it hard to distinguish the text.

изображение

Also, it would be nice to add some extra space to the right of the last column, so that the content doesn't end up flush with the border of the table.

johnd0e:
> For some reason text on my header and footer are black on my mobile

Perhaps because of https://github.com/IITC-CE/ingress-intel-total-conversion/pull/497/commits/94e91299be006b810901f70cf8274fdd63d3ccb7

modos189 (approved):

johnd0e:
Several incompatibilities with KitKat (and old iOS) were found.
The major is lack of [grid support](https://caniuse.com/?search=css%20grid).

Imho it's not important to have the same appearance on those legacy systems, but it should be functional at least (some visual defects are acceptable).

le-jeu:
On KitKat (or webview not supporting grid layout)
image

Using 708a0d9
image

PR #460, opened 2021-03-15
fix: dialog failed to collapse sometimes

le-jeu:
sometimes, `sizeFix` is called before data('collapsed',!collapsed), reseting the dialog size after the collapse

Condition likely to reproduce:
 - open a dialog with a huge content list (that doesn't fit the window) (ie portals list in a dense aera)
 - click on the collapse button

Observation;
 - only the button pane is reduced

### DIFF
--- a/core/code/dialog.js
+++ b/core/code/dialog.js
@@ -153,12 +153,21 @@ window.dialog = function(options) {
 
           // Find the button pane and content dialog in this ui-dialog, and add or remove the 'hidden' class.
           var dialog   = $(this).closest('.ui-dialog');
-          var selector = dialog.find('.ui-dialog-content,.ui-dialog-buttonpane');
+          var content = dialog.find('.ui-dialog-content');
+          var buttonpane = dialog.find('.ui-dialog-buttonpane');
           var button   = dialog.find('.ui-dialog-titlebar-button-collapse');
 
           // Slide toggle
           $(this).css('height', '');
-          $(selector).slideToggle({duration: window.DIALOG_SLIDE_DURATION, complete: sizeFix});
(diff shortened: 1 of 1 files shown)


johnd0e:
> Condition likely to reproduce:

Hm. Still unable to reproduce

le-jeu:
https://user-images.githubusercontent.com/64744459/111290947-f3bcc700-8646-11eb-9de5-9292590944b3.mp4

johnd0e:
Ok, I've got it:

> * a huge content list (that doesn't fit the window)

The list should be really huge to trigger the issue. Like >500 or so.

You fix is working for me, but it also introduces unpleasant jitter on expand: 1. I see the expanded window appears, and some space left on top of screen 2. window moves to very top, and button bar becomes visible too.

le-jeu:
> You fix is working for me, but it also introduces unpleasant jitter on expand: 1. I see the expanded window appears, and some space left on top of screen 2. window moves to very top, and button bar becomes visible too.

I can reproduce this issue without my fix.

johnd0e:
> I can reproduce this issue without my fix.

Ok, than it is separate issue.
But may be you could look into that too?

le-jeu:
my hypothesis: 
since `sizeFix` is called twice (one for the content and one for the buttonpane), the delay between both induces a call to `sizeFix` before the completion of the whole slide.
when the dialog has never be moved (from my observation), `sizeFix` re-center the dialog, being slightly lower.
The second call to `sizeFix` re-center again the dialog, provoking the jitter

johnd0e (approved):

modos189 (approved):

PR #462, opened 2021-03-16
hooks: fix case when hook is removing in event handler

johnd0e:
- set removed hook listener to noop so that's not called if remove happens in `runHooks`
- make copy of listeners array to keep `runHook` loop in sync

Close #387 (solves the same issue but in more reliable way).

### DIFF
--- a/core/code/hooks.js
+++ b/core/code/hooks.js
@@ -61,19 +61,19 @@
 // geoSearch:
 // search:
 
-window._hooks = {}
+window._hooks = {};
 window.VALID_HOOKS = []; // stub for compatibility
 
 var isRunning = 0;
 window.runHooks = function(event, data) {
-  if(!_hooks[event]) return true;
+  if (!_hooks[event]) { return true; }
   var interrupted = false;
   isRunning++;
-  $.each(_hooks[event], function(ind, callback) {
+  $.each(_hooks[event], function (ind, callback) {
     try {
       if (callback(data) === false) {
         interrupted = true;
-        return false;  //break from $.each
+        return false; // break from $.each
       }
     } catch (e) {
       log.error('error running hook '+event+', error: '+e);
@@ -81,7 +81,7 @@ window.runHooks = function(event, data) {
   });
   isRunning--;
   return !interrupted;
-}
+};
 
 window.pluginCreateHook = function() {}; // stub for compatibility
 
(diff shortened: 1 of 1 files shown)


core/code/hooks.js
```diff
@@ -64,9 +64,11 @@
 window._hooks = {}
 window.VALID_HOOKS = []; // stub for compatibility
 
+var isRunning = 0;
```
McBen, line 67: If we assume that `removeHook` calls are rare, we can skip this extra `isRunning` check.

This would save a little bit performance at the cost of a 'splice' in every `removeHook`

core/code/hooks.js
```diff
@@ -99,11 +102,16 @@ window.removeHook = function(event, callback) {
...
-    var index = _hooks[event].indexOf(callback);
+  var listeners = _hooks[event];
+  if (listeners) {
+    var index = listeners.indexOf(callback);
     if(index == -1)
       log.warn('Callback wasn\'t registered for this event.');
     else
-      _hooks[event].splice(index, 1);
+      if (isRunning) {
+        listeners[index] = L.Util.falseFn;
```
McBen: `L.Util.falseFn` always return false -> false here means 'interrupt hook'

```suggestion
        listeners[index] = function() {};
```

core/code/hooks.js
```diff
@@ -99,11 +102,16 @@ window.removeHook = function(event, callback) {
...
+        _hooks[event] = listeners = listeners.slice();
```
McBen, line 114: chain assignment - some may like them, some may not.

same code, same performance:
```suggestion
        listeners = listeners.slice();
        _hooks[event] = listeners;
```

core/code/hooks.js
johnd0e: > This would save a little bit performance

Extremely little bit.
I'd rather accept it with another reason: for the sake of simplicity.

But here we have pretty same code as Leaflet, so I prefer to keep it closer to 'upstream'.

core/code/hooks.js
johnd0e: > chain assignment - some may like them, some may not.

I have no own opinion about this (but do not use it myself).
But here is just copy of Leaflet code, and I do not see strong reason to alter it.

core/code/hooks.js
johnd0e: Thank you for this correction!

Solution with `L.Util.falseFn;` had one advantage: it is one function used in multiple cases, in order to decrease garbage collection.

So I will change it to [`listeners[index] = $.noop;`](https://api.jquery.com/jquery.noop/)

modos189 (approved):

PR #464, opened 2021-03-16
API: getCurrentZoomTileParameters -> getDataZoomTileParameters

johnd0e:
Motivation: make API more consistent
- transfer the function from utils_misc.js to map_data_calc_tools.js (where all related routines reside)
- rename it to `getDataZoomTileParameters`, and let it to take map zoom as argument, to make it more general
- put in near  `getMapZoomTileParameters`, to make it clear what is common between them.

Also: deprecate `TILE_PARAMS.ZOOM_TO_LEVEL` and `getMapZoomTileParameters().level`.

### DIFF
--- a/core/code/map_data_calc_tools.js
+++ b/core/code/map_data_calc_tools.js
@@ -98,6 +98,12 @@ window.getMapZoomTileParameters = function(zoom) {
   };
 }
 
+window.getDataZoomTileParameters = function(zoom) {
+  zoom = arguments.length ? zoom : map.getZoom();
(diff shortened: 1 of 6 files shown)


johnd0e:
`getCurrentZoomTileParameters` is [relatively new](https://github.com/IITC-CE/ingress-intel-total-conversion/commit/54916a8cd2d1490fa6c6dd79df0e66799fa0a569) addition to API (not existing in iitc-me), that's why I believe that we can change it without a pain now.

modos189 (approved):

modos189 (approved):

PR #465, opened 2021-03-16
expose default cooldown time and hack count

le-jeu:
replace magic numbers by global constants for cooldown and hack count
```js
window.BASE_HACK_COOLDOWN = 300; // 5 mins - 300 seconds
window.BASE_HACK_COUNT = 4;
```
type `COOLDOWN_TIME = 90` in console to get covidown time (on next  portal details render)

### DIFF
--- a/core/code/portal_info.js
+++ b/core/code/portal_info.js
@@ -292,21 +292,21 @@ window.getPortalHackDetails = function(d) {
   // first mod of type is fully effective, the others are only 50% effective
   var effectivenessReduction = [ 1, 0.5, 0.5, 0.5 ];
 
-  var cooldownTime = COOLDOWN_TIME;
+  var cooldownTime = BASE_HACK_COOLDOWN;
 
   $.each(heatsinks, function(index,mod) {
     var hackSpeed = parseInt(mod.stats.HACK_SPEED)/1000000;
     cooldownTime = Math.round(cooldownTime * (1 - hackSpeed * effectivenessReduction[index]));
   });
 
-  var numHacks = HACK_COUNT; // default hacks
+  var hackCount = BASE_HACK_COUNT; // default hacks
 
   $.each(multihacks, function(index,mod) {
     var extraHacks = parseInt(mod.stats.BURNOUT_INSULATION);
-    numHacks = numHacks + (extraHacks * effectivenessReduction[index]);
+    hackCount = hackCount + (extraHacks * effectivenessReduction[index]);
   });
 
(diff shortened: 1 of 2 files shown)


core/code/portal_info.js
```diff
@@ -292,14 +292,14 @@ window.getPortalHackDetails = function(d) {
...
-  var cooldownTime = 300; // 5 mins - 300 seconds 
+  var cooldownTime = COOLDOWN_TIME;
 
   $.each(heatsinks, function(index,mod) {
     var hackSpeed = parseInt(mod.stats.HACK_SPEED)/1000000;
     cooldownTime = Math.round(cooldownTime * (1 - hackSpeed * effectivenessReduction[index]));
   });
 
-  var numHacks = 4; // default hacks
+  var numHacks = HACK_COUNT; // default hacks
```
johnd0e: Let's keep names in sync.
- Either `var numHacks = NUM_HACKS;`
- Or `var hackCount = HACK_COUNT;`

johnd0e (changes requested):
What else constants we may need?
May be something like `var DOUBLE_AP = false;` or `var AP_FACTOR = 1;`?
What else?

le-jeu:
some times we have double AP on destroyed link, or destroyed field, or both, even triple AP. Also on deploy etc...
The AP gain by action is already exposed tho

MysticJay:
@le-jeu We would need different factors for all actions. we just got double AP for deploying the 8th Reso.

johnd0e:
> We would need different factors for all actions.

Ok then, that is dead end.
I propose to consider custom function for ap gain refinements (not in this PR of cause)

core/code/portal_info.js
MysticJay: 1) AP_FACTOR, not a boolean
2) a COUNT  is something ongoing, better something like MAX_HACKS.

MysticJay:
We should also discuss about a URL, from where these settings can be retrieved.

johnd0e:
> We should also discuss about a URL, from where these settings can be retrieved.

Not in this PR. And even not in core.

McBen:
> What else constants we may need?
> May be something like `var DOUBLE_AP = false;` or `var AP_FACTOR = 1;`?
> What else?

put it into an extra PR

core/code/portal_info.js
McBen: MAX_HACKS sounds good

johnd0e (approved):

modos189 (approved):

PR #466, opened 2021-03-16
Use CSS vars to homogenize UI style

le-jeu:
Thanks to CSS, we can use variables to:
 - avoid repetition of a color with the same semantic in IITC style
 - expose colors for plugins to use instead of copy/paste
 - make style (color) customization simpler (from plugins or in order to make a darktheme)

a purple lover could use
```css
:root {
  --ui-background-color: purple;
  --highlight-text-color: yellow;
}
```

image

while a dark mode could looks like
```css
@media (prefers-color-scheme: dark) {
 :root {
  --ui-background-color: #000e;
  --highlight-text-color: #ccc;
  --ui-border-color: #000;
  --ui-background-color-no-alpha: #000;
  --ui-title-background-color: #0008;
  --highlight-border-color: #fff;
  --arrow-toggle-color: #fff; 
 }
}
```

image

### DIFF
--- a/core/style.css
+++ b/core/style.css
@@ -8,6 +8,9 @@
   --ui-background-color-no-alpha: rgb(8, 48, 78);
   --ui-title-background-color: rgba(8, 60, 78, 0.9);
   --ui-title-button-background-color: #20a8b1;
+  --ui-title-color: #ffce00;
+  --ui-title-inactive-color: #ffffff;
+
   --highlight-text-color: #ffce00;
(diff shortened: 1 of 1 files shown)


johnd0e:
https://caniuse.com/css-variables

Android 4.4 (KitKat) WebView does not support it.
As well as PhanthomJS (not sure that we still care though).

In general: I'd prefer to refactor whole css from scratch first.

PR #470, opened 2021-03-20
Fix Android KitKat support

johnd0e:
- Get rid of Math.trunc in makePermalink function
- Add polyfill (just in case)

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/470)

### DIFF
--- a/core/code/utils_misc.js
+++ b/core/code/utils_misc.js
@@ -330,6 +330,14 @@ if (typeof String.prototype.startsWith !== 'function') {
   };
 }
 
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#polyfill
+// (required for KitKat support)
+if (!Math.trunc) {
+  Math.trunc = function (v) {
+    return v < 0 ? Math.ceil(v) : Math.floor(v);
+  };
+}
+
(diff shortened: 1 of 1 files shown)


core/code/utils_misc.js
```diff
@@ -483,7 +491,7 @@ window.makePermalink = function (latlng, options) {
   options = options || {};
 
   function round (l) { // ensures that lat,lng are with same precision as in stock intel permalinks
-    return Math.trunc(l*1e6)/1e6;
+    return Math.floor(l*1e6)/1e6;
```
modos189, line 494: `Math.trunc` and `Math.floor` have different values on negative numbers. Maybe we should continue to use `Math.trunc`, since a polyfill has been added?

johnd0e: > `Math.trunc` and `Math.floor` have different values on negative numbers.

Right. But in fact `Math.trunc` choice was arbitrary from the beginning. E.g. we could use `Math.round` instead, with similar consequences (0.5:1, -0.5:0).
Anyway I believe that here it does not matter at all, as for permalink there is no need in precise calculations (in any sense).

I can revert to `Math.trunc` if you still feel that more correct.

modos189 (approved):
ok

PR #474, opened 2021-03-23
core: log more data on error

johnd0e:
...in order to help identify error source and conditions

Supersedes #385.

Optimized for Chrome (incl. Android WebView).

Test: https://gist.github.com/johnd0e/b0354b37a894dc4774a87d4cdc0b06e4/raw/ba8d25e616e2222bbd29276dbb4052d1236cb0fd/error.user.js

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/474)

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -699,7 +699,10 @@ function prepPluginsToLoad() {
       setup.call(this);
     } catch (err) {
       var name = info.script && info.script.name || info.pluginId;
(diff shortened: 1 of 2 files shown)


modos189 (approved):
Additionally, I would like to point out that in IITC Mobile it would be better to add possibility to expand the block with detailed information about error, as it is done in browsers and IITC for iOS. Because in case of an error in a large function, it will take up a lot of space.

Issue #458, opened 2021-03-15
Standardize Date/Time format usage


johnd0e:
Atm we have variety of home-brewed formatting functions, like [`unixTimeToString`], `unixTimeToDateTimeString`, `unixTimeToHHmm`, `formatInterval` (and perhaps a numerous local in-place formatting routines in plugins).

It'd be great to use some standardized approach instead, e.g. [`Intl.DateTimeFormat`].

There are still some controversies:
- [ ] Should we always prefer 24h over AM/PM format? Or let it on client to choose (via locale settings of OS).
     - [ ] locale defines some other date/time formatting options, such as month/day/year order and delimiters.
- [ ] Is local time enough, or we should let arbitrary timezone, or just UTC (may be more convenient for global events)?

Affected parts:
- Core
  - [ ] Chats
  - [ ] Region scoreboard
- Plugins
  - [ ] Scoring cycle / checkpoint times 
  - [ ] Player activity tracker
  - [ ] 3rd-party plugins

Note: started in https://github.com/IITC-CE/ingress-intel-total-conversion/commit/e92dd96334bc490a759876ed6b2aed739f8f365a#commitcomment-42973205

[`unixTimeToString`]: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/core/code/utils_misc.js#L186
[`Intl.DateTimeFormat`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

le-jeu:
I don't find any call of `unixTimeToString` except in `score-cycle-times` with `full=true` (it was used to display the capture time long ago). The remaining calls assume a format with seconds at the end(that is broken on 12h cycle)

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/5d593f0221882345aee92f8d54b7eabd6bd20a1b/core/code/utils_misc.js#L186-L197

Is `unixTimeToHHmm` supposed to format time as HH:mm or is it supposed to respect the locale ?
Is getting '9:04 pm' to avoid ?
Are there places we don't want to be locale dependent ? 

Also, region score uses its own format
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/5d593f0221882345aee92f8d54b7eabd6bd20a1b/core/code/region_scoreboard.js#L432-L452

In chat, we use a date time string with milliseconds, the millisecond part is put in a `<small>` tag. 
[...]

PR #475, opened 2021-03-26
Use Date/toLocale* to format date/time

johnd0e:
New `window.locale` variable to store preferred languages (to use in date/time formatting).
Exact time/date format is defined with [`Intl.DateTimeFormat`] options.
Applied to `chat` and `region_scoreboard`.

Same for plugins: #397.
General discussion: #458.

[`Intl.DateTimeFormat`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

---
This change is [Reviewable](https://reviewable.io/reviews/iitc-ce/ingress-intel-total-conversion/475)

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -487,14 +487,28 @@ window.chat.renderDivider = function(text) {
 }
 
 
+window.chat.timeFormat = {
+  timeStyle: 'short'
+};
+
+window.chat.timeDateFormat = {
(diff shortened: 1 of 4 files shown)


johnd0e:
Todo:
- [ ] better time style in chat?
- [ ] common date/time options to be applied in all formats (e.g. change timezone, or preferred 12hour-style)

PR #480, opened 2021-04-10
improve CSS processing

McBen:
use postcss preprocessor for css stuff

instead of a simple 'string' import this is a full css preprocessor support.


McBen:
ATM not working on windows 'cause of EOL stuff

PR #481, opened 2021-04-11
draw-tools: fix "snap to portals" precision

McBen:
If a draw-tools marker is just slightly off (<1e-9) snap won't change the
location. By using a hard comparison instead of the soft "equal" it will
reset the position to the portal position in every case.

example:
`{"lat":51.16280299999999,"lng":4.3944160000000165},`
won't get "snapped" to
`{ "lat": "51.162803", "lng": "4.394416" },`

### DIFF
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -584,7 +584,7 @@ window.plugin.drawTools.snapToPortals = function() {
       if (visibleBounds.contains(ll)) {
         testCount++;
         var newll = findClosestPortalLatLng(ll);
-        if (newll.lat !== ll.lat || newll.lng !== ll.lng) {
+        if (newll.lat !== ll.lat || newll.lng !== ll.lng) { // must be strict
           layer.setLatLng(new L.LatLng(newll.lat, newll.lng));
           changedCount++;
         }
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -584,8 +584,8 @@ window.plugin.drawTools.snapToPortals = function() {
       if (visibleBounds.contains(ll)) {
         testCount++;
         var newll = findClosestPortalLatLng(ll);
-        if (!newll.equals(ll)) {
-          layer.setLatLng(new L.LatLng(newll.lat,newll.lng));
+        if (newll.lat !== ll.lat || newll.lng !== ll.lng) {
+          layer.setLatLng(new L.LatLng(newll.lat, newll.lng));
           changedCount++;
         }
       }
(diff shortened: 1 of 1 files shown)


plugins/draw-tools.js
```diff
@@ -551,72 +550,75 @@ window.plugin.drawTools.snapToPortals = function() {
...
     }
   });
-
-  if (Object.keys(visiblePortals).length == 0) {
+  if (!Object.keys(visiblePortals).length) {
     alert('Error: No portals visible in this view - nothing to snap points to!');
     return;
   }
 
   var findClosestPortalLatLng = function(latlng) {
```
le-jeu: now, the function returns the closest portal only IF the coordinates are not the same
add comment or change name

johnd0e: Ok

McBen:
why hijacking a PR instead of creating your own?
Refactoring the code has nothing to do with "Fix drawtools snap precision"

johnd0e:
> why hijacking a PR instead of creating your own?

The sole reason is to appreciate your work.
<s>Otherwise this PR just will be left behind.</s>

> Refactoring the code has nothing to do with "Fix drawtools snap precision"

I disagree, as fixing is always good opportunity to improve the code.

plugins/draw-tools.js
```diff
@@ -584,8 +584,8 @@ window.plugin.drawTools.snapToPortals = function() {
       if (visibleBounds.contains(ll)) {
         testCount++;
         var newll = findClosestPortalLatLng(ll);
```
modos189, line 586:
```suggestion
        var newll = findClosestPortalLatLng(ll);
        // Using a hard comparison instead of a soft "equal" to fix "snap to portals" precision.
        // See 
https://github.com/IITC-CE/ingress-intel-total-conversion/pull/481
```

plugins/draw-tools.js
```diff
@@ -596,8 +596,8 @@ window.plugin.drawTools.snapToPortals = function() {
         if (visibleBounds.contains(lls[i])) {
           testCount++;
           var newll = findClosestPortalLatLng(lls[i]);
```
modos189, line 598:
```suggestion
          var newll = findClosestPortalLatLng(lls[i]);
          // Using a hard comparison instead of a soft "equal" to fix "snap to portals" precision.
          // See 
https://github.com/IITC-CE/ingress-intel-total-conversion/pull/481
```

modos189 (changes requested):

plugins/draw-tools.js
johnd0e: I suppose that comments would be redundant here.
And in case of need - more info is available in commit message.

modos189: At the very least, we need a link to PR/commit. Otherwise, during refactoring we may accidentally return the past state of code

plugins/draw-tools.js
```diff
@@ -584,8 +584,8 @@ window.plugin.drawTools.snapToPortals = function() {
       if (visibleBounds.contains(ll)) {
         testCount++;
         var newll = findClosestPortalLatLng(ll);
-        if (!newll.equals(ll)) {
-          layer.setLatLng(new L.LatLng(newll.lat,newll.lng));
+        if (newll.lat !== ll.lat || newll.lng !== ll.lng) {
```
johnd0e:
```suggestion
        if (newll.lat !== ll.lat || newll.lng !== ll.lng) { // must be strict
 ```

I suppose such short comment would be enough.

modos189: Ok

modos189 (approved):

PR #485, opened 2021-04-16
boot.js: replace jQuery(callback) with plain js 'load' listener

johnd0e:
Reason: it's hard to debug boot functions wrapped into jQuery Deferred
Ref: https://api.jquery.com/jquery/#jQuery3

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -855,4 +855,8 @@ try {
   throw e;
 }
 
-$(boot);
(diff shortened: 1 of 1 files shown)


johnd0e:
Unfortunately that does not work in mobile app.. Ideas?

Some links:
https://stackoverflow.com/a/9899701
https://stackoverflow.com/a/37332447

modos189 (approved):

PR #486, opened 2021-04-18
Refactor layerChooser, reorganize boot.js, refactor android-specific code

johnd0e:
* rename android.->app., to be more general (as we have iOS app as well)

* collect (most of) android-specific code into single file `app.js`
  - No more numerous `if (typeof android !== 'undefined' && android && android...` conditions
    in random places of code - all relevant stuff is moved to android.js (whenever possible).
    In those places where still needed - new property `window.isApp` is used.
  - Instead of conditions - some functions is directly overwritten (when needed).

* layerChooser: get scattered code together into separate class.
  - android-specific methods included only when needed (android.js)
  - new method toggleLayer, now used also in IITCm (instead of old incomplete java implementation)
    + so longlick in layerchooser now acts consistently both on desktop/mobile
    + IITCm: fix exception when longclick not on layer.
  - new plugins:
    + hide-portal-levels
    + remove-extra-layers

* boot.js: 
  - sidebar stuff is separated in `sidebar.js`
  - map stuff is separated in `map.js`
  - map options are now defined in `window.mapOptions`
    (overridable by plugins with priority `'boot'`)


core/code/layerchooser.js
```diff
@@ -72,21 +36,58 @@ var LayerChooser = L.Control.Layers.extend({
...
+    }, time);
+  };
+}
+
+if (typeof android !== 'undefined' && android && android.setLayers) {
+  LayerChooser.include({
+    // hook some additional code into the LayerControl so it's easy for the mobile app to interface with it
+    _setAndroidLayers: debounce(function () {
+      var l = this.getLayers();
+      android.setLayers((l.baseLayers), JSON.stringify(l.overlayLayers));
```
le-jeu: JSON missing

core/code/layerchooser.js
```diff
@@ -1,5 +1,26 @@
...
+      layer._statusTracking = function (e) {
+        window.updateDisplayedLayerGroup(name, e.type === 'add');
+      };
+      layer.on('add remove', layer._statusTracking, this);
+    }
+  },
+
+  removeLayer: function (layer) {
+    if (layer && layer._statusTracking) {
+      layer.off('add remove', this._statusTracking, this);
```
le-jeu: `layer._statusTracking` ?

plugins/draw-tools.js
```diff
@@ -713,10 +713,10 @@ window.plugin.drawTools.initMPE = function(){
       window.plugin.drawTools.load();
       console.log('DRAWTOOLS: reset all drawn items (func_post)');
 
-      if (window.plugin.crossLinks !== undefined && window.overlayStatus['Cross Links']) {
+      if (window.plugin.crossLinks) {
```
le-jeu, line 715: I assume crosslink and destroyed plugins have to do this on their own, tracking whether their respective layer are displayed

core/code/layerchooser.js
```diff
@@ -33,6 +34,11 @@ var LayerChooser = L.Control.Layers.extend({
         this._storeOverlayState(name, e.type === 'add');
       };
       layer.on('add remove', layer._statusTracking, this);
+    } else {
+      layer._statusTracking = function () {
+        localStorage['iitc-base-map'] = name;
+      };
+      layer.on('add', layer._statusTracking);
```
le-jeu: +`,this)`

plugins/draw-tools.js
johnd0e: They should listen for draw-tools 'load' hook. But that is out of scope of this PR.

core/code/layerchooser.js
johnd0e: Actually not

le-jeu: right

core/code/layerchooser.js
```diff
@@ -6,6 +6,7 @@
  * Provides 'persistence' of layers display state between sessions, saving it to localStorage.
  * Every overlay is added to map automatically if it's last state was active.
  * When no record exists - active is assumed, except when layer has option `defaultDisabled`.
+ * And with `notPersistent` layer option it is possible to suppress it's status tracking.
```
le-jeu: Nice

johnd0e:
For those who are following: rebased, including these fixups - 1f0c3ec43e6f6154ae2aac4d1d8da4d150f5b703, faaf3557b0f4b43249ae2ee844610d2eeb939a08, 6bc85b50f7fa148598428deb8395ca77ef9c28c1.

core/code/layerchooser.js
```diff
@@ -12,16 +12,39 @@
...
     this.lastBaseLayerName = localStorage['iitc-base-map'];
+    this._lastPriority = -1000; // initial layers get priority <0
     L.Control.Layers.prototype.initialize.apply(this, arguments);
+    this._lastPriority = 0; // any following gets >0
   },
 
   _addLayer: function (layer, name, overlay) {
     L.Control.Layers.prototype._addLayer.apply(this, arguments);
+    // provide stable sort order
+    if (!layer.options.sortPriority) {
```
le-jeu: technically, `0` may be a correct value

core/code/layerchooser.js
```diff
@@ -133,6 +133,47 @@ var LayerChooser = L.Control.Layers.extend({
...
+      }
+    };
+
+    // @event longclick: LayersControlEventExt
+    // Fired on layer
+    el.layer.fire('longclick', obj);
+
+    // @event overlaylongclick: LayersControlEventExt
+    // Fired on map
+    this._map.fire('overlaylongclick', L.extend(obj, el));
```
le-jeu: `L.extend(obj, el)` is this common pattern in Leaflet ?

core/code/layerchooser.js
```diff
@@ -133,6 +133,47 @@ var LayerChooser = L.Control.Layers.extend({
...
+    if (!defaultPrevented) {
```
le-jeu, line 330: why not, but I don't see use case since the default is still toggleOverlay
do you expect an override like:
```js
map.on('overlaylongclick', function (e) { 
  e.preventDefault(); 
  doThings();
});

File: core/code/layerchooser.js [hunk 9]
```diff
@@ -253,13 +257,7 @@ window.addLayerGroup = function (name, layerGroup, defaultDisplay) {
   window.layerChooser.addOverlay(layerGroup, name);
 };
 
+// !!deprecated: use `layerChooser.removeLayer` directly (https://leafletjs.com/reference.html#control-layers-removelayer)
```
le-jeu: leaflet doc is confusing since `layerChooser.removeLayer` removes the layer from the map by default

johnd0e: But a line below you can see clear usage sample that clarifies any confusion)

core/code/layerchooser.js
johnd0e: Yes.
Do you have some concerns regarding our case?

core/code/layerchooser.js
johnd0e: I don't know, will decide tomorrow.

johnd0e: I expect something like this:
```js
map.on('overlaylongclick', function (e) {
  if (!e.originalEvent.metaKey) { return; }
  e.preventDefault();
  doThings();
});
// then in different plugin
map.on('overlaylongclick', function (e) {
  if (!e.originalEvent.ctrlKey) { return; }
  e.preventDefault();
  doOtherThings();
});
```
So you are right, explicit `e.preventDefault();` is not convenient, but I have not found better way to detect if default action is still needed after all events processed.

core/code/layerchooser.js
johnd0e: `if (!('sortPriority' in layer.options)) {`

le-jeu: or `!== undefined`
I use this one, but yours is more expressive :+1:

core/code/layerchooser.js
le-jeu: no, it looks weird to extend event data with a layer
I expect `L.extend(obj.target/progabateTarget/sourceTarget, el)` instead
following the https://leafletjs.com/reference-1.7.1.html#event description

johnd0e: `target` is set by leaflet: https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/core/Events.js#L179

> no, it looks weird to extend event data with a layer

`el` is not layer, and we do basically the same as `L.Control.Layers`: https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/control/Control.Layers.js#L303

plugins/collapsible-layers.js
```diff
@@ -56,22 +63,30 @@ function setup () {
...
-        var name = this._name;
-        if (disabled[0]) {
-          name = '<i>' + name + '</i>';
+        if (disabled.length === this._allLayers.length) {
+          disabled.forEach(this.addLayer, this);
+          disabled.length = 0;
+          disabled.all = true;
+        }
+
+        var label;
```
le-jeu:
```suggestion
        var label = this._name;
```

plugins/collapsible-layers.js
```diff
@@ -56,22 +63,30 @@ function setup () {
...
           this.once('longclick', enableAll);
           this.once('expand', function () {
             this.off('longclick', enableAll);
           });
+          label =  this._name + ' [' + disabled.length + ']';
+          label = '<i>' + label + '</i>';
         }
-        this._ctrl.addOverlay(this, name);
+        this.options.defaultDisabled = disabled.all;
+        this._ctrl.addOverlay(this, label ||  this._name);
```
le-jeu:
```suggestion
        this._ctrl.addOverlay(this, label);
```

plugins/collapsible-layers.js
```diff
@@ -14,26 +14,33 @@ collapsibleLayers.initial = true; // collapsed
...
+
       this._name = options.name || match;
+      this._ctrl = window.layerChooser;
       this._layers = {};
       this._allLayers = [];
       if (match) { this.addLayers(match); }
+
+      this.on('add remove', function (e) { // store status
+        this.eachLayer(function (layer) {
+          this._storeOverlayState(layer._name, e.type === 'add'); //??
```
le-jeu: can this be done by adding/removing the layers from the map ?

plugins/collapsible-layers.js
```diff
@@ -14,26 +14,33 @@ collapsibleLayers.initial = true; // collapsed
...
-      this._ctrl = window.layerChooser;
-
-      L.Util.setOptions(this, options);
+      options = options || {};
       options.notPersistent = true;
+      L.Util.setOptions(this, options);
+
       this._name = options.name || match;
+      this._ctrl = window.layerChooser;
       this._layers = {};
```
le-jeu: `this._layers` is never used

johnd0e: It is used in parent class.
There is also `clearLayers` method but it has undesirable (for our case) side effects

plugins/collapsible-layers.js
johnd0e: I wrote exactly like you suggested first, but then changed for the sake of expression,

plugins/collapsible-layers.js
johnd0e: Unfortunately not, as layers are not tracked anymore after they removed from layerChooser.

But I do not like using private methods as well, so suggestions are welcome.

johnd0e: Atm I'm considering another approach: additional 'active' argument to addOverlay/BaseLayer (to enforce desirable state despite of what is recorded in last status), and use that on layers re-adding.

But it will not help to store states while layers are collapsed (e.g.: set 'Portals' to unchecked, then reload - and they are checked again).
Pretty minor case, but I'm still not completely sure.

plugins/collapsible-layers.js
le-jeu: then, calling `LayerGroup.initialize.call(this.[],options)` sounds more idiomatic

plugins/collapsible-layers.js
le-jeu: in this case, I would separate calls with the `if` above (style anyway)

plugins/collapsible-layers.js
le-jeu: what about not storing anything ? 
the point of the layer is to enable/disable all collapsed layer in one click. 
It's fair not to be persistent changes

johnd0e: Consider such case:
- collapse all-enabled levels into single entry
- disable that entry
- expand

What state do you expect?
Without storing state - all be (unexpectedly) enabled.

plugins/collapsible-layers.js
johnd0e: Still heavily dependent on parent implementation, so in the end there is no difference.

plugins/collapsible-layers.js
le-jeu: this can be handled line 93, calling `showLayer(layer, false)` for disabled layer

johnd0e: That is far from optimal: thousands of layers will be added and then immediately removed.

johnd0e: Experimental: https://github.com/IITC-CE/ingress-intel-total-conversion/pull/486/commits/a32772057504212b743b6721734e773c43460e7e

core/code/layerchooser.js
```diff
@@ -238,29 +275,25 @@ var LayerChooser = L.Control.Layers.extend({
...
+    var data = this.layerInfo(layer);
+    if (!data) {
       log.warn('Layer not found');
       return this;
     }
-    var info = this._layers[idx];
-    label = label || info.layer._chooser.name;
-    info.name = label;
-    var nameEl = this._layerControlInputs[idx].closest('label').querySelector('span');
+    label = label;
```
le-jeu: `label = label || data.name` ?

core/code/layerchooser.js
```diff
@@ -371,11 +404,11 @@ var LayerChooser = L.Control.Layers.extend({
...
     var baseLayers = [];
     var overlayLayers = [];
-    this._layers.forEach(function (info, idx) {
-      (info.overlay ? overlayLayers : baseLayers).push({
+    this._layers.forEach(function (data, idx) {
+      (data.overlay ? overlayLayers : baseLayers).push({
         layerId: idx,
-        name: this._stripHtmlTags(info.name), // IITCm does not support html in layers labels
-        active: this._map.hasLayer(info.layer)
+        name: this._stripHtmlTags(data.label), // IITCm does not support html in layers labels
```
le-jeu: + `|| data.name`

core/code/layerchooser.js
```diff
@@ -99,6 +122,17 @@ var LayerChooser = L.Control.Layers.extend({
...
   },
 
+  _addItem: function (obj) {
+    var labelEl = L.Control.Layers.prototype._addItem.call(this, {
+      layer: obj.layer,
+      overlay: obj.overlay,
+      name: obj.label || obj.name
+    });
+    obj.labelEl = labelEl;
+    obj.inputEl = this._layerControlInputs[this._layerControlInputs.length-1];
```
le-jeu: `inputEL` is not used

core/code/layerchooser.js
```diff
@@ -238,29 +275,25 @@ var LayerChooser = L.Control.Layers.extend({
...
       log.warn('Layer not found');
       return this;
     }
-    var info = this._layers[idx];
-    label = label || info.layer._chooser.name;
-    info.name = label;
-    var nameEl = this._layerControlInputs[idx].closest('label').querySelector('span');
+    label = label;
+    data.label = label;
+    var nameEl = data.labelEl.closest('label').querySelector('span');
```
le-jeu: to my understanding, `L.LayerChooser._addItem` returns the `<label>` already

core/code/layerchooser.js
johnd0e: But can be used in plugins

core/code/layerchooser.js
johnd0e: Right

johnd0e:
New batch of commits
@le-jeu

core/code/map_data_request.js
```diff
@@ -60,7 +60,7 @@ window.MapDataRequest = function() {
   this.RENDER_BATCH_SIZE = window.map.options.preferCanvas ? 1E9 : 1500;
 
   // delay before repeating the render loop. this gives a better chance for user interaction
-  this.RENDER_PAUSE = (typeof android === 'undefined') ? 0.1 : 0.2; //100ms desktop, 200ms mobile
+  this.RENDER_PAUSE = window.isAndroid ? 0.1 : 0.2; // 100ms desktop, 200ms mobile
```
le-jeu:
```suggestion
  this.RENDER_PAUSE = !window.isAndroid ? 0.1 : 0.2; // 100ms desktop, 200ms mobile
```

core/code/map.js
```diff
@@ -206,10 +206,6 @@ window.setupMap = function() {
 
   window.layerChooser = new window.LayerChooser(baseLayers, overlays, {map: map})
     .addTo(map);
-  // hide layer chooser if booted with the iitcm android app
-  if (window.isAndroid && android.setLayers) {
-    $('.leaflet-control-layers').hide();
-  }
```
le-jeu: where is this moved ?

core/code/utils_misc.js
```diff
@@ -106,6 +106,31 @@ window.aboutIITC = function () {
...
+  $('<a>')
+    .html('About IITC')
+    .attr('id', 'about-iitc')
+    .css('cursor', 'help')
+    .click(aboutIITC)
+    .appendTo('#toolbox');
+
+  window.artifact.setup();
+
+  window.RegionScoreboardSetup();
```
le-jeu:
```suggestion
  window.RegionScoreboard.setup();
```

core/code/utils_misc.js
```diff
@@ -106,6 +106,31 @@ window.aboutIITC = function () {
...
+window.setupAddons = function () {
+  $('<a>')
+    .html('Permalink')
+    .attr({
+      id: 'permalink',
+      title: 'URL link to this map view'
+    })
+    .on({
+      mouseover: setPermaLink,
+      click: setPermaLink
```
le-jeu: `setPermalink` is removed

core/code/android.js
```diff
@@ -112,6 +112,18 @@ window.runOnAndroidBeforeBoot = function () {
...
+    window.renderPortalUrl = function (lat, lng, title, guid) {
+      // one share link option - and the android app provides an interface to share the URL,
+      // share as a geo: intent (navigation via google maps), etc
+
+      var shareLink = $('<a>').text('Share portal').click(function () {
+        android.intentPosLink(lat, lng, window.map.getZoom(), title, true);
+      });
+      $('linkdetails').append($('<aside>').append(shareLink));
+    };
+  }
```
le-jeu: this pattern (creating a HTMLElement, and fill if afterward) looks wrong to me in this context.
What about hooking on `renderDetailUpdated` instead ? Or make the function return a `<div>` directly ?

core/code/portal_detail_display.js
```diff
@@ -9,6 +9,24 @@ window.resetScrollOnNewPortal = function() {
   }
 };
 
+// to be ovewritten in android.js
+window.renderPortalUrl = function (lat, lng, title, guid) {
+  var linkDetails = $('linkdetails');
```
le-jeu: isn't it `'.linkdetails'` ?

core/code/map.js
johnd0e: to `android.js`

core/code/android.js
johnd0e: > What about hooking on renderDetailUpdated instead?

Thwt would be perfect, but there is a condition: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/4d46ab7178bc2c07252c05758a393ac34b037b2d/core/code/portal_detail_display.js#L155-L157

> Or make the function return a `<div>` directly ?

Well, there are 2 elements.. Put them under one div?

johnd0e:
@le-jeu 
Final version

johnd0e:
May be better change names
`android.js -> app.js`
`isAndroid() -> isApp`

core/code/android.js
le-jeu: > Thwt would be perfect, but there is a condition: 

yes, the hook doesn't match so that won't work

> Well, there are 2 elements.. Put them under one div? 

there is only one `<div class="linkDetails"/>` so that matches

note: Raw portal also inject it's "Raw data" button in this div

(anyway, that's only a pattern I don't commonly use. I prefer function that returns element smth that function that display smth when the choice)

core/code/sidebar.js
```diff
@@ -0,0 +1,264 @@
...
+    .join('\n');
+
+  var html = ''
+  + '<div><b>About IITC</b></div> '
+  + '<div>Ingress Intel Total Conversion</div> '
+  + '<hr>'
+  + '<div>'
+  + '  <a href="'+'@url_homepage@'+'" target="_blank">IITC Homepage</a> |' +
+    '  <a href="'+'@url_tg@'+'" target="_blank">Telegram channel</a><br />'
+  + '   On the script�s homepage you can:'
```
le-jeu: the `’` is not unicode here

le-jeu:
> May be better change names `android.js -> app.js` `isAndroid() -> isApp`

assuming the iOS app implement parts of the android object, it sounds good wrt current PR state

core/code/sidebar.js
```diff
@@ -0,0 +1,264 @@
...
+    .appendTo('#toolbox');
+
+  $('<a>')
+    .html('About IITC')
+    .attr('id', 'about-iitc')
+    .css('cursor', 'help')
+    .click(aboutIITC)
+    .appendTo('#toolbox');
+
+  window.artifact.setup();
```
le-jeu, line 262: while this add a button in the toolbox, I'd rather put it back into the main boot sequence, as it's a whole intel feature

same could be said about regionscoreboard
that makes me wonder if a some `~IITC.sidebar.addToolboxLink` (to be renamed) would be a better API

core/code/sidebar.js
```diff
@@ -0,0 +1,264 @@
+window.setupSidebar = function() {
```
modos189, line 1: Something is wrong with file encoding

core/code/sidebar.js
```diff
@@ -0,0 +1,264 @@
...
+      '#largepreview.res img { border:2px solid '+COLORS[TEAM_RES]+'; } ',
+      '#largepreview.none img { border:2px solid '+COLORS[TEAM_NONE]+'; } ',
+      '#chatcontrols { bottom: '+(CHAT_SHRINKED+22)+'px; }',
+      '#chat { height: '+CHAT_SHRINKED+'px; } ',
+      '.leaflet-right { margin-right: '+(SIDEBAR_WIDTH+1)+'px } ',
+      '#updatestatus { width:'+(SIDEBAR_WIDTH+2)+'px;  } ',
+      '#sidebar { width:'+(SIDEBAR_WIDTH + HIDDEN_SCROLLBAR_ASSUMED_WIDTH + 1 /*border*/)+'px;  } ',
+      '#sidebartoggle { right:'+(SIDEBAR_WIDTH+1)+'px;  } ',
+      '#scrollwrapper  { width:'+(SIDEBAR_WIDTH + 2*HIDDEN_SCROLLBAR_ASSUMED_WIDTH)+'px; right:-'+(2*HIDDEN_SCROLLBAR_ASSUMED_WIDTH-2)+'px } ',
+      '#sidebar > * { width:'+(SIDEBAR_WIDTH+1)+'px;  }'].join('\n')
```
modos189, line 24: These styles have a higher priority than the styles in smartphone.css.

For example, on mobile the portal information window is not stretched across the entire width

core/code/android.js
```diff
@@ -0,0 +1,168 @@
...
+    }, time);
+  };
+}
+
+function extendLayerChooser () {
+  if (android.setLayers) {
+    // hook some additional code into the LayerControl so it's easy for the mobile app to interface with it
+    window.LayerChooser.include({
+      _setAndroidLayers: debounce(function () {
+        var l = this.getLayers();
```
modos189: I see that it works, but how? Is there a Leaflet method getLayers() that returns the data in the form we want?

modos189:
> May be better change names `android.js -> app.js` `isAndroid() -> isApp`

Then it would be a good to make alias `window.android` → `window.app`

core/code/android.js
johnd0e: We have own `.getLayers()`, see `layerchooser.js`

core/code/sidebar.js
johnd0e: @le-jeu 

> assuming the iOS app implement parts of the android object, it sounds good wrt current PR state

So you do support renaming?

johnd0e:
@le-jeu

> assuming the iOS app implement parts of the android object, it sounds good wrt current PR state

So you do support renaming?

core/code/sidebar.js
johnd0e: It's all about intel features, imho. Compare with `missions`.
So.. I'm not sure.

For me it belongs to sidebar because we cannot see it otherwise - only by clicking sidebar links.

johnd0e:
@le-jeu 
There are also 3 demo plugins (in Layers category).
Should we keep them? And which of them?

le-jeu:
> @le-jeu There are also 3 demo plugins (in Layers category). Should we keep them? And which of them?

I forgot about the demo plugins
They could end into the Wiki. The huge one is somewhat a combination of the small ones.
Maybe keep only `hide-portal-level` or `collapsible-layer`

johnd0e:
@McBen @MysticJay

modos189 (approved):

PR #494, opened 2021-04-29
Multi-Line-Headers

MysticJay:
build_plugin.py

Purpose: concatenate lines to avoid long headers, e.g. for Description.

Usecase: Long lines in the sources trigger eslint and reduce readability. 
As *monkey has no trick to handle multi-line headers we need to re-join these during build-process

How to use: This change will concatenate the next (comment) line to the previous header. Indentation is optional.
```
// @keyword line 1
//      line 2
//      line 3
// @keyword2 text
```
becomes
```
// @keyword line 1 line 2 line 3
// @keyword2 text
```
during build-process.

### DIFF
--- a/build_plugin.py
+++ b/build_plugin.py
@@ -24,6 +24,9 @@ def fill_meta(source, plugin_name, dist_path):
     meta = ['// ==UserScript==']
     keys = set()
 
+    # concatenate subsequent simple comment lines to the previous line (multi-line-headers)
+    source = re.sub(r'\n\/\/\s+([^@].+)$', r' \1', source, flags=re.MULTILINE)
+    
     def append_line(key, value):
         if key not in keys:
             meta.append(f'// @{key:<14} {value}')


modos189:
Maybe it only makes sense to use this for the description tag? The value of the other tags is always short enough.

McBen:
> Maybe it only makes sense to use this for the description tag? The value of the other tags is always short enough.

Sure. But does it do any harm? And why creating a special case for only one tag?

johnd0e:
I feel #510 slightly more in line with current code style

le-jeu:
Third suggestion, repeat the keyword to append

    // @description Foo, contact
    // @description @bar on Twitter

johnd0e:
@le-jeu Why? Is that supported by any userscript manager?

MysticJay:
@le-jeu that was my initial, but rejected proposal

MysticJay:
@johnd0e inly before the build

modos189 (changes requested):
Good idea, but let's use PR #510 with an alternative implementation that gets rid of regex and looks better in our build system

PR #496, opened 2021-05-03
Add Niantic like Cookie dialog

McBen:
a niantic like cookie policy dialog.

will be shown modal on startup
grafik

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -739,6 +739,7 @@ function boot() {
   var loadPlugins = prepPluginsToLoad();
   loadPlugins('boot');
 
+  window.checkCookieLaw();
   window.extendLeaflet();
   window.extractFromStock();
   window.setupIdle();
--- a/core/code/cookieLaw.js
+++ b/core/code/cookieLaw.js
@@ -1,0 +1,35 @@
+window.checkCookieLaw = function () {
+
+    const acceptCookies = window.readCookie('_ncc');
+    if (typeof (acceptCookies) !== "undefined") return;
+
+    var NCC_EXPIRE = 60;
+    const html = '<p>We use cookies to give you the best possible user experience.</p>' +
+        'Read our cookies policy <a href="https://nianticlabs.com/cookies/">here</a> ' +
+        'to learn more about our use of cookies, your choices, and how to change your browser settings.'
+
+    dialog({
+        title: 'Niantic Cookies Policy',
+        html,
+        modal: true,
+        buttons: [
+            {
+                text: "Decline",
+                id: "cookiedecline",
+                click: function () {
+                    writeCookie("_ncc", 0, NCC_EXPIRE);
+                    $(this).dialog("close");
+                },
+            },
+            {
+                text: "Accept",
+                click: function () {
+                    writeCookie("_ncc", 1, NCC_EXPIRE);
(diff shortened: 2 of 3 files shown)


johnd0e:
I would rather not get involved into this legal subtleties **at all**.
What we can do easily: let user to see stock intel's "cookies consent" dialog, and make some choice.
(E.g. delay loading for 5 s if no `_ncc` cookie found).

But another question is whether IITC itself is GDPR-compliant: https://www.iubenda.com/en/help/7515-which-data-can-be-safely-stored-in-cookies-without-having-to-request-consent
According to that info - it is.
But we are never sure in all that law-stuff...

McBen:
A quick list what personal data is stored by Intel:
- Access-Token
- Session-ID

additional by IITC:
- last viewed area 
- Uniques
- Draw-stuff
- Bookmarks
- Missions
- Keys
....and these all can get uploaded to google too

which one are **Technical cookies** and which one are **Statistical cookies** ?

le-jeu:
last viewed area is also stored by intel
all other data are app data... and are not linked in anyway to niantic
there are neither shared with third party (including nia) unless the user explicitly says it (sync plugin).

johnd0e:
> which one are **Statistical cookies** ?

No one.

> * ....and these all can get uploaded to google too

Not by default, but after explicit user action, which is hard to interpret other than consent.

McBen:
> last viewed area is also stored by intel

position is stored for current browser-session and covered by Nia cookie policy - which you don't see if you're using IITC. I usually would argue it's okay to store it for longer if you accept cookies. But vanilla intel doesn't do it and IITC doesn't allow choosing it. IITC always stores it for 10years.

> all other data are app data... and are not linked in anyway to niantic

it's about personal data. IITC is not niantic.

> unless the user explicitly says it (sync plugin).

so it's clear to everybody which data the sync shares with google?

johnd0e:
> so it's clear to everybody which data the sync shares with google?

You decide.
Add some text for clarification, where it is needed.

johnd0e:
But look. Who ever 'shares' data with Google?
I see that as iitc stores user's data on user's google drive, with user's content.

And in this case(imho) that data is really useless for Google for any tracking/advertisement/statistics/(whatever other) unclear purposes.

McBen:
> But look. Who ever 'shares' data with Google?
> I see that as iitc stores user's data on user's google drive, with user's content.
> 
> And in this case(imho) that data is really useless for Google for any tracking/advertisement/statistics/(whatever other) unclear purposes.

first off: it doesn't matter where personal data ist stored  (Art. 2 Abs. 1 DSGVO).
It doesn't matter if YOU think the data is useless for other people (for everyone who can access you google-drive data. like admins, google team, hackers,....).

The question is what data has to be protected / the user must agree that IITC is allowed to work with it.

johnd0e:
> The question is what data has to be protected / the user must agree that IITC is allowed to work with it

Exactly.
But I've already posted link that proves my opinion.

If you really want to be over-protected, than you should
- include IITC's own consent (as we keep own data)
- (like seen on some websites) enumerate all kinds of that data (localstorage keys), describing their purpose.

McBen:
okay, I see this is a useless discussion. I'll save us all some time and nerves.

PR #497, opened 2021-05-05
Do not hide Niantic cookies consent popup

johnd0e:
Close #496

Proof-of-concept.

### DIFF
--- a/core/code/utils_misc.js
+++ b/core/code/utils_misc.js
@@ -145,7 +145,7 @@ window.readCookie = function(name){
 }
 
 window.writeCookie = function(name, val) {
-  var d = new Date(Date.now() + 10 * 365 * 24 * 60 * 60 * 1000).toUTCString();
+  var d = new Date(Date.now() + 14 * 24 * 60 * 60 * 1000).toUTCString();
   document.cookie = name + "=" + val + '; expires='+d+'; path=/';
 }
 
--- a/core/smartphone.css
+++ b/core/smartphone.css
@@ -1,5 +1,5 @@
-body {
-  color: #fff;
+#map {
+  color: white;
 }
 
 #updatestatus {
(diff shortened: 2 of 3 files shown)


johnd0e:
We should also consider alternative solution (in separate PR): does not load IITC at all until we have `_ncc` cookie (or 10 s timeout).

Caveat: to be safe in case they change cookie name (or the whole scheme) we should actually check if  `<ark-cookiebar>` element exists (and not empty).

johnd0e:
@johnd0e That is my current proposition. Will be in next test build

modos189:
This PR is currently breaking IITC on android kitkat. On my desktop, the IITC wouldn't start until I pressed "accept" and reloaded the page.

Anyway,
1. Although IITC uses cookies, it doesn't collect personal data (not any data is personal).
2. It requires a created account in Ingress to run, which on one hand shifts the responsibility to Niantic, but on the other hand IITC does not allow to place in some app stores (Microsoft Edge addons or Huawei AppGalery)
3. GDRP requires, for example, the ability to export user data. We don't have that capability, this is Niantic's domain.
4. I, like many, hate banners offering to accept cookies.

Therefore, I strongly oppose the addition of this banner.

crystalwizard:
The law requires users be given a chance to accept or decline cookies if they are set. Since iitc sets cookies, regardless of what information is being saved, users have to be asked to accept or decline and be allowed to actually click the correct button. If declined, cookies can NOT be set - not allowing the program to start until accept is clicked is what is supposed to happen.

modos189:
What is IITC?
If it's a mobile app - nobody cares about implementation details, the IITC doesn't write anything to user's browser.
If a set of .user.js scripts... Let's wait until the https://addons.mozilla.org/en/firefox/addon/i-dont-care-about-cookies/ extension is blocked first and then come back to this question

johnd0e:
> On my desktop, the IITC wouldn't start until I pressed "accept" and reloaded the page.

Unable to reproduce, Chrome/Firefox, Tampermonkey/IITCButton

PR #499, opened 2021-05-10
IITCm: smarter Google login fix for KitKat and below

johnd0e:
No more 'force desktop' requirement


modos189 (approved):
On kitkat logging in via Google after entering login and pressing "continue" returns to the main page with Google/Facebook selection.
The second login attempt is successful.
Bad, but this is better than opening IITC desktop on mobile, as it's implemented now.

Issue #220, opened 2019-07-10
Simplify mobile panes API


johnd0e:
This API came from here: https://github.com/iitc-project/ingress-intel-total-conversion/pull/675

Here are typical changes needed for plugin to use it: https://github.com/iitc-project/ingress-intel-total-conversion/pull/697/files or 69aad81933fe0657fdf39d36103adef76a37d837.

That involves `android.addPane` + `addHook("paneChanged",...` + handling that hook + additional css.

Imho it is very excessive and inconvenient.

johnd0e:
I'm going to replace this with more simple and straightforward API, involving single call.
Like this:
```
addMobilePane(id,title,icon,onFocus)
```
where `onFocus` is function that should return html element to show.

That's all. No need to handle anything. No need in special css (in most cases).

PR #503, opened 2021-05-11
IITC-Mobile: Mobile pane API

le-jeu:
following #220 idea

> I'm going to replace this with more simple and straightforward API, involving single call.
> Like this:
> 
> ```
> addMobilePane(id,title,icon,onFocus)
> ```
> 
> where `onFocus` is function that should return html element to show.
> 
> That's all. No need to handle anything. No need in special css (in most cases).

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -666,12 +666,13 @@ window.chat.chooseTab = function(tab) {
 }
 
 window.chat.show = function(name) {
-    window.isSmartphone()
-        ? $('#updatestatus').hide()
-        : $('#updatestatus').show();
-    $('#chat, #chatinput').show();
+  window.isSmartphone()
+      ? $('#updatestatus').hide()
+      : $('#updatestatus').show();
+  $('#chat, #chatinput').show();
 
-    window.chat.chooseTab(name);
+  window.chat.chooseTab(name);
+  return $("<div>").append($('#chat, #chatinput'));
 }
 
 window.chat.chooser = function(event) {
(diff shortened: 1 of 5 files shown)


johnd0e:
Currently 'icon' is just mobile resource identifier, which could be enough for standard plugins, but very limiting for any 3rd-party ones.

I wonder if it possible to use icons not embedded into app.

modos189:
> Currently 'icon' is just mobile resource identifier, which could be enough for standard plugins, but very limiting for any 3rd-party ones.
> 
> I wonder if it possible to use icons not embedded into app.

I don't know how hard it is for us to implement, but yes, we can pass the icon from the plugin as a blob. However, there will inevitably be problems with this in the future:
- Such icons will not be optimized for different screen sizes
- plugin developers are rarely designers, and therefore will tend to use icons that are out of style
- such icons will not support a dark/light theme.

For this reason I propose to use only Material icons.

johnd0e:
> * Such icons will not be optimized for different screen sizes

SVG

> * plugin developers are rarely designers, and therefore will tend to use icons that are out of style

Then we will blame them


> For this reason I propose to use only Material icons.

But embedded set of icons is extremely limited.

modos189:
I think Material icons are enough, but ok, we can consider using svg icons

johnd0e:
> I think Material icons are enough

I could agree with you if it would possible to request any icon just by it's id, and then IITCm download (and cache) it.

MysticJay:
there is also a proposal to use FA icons.
they can be addressed/selected by name
have a common style and are available in different formats including SVG

johnd0e:
conflicts!

PR #505, opened 2021-05-16
TWEAKS refactoring

MysticJay:
TWEAKS plugins refactoring
namespace usage
code style fixes according to Eslint rules
Minor code changes

### DIFF
--- a/plugins/fix-china-map-offset.js
+++ b/plugins/fix-china-map-offset.js
@@ -65,7 +65,8 @@ window.plugin.fixChinaMapOffset = fixChinaMapOffset;
 // https://github.com/Artoria2e5/PRCoords
 // (more info: https://github.com/iitc-project/ingress-intel-total-conversion/pull/1188)
 
-var insane_is_in_china = (function () { // adapted from https://github.com/Artoria2e5/PRCoords/blob/master/js/misc/insane_is_in_china.js
+// adapted from https://github.com/Artoria2e5/PRCoords/blob/master/js/misc/insane_is_in_china.js
+var insane_is_in_china = (function () {
 /* eslint-disable */
   'use strict';
 
@@ -164,12 +165,13 @@ var insane_is_in_china = (function () { // adapted from https://github.com/Artor
   }
 
   return { isInChina: isInChina };
-/* eslint-enable */
+  /* eslint-enable */
 })();
 
 fixChinaMapOffset.isInChina = insane_is_in_china.isInChina;
 
-var PRCoords = (function () { // adapted from https://github.com/Artoria2e5/PRCoords/blob/master/js/PRCoords.js
+// adapted from https://github.com/Artoria2e5/PRCoords/blob/master/js/PRCoords.js
+var PRCoords = (function () {
 /* eslint-disable */
   'use strict';
 
@@ -212,7 +214,7 @@ var PRCoords = (function () { // adapted from https://github.com/Artoria2e5/PRCo
   }
 
   return { wgs_gcj: wgs_gcj };
-/* eslint-enable */
+  /* eslint-enable */
 })();
 
 fixChinaMapOffset.wgs_gcj = PRCoords.wgs_gcj;
@@ -221,12 +223,14 @@ fixChinaMapOffset.wgs_gcj = PRCoords.wgs_gcj;
 var fixChinaOffset = {
   _inChina: false,
 
-  _inChinaLastChecked: [0,0],
+  _inChinaLastChecked: [0, 0],
 
   _inChinaValidRadius: 100000,
 
   _isInChina: function (latlng) {
-    if (latlng._notChina) { return false; } // do not check twice same latlng
+    if (latlng._notChina) {
+      return false;
+    } // do not check twice same latlng
 
     if (latlng.distanceTo(this._inChinaLastChecked) > this._inChinaValidRadius) {
       // recheck only when beyond of specified radius, otherwise keep last known value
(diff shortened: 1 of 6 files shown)


plugins/link-show-direction.js
```diff
@@ -1,17 +1,24 @@
...
+/* exported setup --eslint */
+/* global L, dialog, addHook, map, links */
 // use own namespace for plugin
-window.plugin.linkShowDirection = function() {};
-window.plugin.linkShowDirection.ANIMATE_UPDATE_TIME = 1000; // 1000ms = 1s
+var linkShowDirection = {};
+window.plugin.linkShowDirection = linkShowDirection;
+
+// exposed
+// linkShowDirection.showDialog = showDialog;
```
modos189: I think it can be deleted

modos189 (approved):

plugins/fix-china-map-offset.js
```diff
@@ -1,10 +1,12 @@
 ﻿// @author         modos189
 // @name           Fix maps offsets in China
 // @category       Tweaks
-// @version        0.3.0
+// @version        0.3.1
 // @description    Show correct maps for China user by applying offset tweaks.
 
 
+/* exported setup --eslint */
+/* global L */
```
johnd0e, line 8: This should be defined globally.

plugins/link-show-direction.js
```diff
@@ -1,17 +1,24 @@
 // @author         jonatkins
 // @name           Direction of links on map
 // @category       Tweaks
-// @version        0.2.1
+// @version        0.2.2
 // @description    Show the direction of links on the map by adding short dashes to the line at the origin portal.
 
 
+/* exported setup --eslint */
+/* global L, dialog, addHook, map, links */
```
johnd0e, line 8: There is no sense to define them in EVERY file.
Either global definition (for some of entities), or prefix with `window`.

plugins/link-show-direction.js
```diff
@@ -1,17 +1,24 @@
...
+
+// exposed
+// linkShowDirection.showDialog = showDialog;
+
+var ANIMATE_UPDATE_TIME = 1000; // 1000ms = 1s
 
 // Hack:
 // 100000 - a large enough number to be the equivalent of 100%, which is not supported Leaflet when displaying with canvas
-window.plugin.linkShowDirection.styles = {
+var styles = {
```
johnd0e: Someone may want to override the styles, so better expose them.

plugins/scroll-wheel-zoom-disable.js
```diff
@@ -1,17 +1,13 @@
 // @author         jonatkins
 // @name           Disable mouse wheel zoom
 // @category       Tweaks
-// @version        0.1.0
+// @version        0.1.1
 // @description    Disable the use of mouse wheel to zoom. The map zoom controls or keyboard are still available.
 
-
+/* exported setup --eslint */
```
johnd0e, line 7: Let's move it to the end of file.

johnd0e (changes requested):

plugins/scroll-wheel-zoom-disable.js
MysticJay: IMHO lint exclusions should be stated at the start of the code as with all the other plugins.

plugins/fix-china-map-offset.js
MysticJay: This changes from file to file, so having it in the same place without the need to look up at different places is more convenient.

plugins/link-show-direction.js
MysticJay: it was decided NOT to avoid window. wher possible.
as the exclusions/globals may differ from file to file a global definition is inconvenient. Same with splitting these

plugins/link-show-direction.js
MysticJay: exposed

plugins/link-show-direction.js
MysticJay: removed

plugins/link-show-direction.js
```diff
@@ -1,17 +1,23 @@
...
+window.plugin.linkShowDirection = linkShowDirection;
+var styles = {};
+linkShowDirection.styles = styles;
+
+var ANIMATE_UPDATE_TIME = 1000; // 1000ms = 1s
 
 // Hack:
 // 100000 - a large enough number to be the equivalent of 100%, which is not supported Leaflet when displaying with canvas
-window.plugin.linkShowDirection.styles = {
+styles = {
```
le-jeu: the last assignation creates a different object that the one stored in `linkShowDirection.styles`
so it's not exposed anymore

plugins/link-show-direction.js
```diff
@@ -32,84 +38,87 @@ window.plugin.linkShowDirection.styles = {
...
   // this would mean the user has no chance to interact with IITC
   // to prevent this, create a short timer that then sets the timer for the next frame. if the browser is slow to render,
   // the short timer should fire later, at which point the desired ANIMATE_UPDATE_TIME timer is started
-  clearTimeout(window.plugin.linkShowDirection.timer);
-  window.plugin.linkShowDirection.timer = setTimeout(function() {
-    clearTimeout(window.plugin.linkShowDirection.timer);
-    window.plugin.linkShowDirection.timer = setTimeout(
-      window.plugin.linkShowDirection.animateLinks,
-      window.plugin.linkShowDirection.ANIMATE_UPDATE_TIME);
+  clearTimeout(timer);
```
le-jeu, line 68: `timer` is not defined yet here

plugins/link-show-direction.js
MysticJay: pluginName added

plugins/link-show-direction.js
MysticJay: pluginName added

plugins/link-show-direction.js
```diff
@@ -148,23 +168,27 @@ window.plugin.linkShowDirection.showDialog = function() {
   });
 };
 
-window.plugin.linkShowDirection.setup  = function() {
-  $('#toolbox').append(' <a onclick="window.plugin.linkShowDirection.showDialog()">LinkDirection Opt</a>');
+function setup() {
+  $('#toolbox').append('<a onclick="window.plugin.linkShowDirection.showDialog();return false;" title="Change LinkDirection settings">LinkDirection Opt</a>');
```
le-jeu: You can use a similar syntax as
```js
$('<a>', {
    title: 'Conflict Area Info',
    click: machinaTools.showConflictAreaInfoDialog,
    html: 'Conflict Area Info',
  }).appendTo('#toolbox');
```

and keep the function `showDialog` in the local scope, or at least use simply `linkShowDirection.showDialog`

leexmario-oss:
Hey @le-jeu — could you take a look at **“TWEAKS refactoring”** when you get a chance? It hasn’t been updated since May 24, 2025. Thanks!

Issue #395, opened 2020-09-26
Add IITC-CE Mobile to F-Droid


MaxEtMoritz:
Can you add IITC-CE Mobile to [F-Droid](https://f-droid.org/) (App store for FLOSS apps) like the original IITC Mobile App?
I would really appreciate this.

To achieve this, two things should be done:
1. Fastlane file structure should be included into the Repo. [I did the file structure myself](https://github.com/MaxEtMoritz/ingress-intel-total-conversion/tree/fastlane-support), but then i realized it is already included in Pull request #305.
2. The version code in build.gradle should be a literal value to allow the F-Droid repo maintainers to automatically check for a new version of the app. Since i am no pro dev i don't know if this can be achieved automatically somehow, but maybe someone knows a way to do that other than [calculating the version code in the gradle file](https://github.com/IITC-CE/ingress-intel-total-conversion/blob/v0.30.1/mobile/app/build.gradle#L27).

I already opened a [packaging request for the app on F-Droid](https://gitlab.com/fdroid/rfp/-/issues/1503), but if you don't want to publish IITC-CE Mobile on F-Droid, you can of course cancel the request, or i can close it, too.

modos189:
Thanks to IzzySoft, it turns out that it is not necessary to have a literal value for versionCode. We use GitHub Releases, which specifies the version, and that should be enough.
I definitely support the idea of adding app to F-Droid catalog.

johnd0e:
@modos189 
What is current state of the issue?
Is there anything we can do for it?

PR #517, opened 2021-07-07
IITC Mobile: Various build changes

modos189:
* Sending a package to Google Play in .aab format instead of .apk
* Disable sending beta version to GPlay with text *[Release]* in commit message to avoid conflict due to same versionCode (with same commit) when building release version from tags.
* Added "copyonly" build option for F-Droid support #395


johnd0e:
@modos189 
Conflicts

PR #524, opened 2021-07-13
missions: use priority instead of iitcLoaded

johnd0e:
To be reliable enough `iitcLoad` hook should be preceded with `window.iitcLoaded` check, so simpler to use priority management here.

### DIFF
--- a/plugins/missions.js
+++ b/plugins/missions.js
@@ -1058,8 +1058,9 @@ window.plugin.missions = {
 			window.plugin.sync.registerMapForSync('missions', 'checkedWaypoints', this.syncCallback.bind(this), this.syncInitialed.bind(this));
 		}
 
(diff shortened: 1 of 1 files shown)


MysticJay:
@johnd0e A hook is the wrong design from the first moment!
We need a halted queue that queues plugin's init-routines,
get un-haltet when IITC is loaded
later executes and new init-routine that was too late for the first run.

McBen:
> @johnd0e A hook is the wrong design from the first moment!
> We need a halted queue that queues plugin's init-routines,
> get un-haltet when IITC is loaded
> later executes and new init-routine that was too late for the first run.

that's the `bootplugin` queue, isn't it?

modos189 (approved):

PR #536, opened 2021-10-18
IITC-Mobile: Allow plugins to register internal hostnames

ewoerner:
Internal hostnames are not opened in an external browser, but instead in the IITC mobile webview.
This allows plugins to use authentication flows that use a domain different from Google / Facebook.

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_JSInterface.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_JSInterface.java
@@ -218,4 +218,9 @@ public class IITC_JSInterface {
             mIitc.reloadIITC();
         });
     }
+
+    @JavascriptInterface
+    public void addInternalHostname(String hostname) {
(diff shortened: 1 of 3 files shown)


ewoerner:
Here is a security analysis of the patch:

1. The patch adds an additional function to call for IITC plugins. Therefore, it increases the attack surface for malicious IITC plugins.
2. DoS attack:
   A malicious plugin could call `addInternalHostname` until the application crashes with OOM. However, `IITC_JSInterface` is completely unprotected against DoS attacks anyway (`addPane`, `addPortalHighlighter`, `saveFile`)
3. Script injection:
   Since plugins are only injected into `intel.ingress.com`, adding hostnames does not allow plugins to access these hostnames.
4. Content spoofing:
   A malicious plugin could call `addInternalHostname` for a malicious domain which spoofs content of a legitimate domain (e.g. Google login). However, a malicious plugin could already spoof the same content dynamically on `intel.ingress.com`.

PR #537, opened 2021-10-21
Use history data to update uniques status

igorsol:
Use the new history feature to automatically update uniques database

### DIFF
--- a/plugins/uniques.js
+++ b/plugins/uniques.js
@@ -29,6 +29,18 @@ window.plugin.uniques.contentHTML = null;
 
 window.plugin.uniques.isHighlightActive = false;
 
+window.plugin.uniques.onPortalAdded = function(data) {
+    var opt = data.portal.options,
+        history = opt.data.history;
+    if(history) {
+        if(history.captured) {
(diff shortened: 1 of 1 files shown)


johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/wiki/Code-Style

eqeqeq, keyword-spacing

MysticJay:
Honestly I see no point in this.
1) There is no reason to COPY intel data to a local database. What would be the benefit?
2) Doing so would not match our understanding of the TOS.
3) (and most important) this routine would be called each and every time the portals get viewed. It would be executed again and again for the same portal and storing to localstorage on each call. IMHO this would cause a lot of additional CPU with little benefit.
4) I had users in the UNIQUES-Chat that wanted to keep their "own documentation" in uniques rather than creating a copy.

IMHO: in the long term we might think about removing uniques plugin from the main repository or at least the data that is in the intel.

MysticJay (commented):
Apart from my general comment, the logic is incomplete.
1) Captured is always a visited too
2) PortalScanned is always a visited too

PR #538, opened 2021-10-27
Create SECURITY.md

Limbersky23:
****


McBen:
If you wanted to report a vulnerability please check out the telegram group and start a private chat with one ot the developers.

If not: please add a description why this PR is helpful. Especially how an empty template could be useful.

PR #541, opened 2021-11-20
Resizable dialogs (JqueryUI default)

le-jeu:
Current jquery code used by IITC allows resizable dialogs by default (`option.resizable: true`)
The dialog interface from IITC also handles resize event.

https://github.com/IITC-CE/ingress-intel-total-conversion/blob/89175119cb195fe11b0ce6337564e38c91e6db54/core/code/dialog.js#L241

But the CSS for resize HTML handles is missing so it's not possible to resize dialogs yet.
This PR adds a minimal CSS to make those handles interactive.


johnd0e:
What is the source of this css?

le-jeu:
that would be https://github.com/jquery/jquery-ui/blob/main/themes/base/resizable.css

johnd0e:
So let's include it as file, instead of inlining to `style.css`.

johnd0e:
> that would be https://github.com/jquery/jquery-ui/blob/main/themes/base/resizable.css

Please include the url to https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/core/external/versions.md

johnd0e (approved):

modos189 (approved):

Issue #218, opened 2019-07-08
Image resources for IITCm


johnd0e:
I want to add more plugin panes to leftside drawer, and each plugin needs some unique icon.
Currently it's possible to use icons from here: https://github.com/IITC-CE/ingress-intel-total-conversion/tree/master/mobile/app/src/main/res

I suppose it's just a matter of putting there more png files.
Every file is expected in different dimensions: see subdirectories: drawable-hdpi/drawable-mdpi/drawable-xhdpi/drawable-xxhdpi.

It's not good just pick random png files from different sites, it'd better if all icons come from same source.
But I failed to find exact origin of current files. I do not see them at https://github.com/google/material-design-icons.

So questions:
1. What is the source of original files?
2. Is it possible to use single svg instead of png of multiple dimensions?
   At [material-design-icons](https://github.com/google/material-design-icons) I see `drawable-anydpi-v21`

modos189:
1. Probably, icons are taken from old android guidelines. Need to replace it with material icons.

2. As far as I know, there are problems with the use of svg icons in the android.

johnd0e:
Ok, it'd be great

PR #543, opened 2021-11-21
Adaptive Icon for android launcher icon

oldherl:
From Android Oreo (8.0), [Adaptive Icons](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive) are used for App icons. 
I created the image assets for adaptive icon from /assets/IITC_square.svg , hope it fits well.

Screenshot:
compare

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_JSInterface.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_JSInterface.java
@@ -218,4 +218,9 @@ public class IITC_JSInterface {
             mIitc.reloadIITC();
         });
     }
+
+    @JavascriptInterface
+    public void addInternalHostname(String hostname) {
+        mIitc.addInternalHostname(hostname);
+    }
 }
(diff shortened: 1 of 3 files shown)


modos189:
Thanks, good

Not relevant to this PR, just a thought. Because we are using the old method of accessing internal storage to store third-party plugins, it is impossible to move to Android API level 30, required for app updates on Google Play starting November 1, 2021 :( We'll have to do something about it

oldherl:
Oh, but at least we can distribute outside Google Play, such as F-Droid or direct APK download, right?

modos189:
Upload to F-Droid is not finished, but theoretically yes, it is possible

johnd0e:
@oldherl 
What about another icon, used for test builds?
https://github.com/IITC-CE/ingress-intel-total-conversion/raw/master/mobile/app/src/debug/res/drawable-xxhdpi/ic_iitcm.png

oldherl:
@johnd0e I don't know that icon. Where is it used?

johnd0e:
It was used for test builds, but not after your patch https://iitc.app/build/test/IITC_Mobile-test.apk
Compare: https://iitc.app/build/test_archive/

oldherl:
Now I understand the problem. I deleted that icon in my patch since I thought that icon was not used. I will further investigate how to convert that icon to Adaptive Icon.

johnd0e:
@oldherl May be you could help with #218 also

Issue #450, opened 2021-03-07
Fly links not working


Ren-Noir:
I have never had an issue like this. For the life of me, I cannot get fly links to work for. Ivload and reload plugins, force update, but nothing seems to work. Please help

MysticJay:
@Ren-Noir do you have an active draw?

le-jeu:
since last release, you need to draw at least one polygon to include all the portals you want to field with.
https://github.com/IITC-CE/ingress-intel-total-conversion/pull/310#issue-362714616

>   - region selection with the polygon draw tool
>   - exclusion of portals with the marker draw tool
>   - steady drawing (portals is not excluded from the drawing if they are moved out of boundaries)

![image](https://user-images.githubusercontent.com/64744459/110257699-57882580-7f9f-11eb-806f-519b4d3a7770.png)

johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/2daad393c75f45e9f121136c4717d4dec51fca4f/plugins/fly-links.js#L344

I see that plugin now silently breaks if draw-tools is not available.
That is not good at all.
Perhaps better solution would be fallback to map bounds.

@Fly33

johnd0e:
Another bug: fly-links does not react on drawing. So we have to pan slightly to initiate fly-links update.
Annoying.

johnd0e:
Also: I'd like to use Circle too, not only Polyline.

crystalwizard:
Might consider a popup if there's no polygon to inform the user that a polygon must be drawn before flylinks can activate

PR #544, opened 2021-11-22
Tidy Link: use drawtools polygons if available

le-jeu:
Use Fly links mechanism to select a zone and exclude some portals

Also fix Fly links not working without drawtools or without any drawn polygon #450

### DIFF
--- a/plugins/fly-links.js
+++ b/plugins/fly-links.js
@@ -1,7 +1,7 @@
 // @author         Fly33
 // @name           Fly Links
 // @category       Draw
-// @version        0.4.0
+// @version        0.4.1
 // @description    Calculate how to link the portals to create the largest tidy set of nested fields. Enable from the layer chooser.
 
 
@@ -252,20 +252,22 @@ window.plugin.flyLinks.updateLayer = function() {
 
   var markers = [];
   var lines = [];
-  for (var i in plugin.drawTools.drawnItems._layers) {
-    var layer = plugin.drawTools.drawnItems._layers[i];
-    if (layer instanceof L.Marker) {
-      var ll = layer.getLatLng();
-      var p = map.project(ll, window.plugin.flyLinks.PROJECT_ZOOM);
-      markers.push(p);
-    } else if (layer instanceof L.GeodesicPolyline) {
-      var p = []
-      var ll = layer.getLatLngs();
-      for (var j = 0; j < ll.length; ++j) {
-        p.push(map.project(ll[j], window.plugin.flyLinks.PROJECT_ZOOM))
-      }
-      for (var j = 1; j < p.length; ++j) {
-        lines.push([p[j-1], p[j]])
+  if (plugin.drawTools) {
+    for (var i in plugin.drawTools.drawnItems._layers) {
+      var layer = plugin.drawTools.drawnItems._layers[i];
+      if (layer instanceof L.Marker) {
+        var ll = layer.getLatLng();
+        var p = map.project(ll, window.plugin.flyLinks.PROJECT_ZOOM);
+        markers.push(p);
+      } else if (layer instanceof L.GeodesicPolyline) {
+        var p = []
+        var ll = layer.getLatLngs();
+        for (var j = 0; j < ll.length; ++j) {
+          p.push(map.project(ll[j], window.plugin.flyLinks.PROJECT_ZOOM))
+        }
+        for (var j = 1; j < p.length; ++j) {
+          lines.push([p[j-1], p[j]])
+        }
       }
     }
   }
@@ -341,26 +343,45 @@ window.plugin.flyLinks.updateLayer = function() {
 
   locations = filterMarkers(locations, markers);
 
-  for (var i in plugin.drawTools.drawnItems._layers) {
-    var layer = plugin.drawTools.drawnItems._layers[i];
-    if (layer instanceof L.GeodesicPolygon) {
-      var ll = layer.getLatLngs();
-      var polygon = [];
-      for (var i = 0; i < ll.length; ++i) {
-        var p = map.project(ll[i], window.plugin.flyLinks.PROJECT_ZOOM);
-        polygon.push(p);
+  var polygons = [];
+
+  if (plugin.drawTools) {
+    for (var i in plugin.drawTools.drawnItems._layers) {
+      var layer = plugin.drawTools.drawnItems._layers[i];
+      if (layer instanceof L.GeodesicPolygon) {
+        var ll = layer.getLatLngs();
+        var polygon = [];
+        for (var i = 0; i < ll.length; ++i) {
+          var p = map.project(ll[i], window.plugin.flyLinks.PROJECT_ZOOM);
+          polygon.push(p);
+        }
+        polygons.push(polygon);
       }
-      var points = filterPolygon(locations, polygon);
-      if (points.length >= window.plugin.flyLinks.MAX_PORTALS_TO_LINK) {
-        //alert("Some polygon contains more than 100 portals.");
-        continue;
-      }
-      var triangulation = triangulate(points, lines);
-      edges = edges.concat(triangulation.edges);
-      triangles = triangles.concat(triangulation.triangles);
     }
   }
-  
+
+  // fallback to map bounds if no drawn polygon (or no drawtools)
+  if (!polygons.length) {
+    var mapBounds = map.getBounds();
+    var polygon = [];
+    polygon.push(map.project(mapBounds.getNorthWest(), window.plugin.flyLinks.PROJECT_ZOOM));
+    polygon.push(map.project(mapBounds.getSouthWest(), window.plugin.flyLinks.PROJECT_ZOOM));
+    polygon.push(map.project(mapBounds.getSouthEast(), window.plugin.flyLinks.PROJECT_ZOOM));
+    polygon.push(map.project(mapBounds.getNorthEast(), window.plugin.flyLinks.PROJECT_ZOOM));
+    polygons.push(polygon);
+  }
+
+  $.each(polygons, function (idx, polygon) {
+    var points = filterPolygon(locations, polygon);
+    if (points.length >= window.plugin.flyLinks.MAX_PORTALS_TO_LINK) {
+      //alert("Some polygon contains more than 100 portals.");
+      return;
+    }
+    var triangulation = triangulate(points, lines);
+    edges = edges.concat(triangulation.edges);
+    triangles = triangles.concat(triangulation.triangles);
+  });
+
   $.each(edges, function(idx, edge) {
     drawLink(edge.a, edge.b, {
       color: '#FF0000',


johnd0e (changes requested):
A way too much copy-paste code in tidy-links.
Instead we should propose common api.

le-jeu:
we could have a dependency to a third plugin holding algorithm...
not sure it make thing simpler for the common user

I'm not sure about having the calculation in iitc useful if not using those two plugins.

johnd0e:
4. Facilities to filter portals (like limit to polygon and exclude portals with markers) should be shared, because they be useful for a lot of plugins. And it's not good to 'invent the wheel' in every such plugin.
  I am thinking about sort of 'portal filtering api' (will open separate issue about this).

_Originally posted by @johnd0e in https://github.com/IITC-CE/ingress-intel-total-conversion/issues/203#issuecomment-578163490_

1. There are a lot of plugins that could benefit from portal filtering api, incl. Missions, portal-counts, scoreboard, ...
2. Besides of relying on draw-tools there a lot of other ways to implement portal selection: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/27

le-jeu:
since it's bound to drawtools polygon... should it be part of draw tools ?

johnd0e:
> since it's bound to drawtools polygon... should it be part of draw tools ?

Why not.
Still, I'd prefer to see this api in general form, to be able to implement alternative portal selection tools.

MysticJay:
@johnd0e The polygon that is used for the filter could be selected by its color.
so if i draw fields/links in color A,then simply switch to color B draw any other polygon(s) and only filter to polygons with that same color.

one problem with DT polygons is that they are involved in crosslink calculation.

core/code/utils_misc.js
```diff
@@ -474,6 +474,42 @@ window.clampLatLngBounds = function (bounds) {
   return L.latLngBounds(clampLatLng(SW), clampLatLng(NE));
 }
 
+
+/*
+pnpoly Copyright (c) 1970-2003, Wm. Randolph Franklin
```
johnd0e: Please check this with edge cases, such as "portal on the vertice", "portal on the edge".

plugins/draw-tools.js
```diff
@@ -1,7 +1,7 @@
 // @author         breunigs
 // @name           Draw tools
 // @category       Draw
-// @version        0.8.1
+// @version        0.8.2
```
johnd0e:
```suggestion
// @version        0.9.0
```

plugins/draw-tools.js
```diff
@@ -728,6 +728,47 @@ window.plugin.drawTools.initMPE = function(){
   });
 }
 
+//
+// Basic getters for filter to use
+//
+
+window.plugin.drawTools.getMarkers = function () {
+  var markers = [];
+  for (var i in plugin.drawTools.drawnItems._layers) {
```
johnd0e: Better not use private properties

plugins/draw-tools.js
```diff
@@ -728,6 +728,47 @@ window.plugin.drawTools.initMPE = function(){
...
+    if (layer instanceof L.Marker) {
+      markers.push(layer.getLatLng());
+    }
+  }
+  return markers;
+};
+
+window.plugin.drawTools.getPolylines = function () {
+  var polylines = [];
+  for (var i in plugin.drawTools.drawnItems._layers) {
```
johnd0e: Same

plugins/draw-tools.js
```diff
@@ -728,6 +728,47 @@ window.plugin.drawTools.initMPE = function(){
...
+    if (layer instanceof L.GeodesicPolyline) {
+      polylines.push(layer.getLatLngs());
+    }
+  }
+  return polylines;
+};
+
+window.plugin.drawTools.getPolygons = function () {
+  var polygons = [];
+  for (var i in plugin.drawTools.drawnItems._layers) {
```
johnd0e: same

plugins/draw-tools.js
```diff
@@ -728,6 +728,47 @@ window.plugin.drawTools.initMPE = function(){
...
+    var layer = plugin.drawTools.drawnItems._layers[i];
+    if (layer instanceof L.GeodesicPolygon) {
+      polygons.push(layer.getLatLngs());
```
johnd0e: Need to investigate if we can access projected data directly instead.

plugins/draw-tools.js
```diff
@@ -728,6 +728,47 @@ window.plugin.drawTools.initMPE = function(){
...
+    }
+  }
+  return polygons;
+};
+
+//
+//
+//
```
johnd0e: ?

plugins/fix-china-map-offset.js
```diff
@@ -196,11 +196,13 @@ var insane_is_in_china = (function () { // adapted from https://github.com/Artor
...
 
-  function isInChina (lat, lon) {
+  var LLs = lats.map(function (_, i) { return { lat: lats[i], lng: lons[i]}; });
+
+  function isInChina (lat, lng) {
     // Yank out South China Sea as it's not distorted.
     if (lat >= 17.754 && lat <= 55.8271 &&
-        lon >= 72.004 && lon <= 137.8347) {
-      return pnpoly(lats, lons, lat, lon);
+        lng >= 72.004 && lng <= 137.8347) {
```
johnd0e: The names are from upstream, please do not change without need

core/code/utils_misc.js
```diff
@@ -474,6 +474,42 @@ window.clampLatLngBounds = function (bounds) {
...
+     materials provided with the distribution.
+  3. The name of W. Randolph Franklin may not be used to endorse or promote products derived from this Software without
+     specific prior written permission.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+window.pnpoly = function(latlngs, point) {
```
johnd0e: This is terribly wrong, as we cannot use latlngs here, this algorithm is for 2d (projected) points.

plugins/fly-links.js
```diff
@@ -307,76 +306,18 @@ window.plugin.flyLinks.updateLayer = function() {
...
+      return window.pnpoly(polygon, ll);
+    });
     if (points.length >= window.plugin.flyLinks.MAX_PORTALS_TO_LINK) {
       //alert("Some polygon contains more than 100 portals.");
       return;
     }
+    if (!points.length) return;
+    for (var i = 0; i < points.length; i++) {
+      points[i] = map.project(points[i], window.plugin.flyLinks.PROJECT_ZOOM);
+    }
```
johnd0e: We need shared utility for this.

johnd0e (changes requested):

core/code/utils_misc.js
le-jeu: that's from fix-china-offset. 
The usage of the algorithm in IITC was never correct in the first place since everything is about a geodesic, that pnpoly cannot cover, whatever the projection.
Question is, is this algorithm good enough for the usage we have in IITC.

plugins/draw-tools.js
le-jeu: there is no public method/property, it also depends on the zoom and the renderer afaik

core/code/utils_misc.js
le-jeu: On a second aspect, we do want a function that works on latlng, more than a pnpoly.

plugins/draw-tools.js
johnd0e: Right, but we can use private (I hope).
It definitely depends on zoom, but the result (in/out) is the same.

core/code/utils_misc.js
johnd0e: So let's add another,  more high-level function, operating by latlngs. Projecting them on 2d space, passing to pnpoly, and returning results,

core/code/utils_misc.js
le-jeu: in fact, pnpoly (by Franklin) doesn't handle well boundaries
it may not be the algorithm we are looking for...
But portals on edge are not common, so just checking vertices may be enough

plugins/draw-tools.js
le-jeu: the `._rings` may be the one

core/code/utils_misc.js
johnd0e: https://github.com/IITC-CE/ingress-intel-total-conversion/issues/552

> so just checking vertices may be enough

ok

core/code/utils_misc.js
le-jeu: That would still be incorrect as we use geodesic instead of straight lines in draw tools.
But again, this is good enough and would match a tool that would not use geodesic (as any selection tool).
Maybe we need to extend drawtool (or use another plugin) to draw standard polyline/polygon.

le-jeu: Or we can use the interpolated coordinates when using geodesic polygon as filtering

johnd0e: What is interpolated coordinates?

core/code/utils_misc.js
```diff
@@ -495,19 +495,23 @@ WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEM
...
+  // j records previous value. Also handles wrapping around.
+  for (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
+    inside ^= polygon[i].y > point.y !== polygon[j].y > point.y &&
+              point.x - polygon[i].x < (polygon[j].x - polygon[i].x) * (point.y - polygon[i].y) / (polygon[j].y - polygon[i].y);
+  }
+  // Let's make js as magical as C. Yay.
+  return !!inside;
+};
+
+window.latlngPNPOLY = function(latlngs, point) {
```
johnd0e: `window.pnpolyLL`

johnd0e (changes requested):

core/code/utils_misc.js
le-jeu: all points calculated to draw a geodesic polyline

johnd0e: In fact there is not so much difference between geodesic and planar on those zoom levels where filtering makes sense

le-jeu: that's also true with latlng vs projected

johnd0e: Btw, draw items are geodesic, so filtering will be geodesic as well.

le-jeu: my point, but what do you mean by "will be" ? is it "must/should/may be" ?

johnd0e: I mean as we are using geodesic shapes for bounds - our calculations are geodesically correct. Right now.

johnd0e: But why don't you use `_rings`?

core/code/utils_misc.js
johnd0e: Library of my choice: point-in-polygon-hao
Size (min): ~700b

core/code/utils_misc.js
le-jeu: it's not because currently, we only use endpoints. Interpolated points are only used for rendering, `getLatLngs()` ignore them

johnd0e: >  Interpolated points are only used for rendering

So why not make use of them? Why should we care about getLatLngs?

plugins/tidy-links.js
```diff
@@ -24,59 +24,103 @@ tidyLinks.STROKE_STYLE = { // https://leafletjs.com/reference-1.4.0.html#polylin
...
+  // fallback to map bounds if no drawn polygon (or no drawtools)
+  if (!polygons.length) {
+    var bounds = map.getPixelBounds();
+    var origin = map.getPixelOrigin();
+    polygons.push([
+      bounds.getBottomLeft().subtract(origin),
+      bounds.getBottomRight().subtract(origin),
+      bounds.getTopRight().subtract(origin),
+      bounds.getTopLeft().subtract(origin),
+    ]);
```
johnd0e: I do not like this approach.
Tidy links should not count draw-tools polygons.
And tidy links should not even call pnpoly directly.

Things must be much simpler, and more high-level: pass point - get back boolean.

johnd0e (changes requested):

plugins/tidy-links.js
le-jeu: Tidy/Fly links need to know about drawtools polygons : 
image
We use a polygon array because we repeat the same algorithm for each polygon.

johnd0e: Not sure that it's really useful.
If you like this then ok.

But I still propose to invent better API.
E.g. special draw-tools function `getPortalFilters`.
And remember that we have to react on drawings change.

johnd0e: I'd like to see all low-level details incapsulated (inside of draw-tools).
And all plugins access just high-level api.

- `locationFilters:Array` of filters, with such interface:
  + `filter(portal: CircleMarker|LatLng)`.
- Also we need "changed" event to be able to react on selection changes.

And that's all.
Client plugins should not care about points format, bounding polygons, projections, exclusion markers and lines.

le-jeu: It's useful, for instance you want to max AP but a long link cuts the area in two parts.
And, that's current fly links feature.

This PR is about fixing Fly link when drawtools isn't available and add the same feature to tidy links
We can always imagine better API, but it should match the use cases at least.

Here, we'd like a "selection" API to return the sets of portal we are in, their colors for instance, in order to run the processing on each set.

johnd0e: > And, that's current fly links feature.

Ok

> We can always imagine better API, but it should match the use cases at least.

Better API - simpler api. But sure we should cover most use cases.

> we'd like a "selection" API to return the sets of portal we are in, their colors for instance, 

That would be redundant overcomplication.

johnd0e: > This PR is about fixing Fly link when drawtools isn't available

That could be merged in it's initial form.

> and add the same feature to tidy links

But this demands more work.

le-jeu: > But this demands more work.

but it doesn't need either a new API. Here common parts are barely factorized to avoid algorithm duplication. Fly/tidy are coupled to drawtools (forcing links in fly links for instance) and that is fine.

pnpoly can be change later (preserving the signature once released)

Select API is something to be discussed in parallel.

> That would be redundant overcomplication.

why redundant ? 
```
getSelectRegions(): id[];
isSelected(portal/latlng,id?): boolean;
getSelectedPortals(id?): guids/portals[];
```
this would lead to code like 
```
for (id in getSelectRegions()) {
  doSomethingWith(getSelectedPortal(id));
}
```
with something like for instance
```
addFilter(filter: (latlng/portal)=>boolean,id?): id;
```

but this is another discussion imo

johnd0e: > but it doesn't need either a new API

I have severe objections both against massive copy-paste, and early unmature api decisions.

> why redundant ?

I mean redundant in API, but could be implemented in draw-tools UI instead.

> getSelectRegions(): id[]; // etc..

That would provide exactly the same functionality as proposed above https://github.com/IITC-CE/ingress-intel-total-conversion/pull/544#discussion_r762807857
No?

le-jeu: > I have severe objections both against massive copy-paste, and early unmature api decisions.

sure. but here, there is no API change right ? (and copy/paste was removed)
I especially don't want to add filter/select api here because I find it not mature enough

> That has exactly the same functionality as proposed above

yes, this was an illustration (with different naming) of what was saying about : (wrt to redundancy)
> return the sets of portal _[by id/index]_ we are in, their colors for instance

johnd0e: > there is no API change right ?

window.pnpoly, window.pnpolyLL

> because I find it not mature enough

It will get more mature as we develop it. In this PR, or in another.

le-jeu: point in polygon is not restricted to filter/select, that could even extend L.Polygon.
that's why it's exposed regardless of the selection api you want to build

PR #553, opened 2021-12-08
implement location filters api, and other changes

### LINKED ISSUE #450: Fly links not working

johnd0e:
* utils_misc.js: common function `window.pnpoly`.
  Applied to:
  - fix-china-map-offset
  - fly-links
  - layer-count

* draw-tools: implement _location filters api_.
  Includes entities inside of drawn polygons (every polygon separately).
  Portals with markers are excluded from filters.
  * `window.plugin.drawTools.getLocationFilters()`, returns array of functions:
    `filter(point:portal|Point|LatLng):Boolean`
  * `window.plugin.drawTools.filterEvents:Evented`, with single event "changed".
  Applied to:
    - fly-links
    - tidy-links

* Make use of already projected coords (current zoom), instead of reprojecting.
  Applied to:
  - fly-links
  - tidy-links
  - layer-count (before: used unprojected latlngs)
  - location filters api

  Note: fix-china-map-offset still uses unprojected latlngs in pnpoly, like upstream does.

* fly-links
  - fix working w/o draw-tools
  - fix working w/o drawn polygon (fall back to visible map bounds)
  - redraw on drawn polygon changed
  - on draw: use stored latlngs instead of unprojecting points

Fix #450, close #544 (supersedes)

### DIFF
--- a/core/code/utils_misc.js
+++ b/core/code/utils_misc.js
@@ -474,6 +474,37 @@ window.clampLatLngBounds = function (bounds) {
   return L.latLngBounds(clampLatLng(SW), clampLatLng(NE));
 }
 
+/*
+pnpoly Copyright (c) 1970-2003, Wm. Randolph Franklin
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit
+persons to whom the Software is furnished to do so, subject to the following conditions:
+
+  1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+     disclaimers.
+  2. Redistributions in binary form must reproduce the above copyright notice in the documentation and/or other
+     materials provided with the distribution.
+  3. The name of W. Randolph Franklin may not be used to endorse or promote products derived from this Software without
+     specific prior written permission.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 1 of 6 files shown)


core/code/utils_misc.js
```diff
@@ -474,6 +474,37 @@ window.clampLatLngBounds = function (bounds) {
...
+  var inside = 0;
+  // j records previous value. Also handles wrapping around.
+  for (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
+    inside ^= polygon[i].y > point.y !== polygon[j].y > point.y &&
+              point.x - polygon[i].x < (polygon[j].x - polygon[i].x) * (point.y - polygon[i].y) / (polygon[j].y - polygon[i].y);
+  }
+  // Let's make js as magical as C. Yay.
+  return !!inside;
+};
+
```
nheir, line 507: point-in-polygon-hao has a slight different signature, because it accepts multi-polygon 
should this function share the same signature even if only the first polygon is used ?
or should we have concurrent point-in-polygon implementations with different names available ?

plugins/draw-tools.js
```diff
@@ -728,11 +728,65 @@ window.plugin.drawTools.initMPE = function(){
...
   loadExternals();
   window.plugin.drawTools.boot();
   window.plugin.drawTools.initMPE();
+
+  var filterEvents = new L.Evented();
+  window.map.on('draw:created draw:edited draw:deleted', function (e) {
+    filterEvents.fire('changed', { originalEvent: e });
+  });
+  filterEvents.on('changed', function () {
+    cachedFilters = null;
```
nheir, line 785: I suppose cached filter should be clear after zoom (and maybe map move)

plugins/fly-links.js
```diff
@@ -305,62 +267,35 @@ window.plugin.flyLinks.updateLayer = function() {
...
-        //alert("Some polygon contains more than 100 portals.");
-        continue;
+  filters.forEach(function (filter) {
+    var points = [];
+    for (var guid in window.portals) {
+      var location = window.portals[guid];
+      if (filter(location)) {
+        var point = location._point.clone();
+        point.latlng = location._latlng;
+        points.push(point);
```
nheir, line 286: nicely done

nheir (commented):
nvm that

modos189 (approved):

PR #566, opened 2022-06-25
Ornaments & Beacons 

MysticJay:
core/ornaments.js

Currently there are a number of plugins available that allow a larger variety of ornaments and beacons to be selected from. E.g. during an anomaly as an operator you would rather like to suppress ornaments from e.g. "toast" or "NIA"-beacons, or concentrate on everything with shards while no need for battle beacons an their related ornaments for victory and rewards. Existing plugins do this by replacing the core routines.

The possible number of ornaments and beacons is not limited at all. NIA could come up with completely new ones.
Current ornaments.js does not provide the needed flexibility nor does it have any dynamic adoption of new ornaments.

This PR has two design goals:
1) an easy way for any user to show/hide any ornament
2) allow replacing the default ornaments with some other image

Implementation
1) is implemented by offering a global array (ornamentsUrls), which can be filled with own images e.g. by a plugin.

2) is implemented by two things: 
a) additional layers for groups of ornaments
b) an options-dialog that allows to move any ornament to an "Excluded"-Layer, superseding  a)
OrnamentsDialog


core/code/ornaments.js
```diff
@@ -19,6 +19,8 @@ window.ornaments = {
 
   OVERLAY_SIZE: 60,
   OVERLAY_OPACITY: 0.6,
+  iconUrls: [],
```
nheir: maybe `{}` instead of `[]`

core/code/ornaments.js
```diff
@@ -27,49 +29,154 @@ window.ornaments = {
...
+
         return L.marker(portal.getLatLng(), {
-          icon: L.icon({
-            iconUrl: '//commondatastorage.googleapis.com/ingress.com/img/map_icons/marker_images/' + ornament + '.png',
+          icon: L.icon({  
+            iconUrl: iconUrl,
             iconSize: [size, size],
-            iconAnchor: [size/2, size/2] // https://github.com/IITC-CE/Leaflet.Canvas-Markers/issues/4
+            iconAnchor: anchor, // https://github.com/IITC-CE/Leaflet.Canvas-Markers/issues/4
+            className: 'no-pointer-events'
```
nheir, line 138: why is `no-pointer-events` necessary ?

core/code/ornaments.js
```diff
@@ -27,49 +29,154 @@ window.ornaments = {
...
+        if (ornament.startsWith('sc')) {
+          layer = this._scout;
+        }
+
+        if (ornament.startsWith('peB')) {
+          layer = this._battle;
+        }
+
+        if (typeof (window.ornaments.iconUrls[ornament]) !== 'undefined') {
+          opacity = 1;
```
nheir: any reason to make opacity to 1 ?
in addition (but it starts to become too generic), custom ornament icon could have different size, anchor position etc
it could use a specific marker class instead of L.marker

core/code/ornaments.js
```diff
@@ -27,49 +29,154 @@ window.ornaments = {
...
+    this._excluded = layerGroup(); //to keep excluded ornaments in an own layer
+    
     window.layerChooser.addOverlay(this._layer, 'Ornaments');
     window.layerChooser.addOverlay(this._beacons, 'Beacons');
     window.layerChooser.addOverlay(this._frackers, 'Frackers');
+    window.layerChooser.addOverlay(this._scout, 'Scouting');
+    window.layerChooser.addOverlay(this._battle, 'Battle');
+    window.layerChooser.addOverlay(this._excluded, 'Excluded'); //just for testing
+
+    $('#toolbox').append('<a onclick="window.ornaments.ornamentsOpt();return false;" accesskey="o" title="Edit ornament exclusions [o]">Ornaments Opt</a>');
```
nheir: prefer syntax like `artifact.js`
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a154ef3186e31234e9bf20a5028dce2d513c723c/core/code/artifact.js#L30-L37

core/code/ornaments.js
```diff
@@ -27,49 +29,154 @@ window.ornaments = {
...
+             + 'peTOASTY<br>'
+             + 'peBB_BATTLE_RARE<br>'
+             + 'peBB_BATTLE<br>'
+             + '<br>'
+             + '</div>';
+
+    dialog({
+      html:html,
+      id:'ornamentsOpt',
+      dialogClass:'ui-dialog-content',
```
nheir: the `dialogClass` is set to this value by default

core/code/ornaments.js
MysticJay: otherwise the icon would cover other portals. We only want to interact with portal-markers.

core/code/ornaments.js
MysticJay: no longer in code

core/code/ornaments.js
MysticJay: Things might evolve in future, but is there really a need for generic markers? Everyone is looking into fancy markers, but the real issue is the load of markers that are shown during an anomaly.

So initial issue was to hide/exclude ornaments from showing up on intel. Kythera Phase 3 was pretty messy with a lot of ornaments still showing from last rounds, so I looked for an easy way to get rid of them.

But as the existing plugins added iconUrls, this seemed to be "wanted". The settings for offset or opacity were simply taken from there.

What value for opacity would you prefer?

But what would a generic marker offer to the user?
I do not want to implement a lot of stuff, because "it can be done".

core/code/ornaments.js
MysticJay: outdated

core/code/ornaments.js
MysticJay: done

MysticJay:
When implementing the changes I found it would be worth documenting some facts about ornaments

1. Ornaments are non-interactive. No intel function related to them.
2. Ornament-IDs might change or be reused. Examples: "ap2" is currently the ornament to highlight portals that will be used to place RBBs in Munich, but in earlier anomalies they denoted the portals belonging to "Cluster 2". (thus "ap1" used for "Cluster 1"). Currently it is not clear (to me) if "peBB_BLM" was replaced by "peBB_BATTLE"/"pe_BE_BATTLE_RARE", if "peBB_BLM" is still used for an unclear purpose or dropped completely from usage.
3. Some of the ornament-names encode additional information, where the stock icon is the same: The BB-Reward for a Cat-IV-victory is named 'peBR_REWARD-10_200_150'. While for the intel these values have no meaning, we believe they denote the output (duration, percentage, total number of hacks).


So what does that mean for the implementation in the core (IITC itself):

- All layers that we implement are only filtering on those IDs. If Niantic changes anything, the filter will no longer work correctly.
- Using the "additional information" as explained in 3 might result in wrong interpretation.
- Assigning "common" names to the IDs might also result in wrong interpretation.

core/code/ornaments.js
```diff
@@ -27,38 +39,105 @@ window.ornaments = {
...
+    this._scout = layerGroup();
+    this._battle = layerGroup();
+    this._excluded = layerGroup(); // to keep excluded ornaments in an own layer
+
     window.layerChooser.addOverlay(this._layer, 'Ornaments');
     window.layerChooser.addOverlay(this._beacons, 'Beacons');
     window.layerChooser.addOverlay(this._frackers, 'Frackers');
+    window.layerChooser.addOverlay(this._scout, 'Scouting');
+    window.layerChooser.addOverlay(this._battle, 'Battle');
+    window.layerChooser.addOverlay(this._excluded, 'Excluded');
```
modos189:
```suggestion
    window.layerChooser.addOverlay(this._excluded, 'Excluded', {default: false});
```
Hiding excluded ornaments by default

MysticJay:
@MysticJay anyone here who can confirm that OrnamentId='peBB_BLM' belongs to BlackLivesMatter-Beacon?

Confirmed in private message.

core/code/ornaments.js
le-jeu: unless disabling the canvas marker layer, ornaments are render behind the portal/link/field layer in a different canvas. The default canvas (for portal/draw tools/...) captures all events.
The marker is also set as not interactive.

Did you get issues with ornaments preventing interaction with portals ?

core/code/ornaments.js
MysticJay: implemented

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+// can add an entry to excludedOrnaments, which will compared (startsWith) to all known and
+// future IDs. example: "ap" to exclude all Ornaments for anomalies (ap1, ap2, ap2_v)
 
+/* global L, dialog */
 
 window.ornaments = {
 
   OVERLAY_SIZE: 60,
   OVERLAY_OPACITY: 0.6,
+  iconUrls: [],
```
McBen: unused?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+  icon:[],
```
McBen: a) how about something like `ornamentsDefinition`

b) this is used by the plugin only, right? then missing structure information

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+  excludedOrnaments: [],
+  knownOrnaments: {},
 
   setup: function () {
     this._portals = {};
-    var layerGroup = L.layerGroup;
+    window.ornaments.layerGroup = L.layerGroup;
```
McBen: `window.ornaments` equal `this`

When storing the function pointer wouldn't be something like `.createLayerGroup` more intiutive?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
-      layerGroup = L.canvasIconLayer;
+      window.ornaments.layerGroup = L.canvasIconLayer;
       L.CanvasIconLayer.mergeOptions({ padding: L.Canvas.prototype.options.padding });
     }
-    this._layer = layerGroup();
-    this._beacons = layerGroup();
-    this._frackers = layerGroup();
+    this.load();
+
+    this._layer = window.ornaments.layerGroup();
```
McBen: can we use the new `createLayer` function here?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+    this._excluded = window.ornaments.layerGroup(); // to keep excluded ornaments in an own layer
+
     window.layerChooser.addOverlay(this._layer, 'Ornaments');
-    window.layerChooser.addOverlay(this._beacons, 'Beacons');
-    window.layerChooser.addOverlay(this._frackers, 'Frackers');
+    window.layerChooser.addOverlay(this._excluded, 'Excluded', {default: false});
```
McBen: maybe the name `_excluded` could be improved? _layerExcluded ?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+
+    $('<a>')
+      .html('Ornaments Opt')
+      .attr({
+        id: 'ornaments-toolbox-link',
+        title: 'Edit ornament exclusions',
+        accesskey: 'o'
+      })
+      .click(window.ornaments.ornamentsOpt)
+      .appendTo('#toolbox');
```
McBen, line 72: I personally prefer the compact defintion
```suggestion
    $('<a>', {
     text:'Ornaments Opt',
     id: 'ornaments-toolbox-link',  
     title: 'Edit ornament exclusions',
     accesskey: 'o',
     click: window.ornaments.ornamentsOpt})
      .appendTo('#toolbox');
```

BTW: is "id" used?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+  },
+  createLayer: function (layerID) {
+    window.ornaments[layerID] = window.ornaments.layerGroup();
```
McBen: this will be a dangerous call:  `window.ornaments.createLayer("createLayer");`

better store layers in a different structure.

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+              case 0:
+                anchor = [size / 2, size / 2];
+                break;
+              case -1:
+                anchor = [size / 2, - size ];
+              }
+            }
+          }
+        }
+
```
McBen: this is the plugin code and should be in the plugin
and not as dead code in the core.

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+        var exclude = false;
+        if (this.excludedOrnaments && this.excludedOrnaments !== ['']) {
```
McBen, line 123: wouldn't it be easier to make sure "excludedOrnaments" is in a valid state?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+          exclude = this.excludedOrnaments.some( function(pattern) {
+            return ornament.startsWith(pattern);
+          });
+        }
+        exclude += window.ornaments.knownOrnaments[ornament];
```
McBen: don't do arithemtics on booleans
```suggestion
        exclude = exclude | window.ornaments.knownOrnaments[ornament];
```

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
       });
       delete this._portals[guid];
     }
+  },
+  initOrnaments: function () {
+    this.knownOrnaments = {};
+    this.save();
```
McBen, line 161: `excludedOrnaments` doesn't need to be initialized?
note: if initialized here 'load' will reset this if  `knownOrnaments` is (for what reason ever) not in localstorage.

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
...
+  },
+
+  load: function () {
+    var dataStr;
+    try {
+      dataStr = localStorage['excludedOrnaments'];
```
McBen:
```suggestion
      dataStr = localStorage.getItem('excludedOrnaments');
```
geItem/setItem are the standarized way to access localStorage.
(note: getItem will return `null` if value doesn't exists)

anyway, `localStorage['excludedOrnaments'];` will also work fine.

core/external/leaflet.canvas-markers.js
```diff
@@ -170,7 +170,7 @@ function layerFactory (L) {
         },
 
         _redraw: function () {
-            L.Canvas.prototype._redraw.call(this);
+            this._redrawRequest = this._redrawRequest || L.Util.requestAnimFrame(L.Canvas.prototype._redraw, this);
```
McBen, line 173: AFAIK "external" is for 3rd-Party code that shouldn't be altered if possible.
 
Is this code from a missing patch?

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
...
+
+  },
+
+  save: function () {
+    localStorage['excludedOrnaments'] = JSON.stringify(this.excludedOrnaments);
+    localStorage['knownOrnaments'] = JSON.stringify(this.knownOrnaments);
+  },
+
+  // reload: addPortal also calls removePortal
+  reload: function () {
```
McBen, line 192: this is for redraw all ornaments?

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
...
+    }
+  },
+
+  processInput: function () {
+    window.ornaments.excludedOrnaments = $('#ornaments_E').val().split(/[\s,]+/);
+    window.ornaments.excludedOrnaments = window.ornaments.excludedOrnaments.filter(function (ornamentCode) { return ornamentCode !== ''; });
+    // process the input from the checkboxes
+    for (var ornamentCode in window.ornaments.knownOrnaments) {
+      var input = document.getElementById('chk_orn_'+ornamentCode);
+      window.ornaments.knownOrnaments[ornamentCode] = input ? input.checked : false; // <- default value if the input is not found for unexpected reason
```
McBen: with JQuery:
```suggestion
      var input = $("#chk_orn_"+ornamentCode);
      window.ornaments.knownOrnaments[ornamentCode] = input.is(":checked");
```
(not live tested )

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
...
+    }
+  },
+
+  ornamentsList: function() {
+    var text ='';
+    var sortedIDs = Object.keys(window.ornaments.knownOrnaments).sort();
+
+    sortedIDs.forEach ( function (ornamentCode) {
+      var hidden = window.ornaments.excludedOrnaments.some( function(code) {
```
McBen, line 213: so "excludedOrnaments" is always initilized? (see comment for line 107)

core/code/ornaments.js
```diff
@@ -71,5 +139,109 @@ window.ornaments = {
...
+  replaceOL: function () {
+    document.getElementById('ornamentsList').innerHTML = window.ornaments.ornamentsList();
+  },
+
+  ornamentsOpt: function () {
+    var excludedIDs = window.ornaments.excludedOrnaments.toString();
+    var html = '<div class="ornamentsOpts">'
+             + 'Hide Ornaments from IITC that start with:<br>'
+             + '<input type="text" value="' + excludedIDs + '" id="ornaments_E"'
+             + ' onchange="window.ornaments.processInput();window.ornaments.replaceOL();window.ornaments.save();window.ornaments.reload()"></input><br>'
```
McBen: how about putting these 4 calls in an extra function?

plugins/ornament-icons.js
```diff
@@ -0,0 +1,107 @@
+// @author         johtata
```
McBen, line 1: Make sure these images are free to use. 
( just a warning, I didn't reviewed these )

McBen (changes requested):
a) `window.ornaments[layerID] = ` is dangerous code
b) plugin code should go into a plugin not into the core.
(or include all into the core)

core/code/ornaments.js
MysticJay: removed

core/code/ornaments.js
MysticJay: That is the way it used to be implemented. I just needed it available for .createLayer.
"this" should work here. still I don't like it.

core/code/ornaments.js
MysticJay: a) sure, why not. But then it would be `ornamentDefinitions`. it is just a very long name.
imho `window.ornaments.ornamentDefinitions[ornament]` hard to read compared to `window.ornaments.definitions[ornament]`

core/code/ornaments.js
MysticJay: without "Safety checks" code is always easier. 
The content depends on the users's input.
An empty string is a valid input and results in `['']`.
I think the rest of the code would be more complex.

core/code/ornaments.js
MysticJay: changed

core/code/ornaments.js
MysticJay: No I did not want to reset excludedOrnaments here. This is just to "forget" what IDs IITC has learned over the past sessions. excludedOrnaments can be reset by clearing that input field.
usecase: you do never want to see those "sc5_p", simply add it to excluded.

core/external/leaflet.canvas-markers.js
MysticJay: No idea.
@le-jeu ?

core/code/ornaments.js
MysticJay: yes

plugins/ornament-icons.js
MysticJay: redid every image that _we_ "deliver"
while I also redid TOASTY, I am not sure about the original one.

core/external/leaflet.canvas-markers.js
modos189: We use [our own fork of leaflet.canvas-marker.js](https://github.com/IITC-CE/Leaflet.Canvas-Markers), so it's not a big deal if it improves performance and has no side issues

core/code/ornaments.js
MysticJay: Yes, I think so. It gets initialized in .setup.

core/code/ornaments.js
MysticJay: implemented

core/code/ornaments.js
MysticJay: implemented and tested.

core/code/ornaments.js
MysticJay: implemented

core/code/ornaments.js
MysticJay: introduced window.ornaments.layers to hold layers only.

core/code/ornaments.js
MysticJay: which one is used in IITC?

core/external/leaflet.canvas-markers.js
McBen: @modos189 
this change is not in https://github.com/IITC-CE/Leaflet.Canvas-Markers
I only see it here. Hidden inside a "dialog" commit.

Doesn't these kind of changed make future updates difficult to manage?

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+// can add an entry to excludedOrnaments, which will compared (startsWith) to all known and
+// future IDs. example: "ap" to exclude all Ornaments for anomalies (ap1, ap2, ap2_v)
 
+/* global L, dialog */
 
 window.ornaments = {
 
   OVERLAY_SIZE: 60,
   OVERLAY_OPACITY: 0.6,
+  icon:[],
```
le-jeu: `.icon` is an array but it's used as a object map in the code

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
-    this._beacons = layerGroup();
-    this._frackers = layerGroup();
-    window.layerChooser.addOverlay(this._layer, 'Ornaments');
-    window.layerChooser.addOverlay(this._beacons, 'Beacons');
-    window.layerChooser.addOverlay(this._frackers, 'Frackers');
+    this.load();
+
+    this.layers = {};
+    this.layers._layer = window.ornaments.layerGroup();
+    this.layers._excluded = window.ornaments.layerGroup(); // to keep excluded ornaments in an own layer
```
le-jeu: no needs to use undescored names, you can directly use `this.layers` as a mapping object as it's used for the other layer IDs.
```suggestion
    this.layers = {};
    this.layers['Ornaments'] = window.ornaments.layerGroup();
    this.layers['Excluded ornaments'] = window.ornaments.layerGroup(); // to keep excluded ornaments in an own layer
```

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+              case 1:
+                anchor = [size / 2, size];
+                break;
+              case 0:
+                anchor = [size / 2, size / 2];
+                break;
+              case -1:
+                anchor = [size / 2, - size ];
+              }
+            }
```
le-jeu, line 115: this `.offset` in `-1,0,1` looks oddly specific and restrictive to me.
why not having `.offset` an array instead default value  `[size/2,size/2]` ?
if a plugin wants an icon with some offset, then it defines `.offset: [dx, dy]`

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+
+        if (typeof (window.ornaments.icon[ornament]) !== 'undefined') {
+          if (window.ornaments.icon[ornament].layer) {
+            if (window.ornaments.layers[window.ornaments.icon[ornament].layer] === undefined){
+              console.log ('Add missing layer: ',window.ornaments.icon[ornament].layer);
+              window.ornaments.createLayer(window.ornaments.icon[ornament].layer);
+            }
+            layer =  window.ornaments.layers[window.ornaments.icon[ornament].layer];
+          }
+          opacity = 1;
```
le-jeu: I don't understand why you overwrite the opacity to 1 if the ornaments is redefined in a plugin.

Either the plugin changes `window.ornaments.OVERLAY_OPACITY` for all ornaments, or you can use a `.opacity` attribute that would be set in the `.icon[theOrnament]` data

core/code/ornaments.js
le-jeu: this is not resolved, ornaments are not rendered as html elements, so they don't have class. 

> Did you get issues with ornaments preventing interaction with portals ?

core/code/ornaments.js
```diff
@@ -71,5 +139,116 @@ window.ornaments = {
...
+  },
+
+  load: function () {
+    var dataStr;
+    try {
+      dataStr = localStorage.getItem('excludedOrnaments');
+      if (dataStr === undefined) { return; }
+      this.excludedOrnaments = JSON.parse(dataStr);
+    } catch (e) {
+      console.warn('ornaments: failed to load excludedOrnaments from localStorage: '+e);
```
le-jeu: use `log.warn` instead of `console.warn`

note: to see log messages (as a dev), you need to define `localStorage['log'] = 'warn'` (for warn log level) (see https://ulog.js.org/tutorial/)

core/code/ornaments.js
```diff
@@ -13,52 +13,120 @@
...
+
+        if (!this.knownOrnaments[ornament]) {
+          this.knownOrnaments[ornament]=false;
+        }
+
+        if (typeof (window.ornaments.icon[ornament]) !== 'undefined') {
+          if (window.ornaments.icon[ornament].layer) {
+            if (window.ornaments.layers[window.ornaments.icon[ornament].layer] === undefined){
+              console.log ('Add missing layer: ',window.ornaments.icon[ornament].layer);
+              window.ornaments.createLayer(window.ornaments.icon[ornament].layer);
```
le-jeu: this function mixes `this` and `window.ornaments`, this is error-prone
note that `this` may not be `window.ornaments` depending on how the function is called. (for instance, given the current code, when `ornamentOpt` is called, `this` is not `window.ornaments`)

core/code/ornaments.js
McBen: IITC has no rule here.  Every dev had his own way to do it.

as said both ways are fine. I only switched "html" to "text" here.

core/code/ornaments.js
MysticJay: changed

core/code/ornaments.js
MysticJay: _layer is legacy. other, (3rd Party) plugins might be using it?

le-jeu: you are right, but `this._layer` was legacy, this is `this.layers._layer` that's entirely new

core/code/ornaments.js
le-jeu: sidenote: avoid saying things changed when they are not pushed yet
for the reviewer perspective, nothing has changed, it's confusing

MysticJay: fixed

core/code/ornaments.js
MysticJay: should we then have `_layer` redirect to `layers.['Ornaments']`

core/code/ornaments.js
MysticJay: dx and dy would no longer calculate from the image's size. That means, whenever we change the (size of) the icon we would also need to adjust these parameters. We have just moved the images to iitc.app and that would cause a lot of testing. IMHO if you really need to dislocate an icon from the offered offsets you are free to simply draw a larger box and dislocate in that box.
down

core/code/ornaments.js
MysticJay: agreed, now optional parameter.

core/code/ornaments.js
MysticJay: changed

core/code/ornaments.js
le-jeu: > dx and dy would no longer calculate from the image's size.

the size of the pictures is a constant set as `window.ornaments.OVERLAY_SIZE`, that is available from any plugin.

> whenever we change the (size of) the icon we would also need to adjust these parameters.

you would need to change the value of `.OVERLAY_SIZE` and the icons of the plugin, so this offset doesn't really help
Also, all ornaments have the same fixed size due to how stock intel works, that's something out of our control.

> you are free to simply draw a larger box and dislocate in that box.

the code doesn't support different ornament size (see above)

core/code/ornaments.js
le-jeu: yes, but what about legacy `._beacons` and `._frackers`?
core will break this anyway, maybe create them in the plugins or in `.createLayer` if called with `'Beacons'` or `'Frackers'`

core/code/ornaments.js
McBen: case -1 = ,offset=[0.5,-1] =>  ` anchor = [size * offset[0], size * offset[1]];`

But this is excatly what I meant when I said: this is code that should go into the plugin and not in the core.
The plugin knows excatly how to display stuff. The ornaments class is just doing assumptation what could be required based on his own plugin.

as said in another chat: what about mouse events? animations? tooltips? (opacity?)

le-jeu: > as said in another chat: what about mouse events? animations? tooltips? (opacity?)

current ornament layer is not suited for mouse event (so tooltips), because of how leaflet and canvas works.
this events can be added to the portal itself instead of the ornaments.

But, we can discuss if using a specific canvas for each ornaments layer is more clever (performance wise) than drawing the ornaments directly on the canvas layer.

core/code/ornaments.js
MysticJay: done

core/code/ornaments.js
MysticJay: > the size of the pictures is a constant set as `window.ornaments.OVERLAY_SIZE`, that is available from any plugin.

yep, missed that. 

But still: I think it is allowed to define certain limits to what kind of images we accept and how we display them (KISS). Initially I only wanted to be able to "hide" ornaments and now the whole discussion is now about what fancy things can be done with the images/icons/ornaments shown and that we should not limit that at all, where actually I have not yet seen a request nor a reasonable use-case.

IITC is limited in so may things by the poor data we receive from stock intel, trying to improve things by adding fancy animations or similar to "no data" is imho only serving DEVs not the user. 

my2ct

le-jeu: you are the one introducing a new offset feature, which is likely to be changed because too restrictive as it is imo, and a future change implies backward compatibility or breaking plugins... more work afterwards...
 
@McBen suggestion looks kiss enough

```suggestion
            if (this.icon[ornament].offset) {
              var offset = this.icon[ornament].offset;
              anchor = [size * offset[0], size * offset[1]];
            }
```

core/code/ornaments.js
```diff
@@ -71,5 +155,116 @@ window.ornaments = {
...
+
+  onChangeHandler: function () {
+    window.ornaments.processInput();
+    window.ornaments.replaceOL();
+    window.ornaments.save();
+    window.ornaments.reload();
+  },
+
+  ornamentsOpt: function () {
+    var excludedIDs = window.ornaments.excludedOrnaments.toString();
```
le-jeu, line 239:
```suggestion
    var excludedIDs = window.ornaments.excludedOrnaments.join(',');
```

core/code/ornaments.js
le-jeu: `this.excludedOrnaments !== ['']` is _always_ true, even if `this.excludedOrnaments` is an array with a single element `''`
the input already filter out empty strings, so you just need to check whether the array is empty

```suggestion
        if (this.excludedOrnaments.length) {
```

core/code/ornaments.js
le-jeu: @MysticJay ↑

core/code/ornaments.js
```diff
@@ -71,5 +155,116 @@ window.ornaments = {
...
+    }
+    try {
+      dataStr = localStorage.getItem('knownOrnaments');
+      if (dataStr === undefined) {
+        this.initOrnaments ();
+        return;
+      }
+      this.knownOrnaments = JSON.parse(dataStr);
+    } catch (e) {
+      console.warn('ornaments: failed to load data from localStorage: '+e);
```
le-jeu, line 181:
```suggestion
      log.warn('ornaments: failed to load data from localStorage: '+e);
```

core/code/ornaments.js
```diff
@@ -13,52 +13,136 @@
...
+        var iconUrl = '//commondatastorage.googleapis.com/ingress.com/img/map_icons/marker_images/' + ornament + '.png';
+
+        if (!this.knownOrnaments[ornament]) {
+          this.knownOrnaments[ornament]=false;
+        }
+
+        if (typeof (window.ornaments.icon[ornament]) !== 'undefined') {
+          if (window.ornaments.icon[ornament].layer) {
+            if (window.ornaments.layers[window.ornaments.icon[ornament].layer] === undefined){
+              console.log ('Add missing layer: ',window.ornaments.icon[ornament].layer);
```
le-jeu:
```suggestion
              log.log ('Add missing layer: ',this.icon[ornament].layer);
```

core/code/ornaments.js
```diff
@@ -13,52 +13,136 @@
...
+          }
+        }
+
+        var exclude = false;
+        if (this.excludedOrnaments && this.excludedOrnaments !== ['']) {
+          exclude = this.excludedOrnaments.some( function(pattern) {
+            return ornament.startsWith(pattern);
+          });
+        }
+        exclude = exclude | window.ornaments.knownOrnaments[ornament];
```
le-jeu:
```suggestion
        exclude = exclude | this.knownOrnaments[ornament];
```

core/code/ornaments.js
```diff
@@ -13,52 +13,136 @@
...
+              case 0:
+                anchor = [size / 2, size / 2];
+                break;
+              case -1:
+                anchor = [size / 2, - size ];
+              }
+            }
+            if (window.ornaments.icon[ornament].opacity) {
+              opacity = window.ornaments.icon[ornament].opacity;
+            }
```
le-jeu:
```suggestion
            if (this.icon[ornament].opacity) {
              opacity = this.icon[ornament].opacity;
            }
```

core/code/ornaments.js
```diff
@@ -13,52 +13,136 @@
...
+          this.knownOrnaments[ornament]=false;
+        }
+
+        if (typeof (window.ornaments.icon[ornament]) !== 'undefined') {
+          if (window.ornaments.icon[ornament].layer) {
+            if (window.ornaments.layers[window.ornaments.icon[ornament].layer] === undefined){
+              console.log ('Add missing layer: ',window.ornaments.icon[ornament].layer);
+              window.ornaments.createLayer(window.ornaments.icon[ornament].layer);
+            }
+            layer =  window.ornaments.layers[window.ornaments.icon[ornament].layer];
```
le-jeu:
```suggestion
            layer =  this.layers[window.ornaments.icon[ornament].layer];
```

core/code/ornaments.js
```diff
@@ -13,52 +13,136 @@
...
+        var anchor = [size / 2, size / 2];
+        var iconUrl = '//commondatastorage.googleapis.com/ingress.com/img/map_icons/marker_images/' + ornament + '.png';
+
+        if (!this.knownOrnaments[ornament]) {
+          this.knownOrnaments[ornament]=false;
+        }
+
+        if (typeof (window.ornaments.icon[ornament]) !== 'undefined') {
+          if (window.ornaments.icon[ornament].layer) {
+            if (window.ornaments.layers[window.ornaments.icon[ornament].layer] === undefined){
```
le-jeu:
```suggestion
        if (ornament in this.icon) {
          if (this.icon[ornament].layer) {
            if (this.layers[this.icon[ornament].layer] === undefined){
```

le-jeu (commented):
I match the original usage of `this` in `.setup` and `.addPortal`

core/code/ornaments.js
MysticJay: outdated

core/code/ornaments.js
MysticJay: going to accept that, but should we add some limits to the value just for safety?  -1..1 will keep it close to the portal.

modos189: I don't think validation is necessary. It is enough to describe the limits of possible values. If the developer wants to specify values outside the boundaries, that's up to him

core/code/ornaments.js
MysticJay: accepted to new PR

core/code/ornaments.js
MysticJay: Accepted to new PR

core/code/ornaments.js
MysticJay: Accepted to new PR

core/code/ornaments.js
MysticJay: Accepted to new PR

McBen:
> current ornament layer is not suited for mouse event (so tooltips), because of how leaflet and canvas works. this events can be added to the portal itself instead of the ornaments.

> IITC is limited in so may things by the poor data we receive from stock intel, trying to improve things by adding fancy animations or similar to "no data" is imho only serving DEVs not the user.

okay, I see you get me wrong here. I'm not talking about adding more and more options.   
I just wonder why the core is doing the job of the plugin.

See the offset and opacity discussion above.  Nobody knows what other devs will do with the icons.
Instead of making customization easier we limit the devs to a handful options. 
Instead of adding a real plugin we add a configuration file.


example workarounds: 
a) add some kind of "getIcon" function which can be overloaded by the plugin
b) trigger a event which can be received by the plugin when a icon is about to be added (data manipulation can be tricky)

MysticJay (commented):
can not vote changes to leaflet.canvas-markers.js

PR #568, opened 2022-07-13
Ornaments & Beacons

MysticJay:
Continuing commits here as original PR was accidentally merged prematurely


core/code/ornaments.js
```diff
@@ -13,52 +13,127 @@
...
+          if (this.icon[ornament].layer) {
+            if (this.layers[this.icon[ornament].layer] === undefined){
+              log.log ('Add missing layer: ',this.icon[ornament].layer);
+              window.ornaments.createLayer(window.ornaments.icon[ornament].layer);
+            }
+            layer =  this.layers[window.ornaments.icon[ornament].layer];
+          }
+          if (window.ornaments.icon[ornament].url) {
+            iconUrl = window.ornaments.icon[ornament].url;
+            if (this.icon[ornament].offset) {
```
MysticJay, line 103: @McBen this change is very inconvenient and causing headaches.
The default for an icon's anchor=[0.5,0,5] (center)
to place the icon above the portal marker, we would need [0.5,1]
Should we not implement something, that the offset is calculated from the portal's center?
anchor = [size * offset[0]+0.5, size * offset[1]+0.5];

core/code/ornaments.js
```diff
@@ -71,5 +146,116 @@ window.ornaments = {
...
+  initOrnaments: function () {
+    this.knownOrnaments = {};
+    this.save();
+  },
+
+  load: function () {
+    var dataStr;
+    try {
+      dataStr = localStorage.getItem('excludedOrnaments');
+      if (dataStr === undefined) { return; }
```
le-jeu:
```suggestion
      if (!dataStr) { return; }
```

core/code/ornaments.js
```diff
@@ -71,5 +146,116 @@ window.ornaments = {
...
+      this.excludedOrnaments = JSON.parse(dataStr);
+    } catch (e) {
+      log.warn('ornaments: failed to load excludedOrnaments from localStorage: '+e);
+    }
+    try {
+      dataStr = localStorage.getItem('knownOrnaments');
+      if (dataStr === undefined) {
```
le-jeu:
```suggestion
      if (!dataStr) {
```

le-jeu (commented):
`localStorage.getItem` returns `null` on unknown key
empty string would also be a wrong value so you can be less restrictive on `.load` test

core/code/ornaments.js
MysticJay: implemented in a way that the anchor of an image appears to be in the center of it.
This results in the offset to be relative to the portal marker. I.e. [0, 0.5] will place the image above the portal marker

core/code/ornaments.js
MysticJay: done

core/code/ornaments.js
MysticJay: done

PR #592, opened 2022-10-17
highlight-missing-resonators: fixed typo, which broke the plugin (issue#590)

sndirsch:


sndirsch:
PR already available as #591

PR #593, opened 2022-10-18
Cleanup-code

McBen:
remove obsolete or unused (debug) code and comments

- API Changes:
`window.Render.prototype.startRenderPass(bound)` removed first (obsolete) parameter `level`

### DIFF
--- a/core/code/data_cache.js
+++ b/core/code/data_cache.js
@@ -105,9 +105,3 @@ window.DataCache.prototype.runExpire = function() {
     }
   }
 }
-
-
-window.DataCache.prototype.debug = function() {
-//NOTE: ECMAScript strings use 16 bit chars (it's in the standard), so convert for bytes/Kb
-  return 'Cache: '+Object.keys(this._cache).length+' items, '+(this._cacheCharSize*2).toLocaleString()+' bytes ('+Math.ceil(this._cacheCharSize/512).toLocaleString()+'K)';
-}
--- a/core/code/data_cache.js
+++ b/core/code/data_cache.js
@@ -24,11 +24,7 @@ window.DataCache = function() {
 
 }
 
-window.DataCache.prototype.store = function(qk,data,freshTime) {
-  // fixme? common behaviour for objects is that properties are kept in the order they're added
-  // this is handy, as it allows easy retrieval of the oldest entries for expiring
-  // however, this is not guaranteed by the standards, but all our supported browsers work this way
-
+window.DataCache.prototype.store = function (qk, data, freshTime) {
   this.remove(qk);
 
   var time = new Date().getTime();
@@ -93,11 +89,7 @@ window.DataCache.prototype.runExpire = function() {
 
   var cacheSize = Object.keys(this._cache).length;
 
-  for(var qk in this._cache) {
-
-    // fixme? our MAX_SIZE test here assumes we're processing the oldest first. this relies
-    // on looping over object properties in the order they were added. this is true in most browsers,
-    // but is not a requirement of the standards
+  for (var qk in this._cache) {
     if (cacheSize > this.REQUEST_CACHE_MAX_ITEMS || this._cacheCharSize > this.REQUEST_CACHE_MAX_CHARS || this._cache[qk].time < t) {
       this._cacheCharSize -= this._cache[qk].dataStr.length;
       delete this._cache[qk];
--- a/core/code/map.js
+++ b/core/code/map.js
@@ -251,13 +251,11 @@ window.setupMap = function () {
   // map update status handling & update map hooks
   // ensures order of calls
   map.on('movestart', function () {
-    window.mapRunsUserAction = true;
     window.requests.abort();
     window.startRefreshTimeout(-1);
   });
   map.on('moveend', function () {
-    window.mapRunsUserAction = false;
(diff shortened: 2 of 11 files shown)


core/code/data_cache.js
```diff
@@ -24,11 +24,7 @@ window.DataCache = function() {
 
 }
 
-window.DataCache.prototype.store = function(qk,data,freshTime) {
-  // fixme? common behaviour for objects is that properties are kept in the order they're added
-  // this is handy, as it allows easy retrieval of the oldest entries for expiring
-  // however, this is not guaranteed by the standards, but all our supported browsers work this way
-
```
le-jeu, line 31: I found some info for es2020 about this, but not strong result for es2015, do you have one ?

core/code/map_data_render.js
```diff
@@ -1,14 +1,13 @@
...
-
-
-window.Render = function() {
+window.Render = function () {
   this.portalMarkerScale = undefined;
 }
 
 // start a render pass. called as we start to make the batch of data requests to the servers
-window.Render.prototype.startRenderPass = function(level,bounds) {
+window.Render.prototype.startRenderPass = function (bounds) {
```
le-jeu, line 10: for api compatibility, should this check if the function was given 2 arguments to ignore the first one ?

le-jeu (commented):
looks good to me so far

core/code/data_cache.js
McBen: to tell the truth: no.
I also only have my personal experience and other blog-posts talking about "since ES5 ...". 
Like: https://dev.to/frehner/the-order-of-js-object-keys-458d
Which leads to current spec: https://tc39.es/ecma262/#sec-ordinaryownpropertykeys

On the other hand if the order is random it wouldn't break the cache.
It would just decrease it's performance. If it hits the entry or memory limit (more then 300Tiles or 5MB in 5minuts) it could 'forget' freshly loaded tile.

core/code/map_data_render.js
McBen: I don't believe this is neccessary. 
I hardly can think of a use-case for plugins here.

If you prefer to support current format keep parameter count instead of adding special cases.
```suggestion
window.Render.prototype.startRenderPass = function (_, bounds) {
```

core/code/data_cache.js
McBen: BTW: also overlay creation on map init assumes that object are sorted

core/code/map_data_render.js
le-jeu: no preference, but can you note this in the PR text so it's not forgotten ?

core/code/map_data_render.js
```diff
@@ -435,7 +435,7 @@ window.Render.prototype.createFieldEntity = function(ent) {
   fieldsFactionLayers[poly.options.team].addLayer(poly);
 }
 
-window.Render.prototype.createLinkEntity = function(ent,faked) {
+window.Render.prototype.createLinkEntity = function (ent) {
```
le-jeu, line 434: correct, it's not used since b9c8e9c6c59dbf4a846628033e6442f488792f13

core/code/map.js
```diff
@@ -289,7 +289,7 @@ window.setupMap = function () {
       pos = {center: [0, 0], zoom: 1};
       map.locate({setView: true});
     }
-    map.setView(pos.center, pos.zoom, {reset: true});
+    map.setView(pos.center, pos.zoom);
```
le-jeu: good point
curiously, it's still used in the leaflet code to force _resetView call.

core/code/map.js
```diff
@@ -251,13 +251,11 @@ window.setupMap = function () {
   // map update status handling & update map hooks
   // ensures order of calls
   map.on('movestart', function () {
-    window.mapRunsUserAction = true;
```
le-jeu, line 254: ftr, used until 571a17d84b22b1f9b2af8845315bd6473fbc1dbf for the status bar

core/code/request_handling.js
```diff
@@ -11,32 +12,30 @@ window.statusSuccessMapTiles = 0;
...
 window.statusErrorMapTiles = 0;
 
+window.requests = function () {};
 
-window.requests = function() {}
-
-//time of last refresh
+// time of last refresh
 window.requests._lastRefreshTime = 0;
-window.requests._quickRefreshPending = false;
```
le-jeu, line 19: ftr, used until 571a17d84b22b1f9b2af8845315bd6473fbc1dbf for status bar

core/code/request_handling.js
```diff
@@ -99,14 +99,3 @@ window.requests._callOnRefreshFunctions = function() {
 window.requests.addRefreshFunction = function (f) {
   window.requests._onRefreshFunctions.push(f);
 }
-
-window.requests.isLastRequest = function(action) {
```
le-jeu, line 106: ftr used for removed API `getThinnedEntitiesV4` in some deprecated plugin

le-jeu (approved):
looks good, the explicit usage of `window.` for global variables is nice :)

core/code/map.js
McBen: looks like a now undocumented leaflet feature
reverted in a8e5ee9

PR #597, opened 2022-11-13
red faction

xscreach:
Adds 'UN̴̖̈KNȮWṈ' layer to layers chooser and colors everything red when data faction is M

image

### DIFF
--- a/core/code/entity_info.js
+++ b/core/code/entity_info.js
@@ -16,10 +16,10 @@ window.teamStringToId = function(teamStr) {
   var team = TEAM_NONE;
   if(teamStr === 'ENLIGHTENED') team = TEAM_ENL;
   if(teamStr === 'RESISTANCE') team = TEAM_RES;
-  if (teamStr === 'U̶͚̓̍N̴̖̈K̠͔̍͑̂͜N̞̥͋̀̉Ȯ̶̹͕̀W̶̢͚͑̚͝Ṉ̨̟̒̅') team = TEAM_MAC;
+  if (teamStr === window.TEAM_NAME_MAC) team = TEAM_MAC;
   if(teamStr === 'E') team = TEAM_ENL;
   if(teamStr === 'R') team = TEAM_RES;
-  if (teamStr === 'M') team = TEAM_MAC;
+  if (teamStr === window.TEAM_CODE_MAC) team = TEAM_MAC;
   return team;
 }
 
--- a/core/code/entity_info.js
+++ b/core/code/entity_info.js
@@ -11,20 +11,9 @@ window.getTeam = function (details) {
 };
 
 window.teamStringToId = function (teamStr) {
-  switch (teamStr) {
-    case window.TEAM_NAME_ENL:
-    case window.TEAM_CODE_ENL:
-      return window.TEAM_ENL;
-
-    case window.TEAM_NAME_RES:
-    case window.TEAM_CODE_RES:
-      return window.TEAM_RES;
-
-    case window.TEAM_NAME_MAC:
-    case window.TEAM_CODE_MAC:
-      return window.TEAM_MAC;
-
-    default:
-      return window.TEAM_NONE;
-  }
+  var teamIndex = window.TEAM_NAMES.indexOf(teamStr);
+  if (teamIndex >= 0) return teamIndex;
+  teamIndex = window.TEAM_CODES.indexOf(teamStr);
+  if (teamIndex >= 0) return teamIndex;
+  return window.TEAM_NONE;
 };
--- a/core/code/entity_info.js
+++ b/core/code/entity_info.js
@@ -16,7 +16,7 @@ window.teamStringToId = function(teamStr) {
   var team = TEAM_NONE;
   if(teamStr === 'ENLIGHTENED') team = TEAM_ENL;
   if(teamStr === 'RESISTANCE') team = TEAM_RES;
-  if (teamStr === 'MACHINA') team = TEAM_MAC;
+  if (teamStr === 'U̶͚̓̍N̴̖̈K̠͔̍͑̂͜N̞̥͋̀̉Ȯ̶̹͕̀W̶̢͚͑̚͝Ṉ̨̟̒̅') team = TEAM_MAC;
   if(teamStr === 'E') team = TEAM_ENL;
   if(teamStr === 'R') team = TEAM_RES;
   if (teamStr === 'M') team = TEAM_MAC;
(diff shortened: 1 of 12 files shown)


xscreach:
#598

Eccenux:
Nice to see some progress, but the contrast here is bad. I mean the contrast between red text and blue background of the portal sidebar. It kind of works for portal level and portal name which are bigger, but owner name etc is not very readable to me.

Eccenux:
This PR seem to be missing some commits. 
In `window.Render.prototype.createPortalEntity` I added:
```
var team = teamStringToId(data.team);
  // the data returns unclaimed portals as level 1 - but IITC wants them treated as level 0
  if (team == TEAM_NONE) portalLevel = 0;
  if (data.team === 'N' && data.resCount > 0) {
    team = TEAM_MAC;
    portalLevel = 8;
  }
```
I guess `portalLevel = 1;` would be more accurate, but using L8 works better for filtering.

xscreach:
> This PR seem to be missing some commits. 
> In `window.Render.prototype.createPortalEntity` I added:
> ```
> var team = teamStringToId(data.team);
>   // the data returns unclaimed portals as level 1 - but IITC wants them treated as level 0
>   if (team == TEAM_NONE) portalLevel = 0;
>   if (data.team === 'N' && data.resCount > 0) {
>     team = TEAM_MAC;
>     portalLevel = 8;
>   }
> ```
> I guess `portalLevel = 1;` would be more accurate, but using L8 works better for filtering.

The portal data for this new faction contain team: M - visible with intel-x or on stock intel with luck on some portals - so no need to assume red faction on neutral portals 😉

Also, the portals have levels similar to faction portals (green and blue) - so making them all L8 (or any other level) would not be correct 🤷

Eccenux:
> The portal data for this new faction contain team: M - visible with intel-x or on stock intel with luck on some portals - so no need to assume red faction on neutral portals 😉

I don't think you can assume new faction will be added to portals on standard intel. I think they added separate intel on purpose and IITC have to support N too.

Eccenux:
In links parsing you can assume new faction:
```
window.Render.prototype.createLinkEntity = function(ent,faked) {
//...
  var data = { // TODO add other properties and check correction direction
//    type:   ent[2][0],
    team:   ent[2][1],
    oGuid:  ent[2][2],
    oLatE6: ent[2][3],
    oLngE6: ent[2][4],
    dGuid:  ent[2][5],
    dLatE6: ent[2][6],
    dLngE6: ent[2][7]
  };

  // assume non-neutral links
  if (data.team === 'N') {
	  data.team = 'M'
  }
  // ...
  var team = teamStringToId(data.team);
```
This should make links red on current intel.

Eccenux:
The change in `createPortalEntity` is not needed. Instead in `window.decodeArray.portal` just add this:
```
  if (a.length >= SUMMARY_PORTAL_DATA_LENGTH) {
    $.extend(data, summaryPortalData(a));
    // fix team
    if (data.team === 'N' && data.resCount > 0) {
      data.team = 'M';
    }
  }
```

This and change in `createLinkEntity` should be enough to mark portals as red plague aka machine aka unknow.

Still don't like the red shade in portal details but it seems to look better in IITC then on the screen.

plugins/portals-list.css
```diff
@@ -56,7 +56,7 @@
 
 #portalslist .filters {
   display: grid;
-  grid-template-columns: 1fr auto 1fr auto 1fr auto;
```
le-jeu, line 59: any reason to go from 3 to 2 columns ?

core/code/entity_info.js
```diff
@@ -15,8 +16,10 @@ window.teamStringToId = function(teamStr) {
   var team = TEAM_NONE;
   if(teamStr === 'ENLIGHTENED') team = TEAM_ENL;
   if(teamStr === 'RESISTANCE') team = TEAM_RES;
+  if (teamStr === 'U̶͚̓̍N̴̖̈K̠͔̍͑̂͜N̞̥͋̀̉Ȯ̶̹͕̀W̶̢͚͑̚͝Ṉ̨̟̒̅') team = TEAM_MAC;
```
le-jeu: I didn't see this string except for portal owner and layer display name, I'm expecting `"MACHINA"` instead

core/style.css
```diff
@@ -106,6 +106,10 @@ i.large { font-size: 6rem; }
   color: #00c5ff;
 }
 
+.mac {
+  color: #ff0028;
```
le-jeu: I recommend a lighter color, it doesn't have to match map layers (as other teams)

plugins/portals-list.css
```diff
@@ -191,6 +152,11 @@
   background-color: #017f01;
 }
 
+#portalslist table tr.mac td,
+#portalslist .filters .filterUnk {
+  background-color: #ff0028;
```
le-jeu: and darker here

le-jeu (commented):
`intel-x` and `N` => `M` may be dropped

core/style.css
AlterTobi: I have to agree. This kind of red on blue is really hard to read.

xscreach:
> `intel-x` and `N` => `M` may be dropped

Dropped

core/code/entity_info.js
xscreach: Introduced few constants so we can stop hardcoding these things easily.
But the string stays ""unknown"" for the display in layer chooser and other places like portal list/count

plugins/portals-list.css
xscreach: Visuals - 4 factions (n,e,r,m) won't fit into 3 columns 😉
Maybe we can instead add one more to have only 2 rows there, one with all 4 factions and second with the scouting...
And 2 columns only for mobile

core/style.css
xscreach: Red on blue is always pain to read, literally...
So I'm open to any suggestions.

Maybe we could just drop to white on blue...
Or if you find better red, just gimme the number 😉

And this change is supposed only for the info tab, right? Portal and link color on map is ok, isn't it?

Eccenux: Personally I've been using this via Stylus:
```
    .mac {
	  color: #ff1030;
	  text-shadow: 1px 1px 3px black, 1px 1px 3px black;
	}
```
It is dark, but looks kind of cool and I got used to it.

xscreach: i've found `#ff2020` a bit better than the original..

If we want to add shadow, we should do it for all the factions, right?

Eccenux: Actually you might want to add `text-shadow: none;` for other texts as some non-faction text would have shadow with that addition. I think the red with darker background looks kind of creepy and bloody. So as it should I think ;-)

PR #603, opened 2022-11-23
Xm amount spacer and cell search fix 

maxngrss:
Changes to agent stats:
XM spacers
Uppercase for text

Fixed search results:

on L6 cells search result are completely wrong in most cases

 for example "NR02-JULIET-12" returns NR02-JULIET-14 and etc.

Used part of code from 2015, src: https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a98809ccd51c223735553faa6e4d09ac2d44718c/plugins/regions.user.js

### DIFF
--- a/plugins/regions.js
+++ b/plugins/regions.js
@@ -131,11 +131,16 @@ window.plugin.regions.d2xy = function(n, d) {
   return xy;
 }
 
+// using some old search, because in actual version the it's broken,
+// for example "NR02-JULIET-12" returns NR02-JULIET-14 and etc.
+// source - https://github.com/IITC-CE/ingress-intel-total-conversion/blob/a98809ccd51c223735553faa6e4d09ac2d44718c/plugins/regions.user.js
 window.plugin.regions.getSearchResult = function(match) {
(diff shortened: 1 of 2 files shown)


plugins/regions.js
```diff
@@ -147,34 +152,52 @@ window.plugin.regions.getSearchResult = function(match) {
...
-    result.description = 'Regional score cells (cluster of 16 cells)';
+    result.bounds = L.latLngBounds(corners);
     result.icon = 'data:image/svg+xml;base64,'+btoa('@include_string:images/icon-cell.svg@'.replace(/orange/, 'gold'));
-    level = 4;
-  } else {
-    result.description = 'Regional score cell';
-    result.icon = 'data:image/svg+xml;base64,'+btoa('@include_string:images/icon-cell.svg@');
-    level = 6;
-
-    var xy = window.plugin.regions.d2xy(4, id2);
```
le-jeu, line 161: this line is indeed wrong, the position of the id2 cell depends on the region (cluster) position.
maybe a clever use of `rot` with the correct arguments can solve this (to avoid a loop)

the loop is good enough tho

modos189:
fixed in #623

PR #612, opened 2023-01-04
Machina tools

### LINKED ISSUE #611: Machina Tools
Add @perringaiden 's Machina Tools

MysticJay:
Machina Tools by @perringaiden imported to standard plugins.
- added new function to easier detect critical zones of machina clusters.

WARNING: this plugin is experimental. As the detected zones are cumulative they could sum up and use a lot of memory.
Use with care!


plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+    machinaTools.drawPortalExclusion(window.selectedPortal);
+  }
+};
+
+/**
+ * Indicates whether portals are displayed at the current level.  Simply using zoom level
+ * does not factor in other tools that adjust display capabilities.
+ */
+machinaTools.zoomLevelHasPortals = function () {
+  return window.getMapZoomTileParameters(window.getDataZoomForMapZoom(window.map.getZoom())).hasPortals;
```
le-jeu: short-hand used in other plugins

```suggestion
  return window.getDataZoomTileParameters().hasPortals;
```

plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+  for (var level = 0; level <= 8; ++level) {
+    if (window.LINK_RANGE_MAC[level] >= linkLength) {
+      return level;
+    }
+  }
+  return 0;
+};
+
+machinaTools.linkAdded = function (data) {
+  data.link.on('add', function () {
```
le-jeu: this occurs when disabling/re-enabling links/machina layer
you can directly use the `data.link` as machina links don't change

plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+      var level = machinaTools.guessLevelByRange(linkLength);
+      machinaTools.drawExclusion(link.oGuid, level, machinaTools.getOLatLng(link), true);
+    }
+    // }
+  });
+};
+
+function humanFileSize(size) {
+  var i = size === 0 ? 0 : Math.floor(Math.log(size) / Math.log(1024));
+  return (size / Math.pow(1024, i)).toFixed(2) * 1 + ' ' + ['B', 'kB', 'MB', 'GB', 'TB'][i];
```
le-jeu: since it's about 1024, the units are `KiB/MiB,GiB,TiB`

plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+
+machinaTools.showConflictAreaInfoDialog = function () {
+  var html = $('<div>');
+  if (machinaTools.conflictAreaLast && machinaTools.conflictAreaLast.getLayers().length > 0) {
+    var ul = $('<ul>');
+    ul.appendTo(html);
+
+    var conflictAreaLayers = machinaTools.conflictAreaLast.getLayers();
+    var conflictLayerGeometry = conflictAreaLayers[0].feature.geometry;
+    ul.append($('<li>', { html: `Cluster area size: ${humanFileSize(sizeof(conflictLayerGeometry))}` }));
```
le-jeu: about 16 x totalPoints (+ 60 for the keys/strings), sizeof doesn't count JS structures (array/object)

plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+      machinaTools.displayLayer.removeLayer(machinaTools.circleDisplayLayer);
+      $('.leaflet-control-layers-list span:contains("Machina Level Up Link Radius")')
+        .parent('label')
+        .addClass('disabled')
+        .attr('title', 'Zoom in to show those.');
+    }
+  }
+};
+
+machinaTools.showOrHideMachinaConflictArea = function () {
```
le-jeu: deadcode currently

plugins/machina-tools.js
```diff
@@ -0,0 +1,554 @@
...
+
+machinaTools.resetConflictArea = function () {
+  delete machinaTools.conflictArea;
+  delete machinaTools.conflictAreaLast;
+  machinaTools.conflictAreaLayer.clearLayers();
+};
+
+function setupLayers() {
+  // This layer is added to the layer chooser, to be toggled on/off, regardless of zoom.
+  machinaTools.displayLayer = new L.LayerGroup([], { minZoom: 15 });
```
le-jeu, line 549: this gives to the user a feedback in the layer chooser that the layer can not be display
then does this replace the line 409 ?
 ```js
$('.leaflet-control-layers-list span:contains("Machina Level Up Link Radius")').parent('label').removeClass('disabled').attr('title', '');
```

Comment (xscreach) [reply plugins/machina-tools.js:549]: The minZoon only makes the checkbox disabled, the additional line adds strikethrough to its label 😉

File: plugins/machina-tools.js (hunk 4 above)
Comment (xscreach) [reply plugins/machina-tools.js]: replaced with implementation based on https://stackoverflow.com/questions/1248302/how-to-get-the-size-of-a-javascript-object#29387295

File: plugins/machina-tools.js (hunk 6 above)
Comment (le-jeu) [reply plugins/machina-tools.js:549]: yes, but I see this a redundant feedback for the user
not important anyway ^^

File: plugins/machina-tools.js [hunk 7]
```diff
@@ -0,0 +1,654 @@
...
+};
+
+machinaTools.loadConflictAreas = function () {
+  Object.values(window.portals)
+    .filter((p) => window.TEAM_NAMES[p.options.team] === window.TEAM_NAME_MAC)
+    .forEach((portal) => machinaTools.drawPortalExclusion(portal));
+
+  Object.values(window.links)
+    .filter((l) => window.TEAM_NAMES[l.options.team] === window.TEAM_NAME_MAC)
+    .forEach((link) => machinaTools.drawLinkExclusion(link.options.data));
```
le-jeu: idea: since `machinaTools.drawPortalExclusion` takes a portal as argument, the `drawLinkExclusion` could take a link instead of link data.

 ```js
 // ...
   .forEach(machinaTools.drawPortalExclusion);
// ...
   .forEach(machinaTools.drawLinkExclusion);
```

modos189 (approved):

plugins/machina-tools.js
MysticJay: Valid comment, but in the end we are discussing about an experimental plugin. IMHO this would be different, if this was a core function or we'd have an idea to export that function to other plugins.

le-jeu: this is about code quality/consistency
while you say "experimental", the word "experimental" is nowhere to be found in the code, the experimental folder could be used, but it looks plugins there are not part of the releases

xscreach: i agree, it's more about code quality also, any PR should be a "closed" or "finished" feature meaning that we won't need to update the code unless we'll really need or want to - changes in 3rd party apis, new features, etc, but not to "finish tests later".
It might seem a bit tidious... and it sometimes really is, but it makes the code better understandable and extendable so even if we spend some more time now on this, we'll save much more time cumulatively moving forward.
Even more so with 'experimental' features because we know we'll want to update it quite soon, so why make it harder for ouselfs 😉

xscreach: aand done

PR #621, opened 2023-03-10
WIP: fix some Intent filters

le-jeu:


mobile/app/src/main/AndroidManifest.xml
```diff
@@ -43,29 +43,25 @@
             <!-- Handles the implicit intent to VIEW the intel.ingress.com/ URI -->
             <intent-filter>
                 <action android:name="android.intent.action.VIEW"/>
-                <action android:name="android.nfc.action.NDEF_DISCOVERED"/>
```
modos189, line 46: This is not required for NFC to work? I don't have the proper device to test

mobile/app/src/main/AndroidManifest.xml
```diff
@@ -43,29 +43,25 @@
...
 
                 <category android:name="android.intent.category.DEFAULT"/>
                 <category android:name="android.intent.category.BROWSABLE"/>
 
-                <data android:scheme="http"/>
-                <data android:scheme="https"/>
-                <data android:host="intel.ingress.com"/>
+                <data android:scheme="http" android:host="intel.ingress.com"/>
```
modos189, line 50: As far as I remember, it follows from the documentation that the previous version also works

mobile/app/src/main/AndroidManifest.xml
le-jeu: I'm playing around trying to figure why some apps are fine with urls, and why iitc isn't

PR #623, opened 2023-03-15
Region-Score name search fix (closes #622)

### LINKED ISSUE #622: (Plugin-Ingress scoring regions) Region-name search results show wrong sub-cell
The plugin Ingress scoring regions not only add the region grid it also alows to search for region names.  
But sometime when searching for a cell the last number is wrong.

Example:
Searching for `NR02-echo-08` resulted in `NR02-echo-02`
and vice versa: searching for `NR02-echo-02` resulted in `NR02-echo-08`

McBen:
this includes 2 things in S2geomety + region-plugin:

- a fix in the hilbercurve calculation
Rotate the hilbercurve on some faces. (like the original google-lib does)
This way the 'workaround' in region-name generation is no longer needed

- fix region-search by enhance the cell-id ..
instead of the custom calculation which resulted in the wrong values


xscreach (approved):
Looks good and seems to be working. ~~Needs cleanup 😬~~

PR #627, opened 2023-03-26
IITCm: Add plugin by URL

le-jeu:
Since loading a plugin directly from its url doesn't work as it used to be, this PR adds the capability to do it by hand from the Plugin activity.

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/PluginPreferenceActivity.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/PluginPreferenceActivity.java
@@ -1,7 +1,9 @@
 package org.exarhteam.iitc_mobile.prefs;
 
+import android.app.AlertDialog;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
+import android.content.DialogInterface;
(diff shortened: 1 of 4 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/PluginPreferenceActivity.java
```diff
@@ -156,6 +160,34 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
+                    final EditText input = new EditText(this);
+                    input.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI);
+                    builder.setView(input);
+
+                    builder.setPositiveButton(android.R.string.ok, new DialogInterface.OnClickListener() { 
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+                            final String url = input.getText().toString();
+                            final Uri uri = Uri.parse(url);
+                            if (uri != null) {
```
modos189, line 177: The app crashes if I enter an invalid URL (random sets of letters). Seems like this check should be done differently

PR #629, opened 2023-05-14
Machina Player

McBen:
Machina is now included in chat log.
This leads to some minor problems like chat format and player icons in player-tracker

This PR 
a) adds the neutral chat names
grafik
white because it's reported as "neutral" and not as MAC.

b) ignore neutral players in player-tracker plugin

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -356,7 +356,14 @@ window.chat.updateOldNewHash = function(newData, storageHash, isOlderMsgs, isAsc
   }
 };
 
-window.chat.parseMsgData = function(data) {
+function team2id(team) {
+  if (team === 'RESISTANCE') return TEAM_RES;
+  if (team === 'ENLIGHTENED') return TEAM_ENL;
+  return TEAM_NONE;
+}
+
+
+window.chat.parseMsgData = function (data) {
   var categories = data[2].plext.categories;
   var isPublic = (categories & 1) === 1;
   var isSecure = (categories & 2) === 2;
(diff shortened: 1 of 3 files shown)


plugins/player-activity-tracker.js
```diff
@@ -169,7 +169,7 @@ window.plugin.playerTracker.processNewData = function(data) {
     });
 
     // skip unusable events
-    if(!plrname || !lat || !lng || !id || skipThisMessage) return true;
+    if (!plrname || !lat || !lng || !id || skipThisMessage || json[2].plext.team === 'NEUTRAL') return true;
```
xscreach, line 172: Rather use `teamStringToId(...team) === window.TEAM_NONE`

McBen:
7219019 added a new plugin for machina tracking (show markers instead of the path)

plugins/machina_tracker.user.js
```diff
@@ -0,0 +1,263 @@
...
+          id = markup[1].latE6 + ',' + markup[1].lngE6;
+
+          name = name ? name : markup[1].name;
+          address = address ? address : markup[1].address;
+          break;
+      }
+    });
+
+    // skip unusable events
+    if (!plrname || !lat || !lng || !id || skipThisMessage || json[2].plext.team !== 'NEUTRAL') return true;
```
xscreach, line 135: Rather use `teamStringToId(...team) !== window.TEAM_NONE`

McBen:
Example of the new plugin:
grafik

core/code/chat.js
```diff
@@ -1,3 +1,5 @@
+/* global TEAM_TO_CSS */
```
xscreach, line 1: These are defined under window. We should imho rather use e.g. `window.teamToStringId` than keep defining these in files again

xscreach (changes requested):
Small clean up required, see commented lines

PR #632, opened 2023-05-21
code cleanup

McBen:


xscreach (changes requested):
Not sure what's wrong but it's not working. No plugins enabled, just pure iitc.

McBen:
99d9e10 looks broken.. not sure why

McBen:
javascript is fun they said

McBen:
fixed and added more...and 1h wasted to make the dog happy

plugins/portals-list.js
```diff
@@ -4,6 +4,8 @@
 // @version        0.4.2
 // @description    Display a sortable list of all visible portals with full details about the team, resonators, links, etc.
 
+/* global plugin,DEPLOY_RESONATOR,CAPTURE_PORTAL,COMPLETION_BONUS,DESTROY_RESONATOR,DESTROY_LINK,DESTROY_FIELD -- eslint */
```
xscreach: Nooo, don't define them like this.
Use `window.` prefix to reference constants.
Also you might want to use some `var thisPlugin = window.plugins.thisPlugin` instead of defining 'plugin' here.

core/code/portal_data.js
```diff
@@ -48,54 +48,14 @@ window.getPortalFields = function(guid) {
 window.getPortalFieldsCount = function(guid) {
   var fields = getPortalFields(guid);
   return fields.length;
-}
-
-
-// find the lat/lon for a portal, using any and all available data
-// (we have the list of portals, the cached portal details, plus links and fields as sources of portal locations)
-window.findPortalLatLng = function(guid) {
```
xscreach, line 56: Do we really want to remove this function?
It might not be used by IITC itself but since it's exposed, it could be used by some plugins.

Maybe we should just mark it deprecated and remove in some future version 🤷

core/code/portal_data.js
```diff
@@ -143,46 +103,3 @@ window.findPortalLatLng = function(guid) {
 })();
 
 
-// get the AP gains from a portal, based only on the brief summary data from portals, links and fields
-// not entirely accurate - but available for all portals on the screen
-window.getPortalApGain = function(guid) {
```
xscreach, line 148: Same here - we can't just remove exposed functions for sale of backward compatibility

core/code/data_cache.js
```diff
@@ -24,13 +24,11 @@ window.DataCache = function() {
 
 }
 
-window.DataCache.prototype.store = function (qk, data, freshTime) {
+window.DataCache.prototype.store = function (qk, data) {
```
xscreach, line 27: Again, this is exposed function allowing everyone to set their data life expectancy.
Plugins might need and relly on that. We should be removing it even if it's not used internally.

core/code/chat.js
```diff
@@ -787,9 +773,6 @@ window.chat.chooseTab = function(tab) {
 
       chat.renderAlerts(false);
       break;
-
-    default:
-      throw new Error('chat.chooser was asked to handle unknown button: ' + tt);
```
xscreach, line 793: Let's keep this here to help developers while adding new chat tabs 😉

xscreach (changes requested):
Not horrible but certainly needs more work.
Mainly around readability and removing functionality

core/code/chat.js
McBen: if people add new tab they will never end here...see the function start

core/code/portal_data.js
McBen: > Maybe we should just mark it deprecated and remove in some future version

that's something that will never happen.

if we keep it, it's on the IITC devs to keep it up-to-date instead of the one who is using it

core/code/portal_data.js
McBen: > Maybe we should just mark it deprecated and remove in some future version

that's something that will never happen.

if we keep it, it's on the IITC devs to keep it up-to-date instead of the one who is using it

plugins/portals-list.js
McBen: yeah, I don't like it...it's just to make the dog happy - this is no code!!!!!!

core/code/data_cache.js
McBen: and again: don't develop for someoneelse or an unexpected future 
(except there is a reasonable case .. a well defined interface)

things like this will keep things complicated and unmaintainable

core/code/portal_data.js
xscreach: Ok, this is what you should do:
- put the method back
- add note to changelog that this method is deprecated
- add note to documentation (if present) about the depreciation
- open pr for after release where the method will be removed

That way we atleast give time developers who use this function in their plugins to make an update.

xscreach, line 56: Ok, this is what you should do:
- put the method back
- add note to changelog that this method is deprecated
- add note to documentation (if present) about the depreciation
- open pr for after release where the method will be removed

That way we atleast give time developers who use this function in their plugins to make an update.

core/code/portal_data.js
xscreach, line 148: Ok, this is what you should do:
- put the method back
- add note to changelog that this method is deprecated
- add note to documentation (if present) about the depreciation
- open pr for after release where the method will be removed

That way we atleast give time developers who use this function in their plugins to make an update.

plugins/portals-list.js
xscreach:
```suggestion
```

Yeah, really no!

When using these in code use the `window.` prefix.
Then they won't be needed here.

plugins/portals-list.js
```diff
@@ -466,5 +468,29 @@ var setup =  function() {
     .prop("type", "text/css")
     .html('@include_string:portals-list.css@')
     .appendTo("head");
+};
 
-}
+// given counts of resonators, links and fields, calculate the available AP
+// doesn't take account AP for resonator upgrades or AP for adding mods
+window.plugin.portalslist.portalApGainMaths = function (resCount, linkCount, fieldCount) {
+  var deployAp = (8 - resCount) * DEPLOY_RESONATOR;
```
xscreach:
```suggestion
  var deployAp = (8 - resCount) * window.DEPLOY_RESONATOR;
```

Use constants with the window prefix to make the dog happy instead of defining them on top

xscreach (changes requested):
Please, try to understand the comments.
we can't _just_ remove exposed functions, etc

core/code/_deprecated.js
```diff
@@ -0,0 +1,142 @@
+/**
+ * functions that are not use by IITC itself
+ * and won't most likely not receive any updated
+ */
+/* global L,PLAYER -- eslint */
```
xscreach: Use window.PLAYER where needed and dont define it here as global

modos189 (approved):

core/code/portal_detail_display.js
MysticJay, line 1: guid is needed is line 164

core/code/portal_data.js
MysticJay: Found at least one plugin that's using it.

core/code/portal_detail_display.js
xscreach: it's actually line 13 where it is needed ;)

line 164 just uses the same guid that should be also used on that line 159

(fixed in https://github.com/IITC-CE/ingress-intel-total-conversion/pull/696)

McBen: 1year ago this parameter was unused.
now it's used by  419f6a5cdea20e914376037b9466dfcf3f892fe9

xscreach: yep, thats why im suggesting the FF only merges with review only when the PR is based on latest master, not on some history long gone 🤷 

i can understand why you (@McBen ) did the change, but maybe we could've catched that... with better process ;)

PR #635, opened 2023-05-30
fixed machina link range

xscreach:
Do not count with link amps on machina portals for link range

### DIFF
--- a/core/code/portal_info.js
+++ b/core/code/portal_info.js
@@ -41,32 +41,21 @@ window.getCurrentPortalEnergy = function(d) {
 window.getPortalRange = function(d) {
   // formula by the great gals and guys at
   // http://decodeingress.me/2012/11/18/ingress-portal-levels-and-link-range/
-
-  var resoMissing = false;
-  // currently we get a short resonator array when some are missing
-  if (d.resonators.length < 8) {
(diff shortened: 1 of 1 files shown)


core/code/portal_info.js
```diff
@@ -41,32 +41,21 @@ window.getCurrentPortalEnergy = function(d) {
...
   var range = {
-    base: d.team === 'M' ? window.LINK_RANGE_MAC[d.level + 1] : 160 * Math.pow(window.getPortalLevel(d), 4),
-    boost: getLinkAmpRangeBoost(d)
+    base: window.teamStringToId(d.team) === window.TEAM_MAC ? window.LINK_RANGE_MAC[d.level + 1] : 160 * Math.pow(window.getPortalLevel(d), 4),
+    boost: window.getLinkAmpRangeBoost(d),
   };
 
   range.range = range.boost * range.base;
-  range.isLinkable = !resoMissing;
+  range.isLinkable = d.resonators.filter((r) => !!r).length === 8;
```
MysticJay: In core we do not yet use arrow functions.
What is the need for this change?

le-jeu: the number of resonators is available in the attribute `resCount`

PR #636, opened 2023-06-01
identify moved portals

xscreach:
Highlighter marking moved portals and layer with links to moved portals

image


McBen:
obsolete, isn't it?

xscreach:
Just a note, there were few occasions where people complained about inability to link to a portal - it was moved and its own links were blocking the new ones; so imho this plugin could get occasionally handy.

PR #638, opened 2023-06-05
Changed _UNKNOWN_ to _MACHINA_

Heistergand:
Changed unicode decorated text `UN̴̖̈KNȮWṈ ` to simple text `__MACHINA__`.

Following the In-Game story, the unknown faction was given the name `__MACHINA__`, shown in scanners and stock intel as `_̶̱̍_MĄ̷̍CHḬNÄ__`  .

According to the discussion in the Telegram IITC Dev channel, it is more important that the user interface is easy to use. Better readability contributes to this.

### DIFF
--- a/core/code/map.js
+++ b/core/code/map.js
@@ -167,7 +167,7 @@ function createDefaultOverlays () {
     addLayers[window.TEAM_NAME_RES] = factionLayers[window.TEAM_RES];
   }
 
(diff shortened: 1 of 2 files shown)


modos189:
Please change to plain `__MACHINA__`

That's why:
1. In the developers telegram chat most people like the readable font
2. ~~I think it is more important: that the portal owner and the owner of the resonators have the same text~~ (up: this change does not affect the name of the owner)

xscreach:
Imho the only valid reason to make this change is that stock intell now shows the layer as weird machina too (it was "unknown" before, not sure when that changed).

modos189 (approved):

PR #649, opened 2023-07-12
fix jumpy portals on ghostlinks

xscreach:
also added timestamp (need for the fix) to the entities data (link/field.options.data.timestamp) and it's human readable display to raw data plugin

### DIFF
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -244,14 +244,15 @@ window.Render.prototype.createPlaceholderPortalEntity = function (guid, latE6, l
   timestamp = timestamp || 0;
 
   var ent = [
-    guid,       //ent[0] = guid
-    0,          //ent[1] = timestamp - zero will mean any other source of portal data will have a higher timestamp
-                //ent[2] = an array with the entity data
-    [ 'p',      //0 - a portal
-      team,     //1 - team
-      latE6,    //2 - lat
-      lngE6     //3 - lng
-    ]
+    guid, // ent[0] = guid
+    -1, // ent[1] = timestamp - zero will mean any other source of portal data will have a higher timestamp
+    // ent[2] = an array with the entity data
+    [
+      'p', // 0 - a portal
+      team, // 1 - team
+      latE6, // 2 - lat
+      lngE6, // 3 - lng
+    ],
   ];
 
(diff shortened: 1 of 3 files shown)


core/code/map_data_render.js
```diff
@@ -259,7 +261,7 @@ window.Render.prototype.createPlaceholderPortalEntity = function(guid,latE6,lngE
   if (guid in window.portals) {
     var p = window.portals[guid];
     portalMoved = latE6 !== p.options.data.latE6 || lngE6 !== p.options.data.lngE6;
-    if (team !== p.options.data.team) {
+    if (team !== p.options.data.team && p.options.timestamp < timestamp) {
```
le-jeu, line 265: It looks working, but now It looks redundant with parts of the `createPortalEntity` function. Is it possible to check this at one place ?

xscreach: I don't think so. Portal placeholders still have 0 timestamp and we check here the actual timestamp.
Placeholders with proper timestamp made portals disappear much too often (with incoming link).

So we check here the actual timestamp to delete the entity only when faction changed, the other check then works on proper portal data and only checks this timestamp for entity data changes/updates

le-jeu: oh right, I though the placeholder gets the link/field timestamp.

core/code/map_data_render.js
```diff
@@ -244,14 +244,15 @@ window.Render.prototype.createPlaceholderPortalEntity = function (guid, latE6, l
...
-    guid,       //ent[0] = guid
-    0,          //ent[1] = timestamp - zero will mean any other source of portal data will have a higher timestamp
-                //ent[2] = an array with the entity data
-    [ 'p',      //0 - a portal
-      team,     //1 - team
-      latE6,    //2 - lat
-      lngE6     //3 - lng
-    ]
+    guid, // ent[0] = guid
+    -1, // ent[1] = timestamp - zero will mean any other source of portal data will have a higher timestamp
```
le-jeu, line 248: comment doesn't match now
do we get the value `0` from intel ?

xscreach: Found it here - https://github.com/IITC-CE/ingress-intel-total-conversion/pull/324

PR #650, opened 2023-07-23
Share url, share file and delete actions for custom plugins from IITC Mobile

### LINKED ISSUE #624: Remove custom plugins from IITC Mobile
There doesn't seem to be a way to remove old/unwanted plugins in an Android app.

It would be nice to have an action to remove plugins in the plugin list or under long tap menu maybe.

xscreach:
Added context menu for user installed plugins with share download url, share plugin file and delete options

Fixes: #624

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
@@ -25,6 +25,7 @@ import androidx.core.app.ActivityCompat;
 
 import org.exarhteam.iitc_mobile.IITC_Mobile.ResponseHandler;
 import org.exarhteam.iitc_mobile.async.UpdateScript;
+import org.exarhteam.iitc_mobile.prefs.PluginInfo;
 import org.exarhteam.iitc_mobile.prefs.PluginPreferenceActivity;
 import org.json.JSONObject;
 
@@ -44,7 +45,6 @@ import java.io.StringReader;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLEncoder;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.regex.Matcher;
@@ -90,8 +90,8 @@ public class IITC_FileManager {
         }
     }
 
-    public static HashMap<String, String> getScriptInfo(final String js) {
-        final HashMap<String, String> map = new HashMap<String, String>();
+    public static PluginInfo getScriptInfo(final String js) {
+        final PluginInfo info = new PluginInfo();
         String header = "";
         // get metadata of javascript file
         if (js != null && js.contains("==UserScript==") && js.contains("==/UserScript==")) {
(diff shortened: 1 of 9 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/PluginPreference.java
```diff
@@ -31,4 +44,62 @@ protected void makeMultiline(View view) {
...
+        view.setOnCreateContextMenuListener((menu, v, menuInfo) -> {
+            if (getPluginInfo().getDownloadURL() != null) {
+                addShareUrlMenuItem(menu, R.string.menu_share_plugin_download_url, getPluginInfo().getDownloadURL());
+            }
+            if (getPluginInfo().getUpdateURL() != null) {
+                addShareUrlMenuItem(menu, R.string.menu_share_plugin_update_url, getPluginInfo().getUpdateURL());
+            }
+            MenuItem sharePluginMenu = menu.add(R.string.menu_share_plugin_file);
+            sharePluginMenu.setOnMenuItemClickListener(item -> {
+                getContext().startActivity(ShareActivity.forFile(getContext(), new File(getPluginInfo().getKey()), "application/json"));
```
le-jeu: should it be `application/javascript` instead ?

mobile/app/src/main/res/values/strings.xml
```diff
@@ -262,4 +262,12 @@
     </string>
     <string name="iitc_ce_mobile" translatable="false">IITC-CE Mobile</string>
     <string name="continue_in_english" translatable="false">Continue in English</string>
+    <string name="menu_delete_plugin">Delete</string>
+    <string name="delete_plugin_question">Delete plugin \"%s\"?</string>
+    <string name="plugin_deleted">Plugin \"%s\" deleted</string>
+    <string name="button_confirm_description">Yes</string>
+    <string name="button_cancel_description">No</string>
+    <string name="menu_share_plugin_download_url">Share download URL</string>
```
le-jeu, line 270: is there a situation where we would want to share the update URL while having the download URL ?

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_FileManager.java
```diff
@@ -90,38 +90,34 @@ public static void copyStream(final InputStream inStream, final OutputStream out
...
-        map.put("id", "unknown");
-        map.put("version", "not found");
-        map.put("name", "unknown");
-        map.put("description", "");
-        map.put("category", "Misc");
+        info.setId("unknown");
+        info.setVersion("not found");
+        info.setName("unknown");
+        info.setDescription("");
+        info.setCategory("Misc");
```
le-jeu: It looks like this is the only place the explicit setters are used, I feel this can be set in the PluginInfo constructor

le-jeu (commented):
It looks good and works great. Maybe few marginal changes.

mobile/app/src/main/res/values/strings.xml
xscreach: When the URLs are not the same - will add contituons for that 👍

le-jeu: when it defers, the update URL is usually a meta.js file that's not that useful to share.

xscreach: Well... it's there for people to copy/share. Maybe it will be helpful for someone having problems with update because the URL differ and because the update URL is wrong.

Better to have this and not need it rather than need it and not have it.

PR #653, opened 2023-07-28
Alphabetical sorting of all plugin links under the portal details panel

### LINKED ISSUE #631: Alphabetical sorting of all plugin links under the portal details panel
I propose that IITC automatically sort all the links/buttons under the portal details panel in alphabetical order. This would be a helpful improvement for usability.

It would make it easier to find the plugin buttons you need. Currently, their order is not fixed, so it can be difficult to find the one you're looking for, especially if you have a lot of plugins installed. Sorting that list alphabetically would make them much easier to find.

It would prevent the links from getting reshuffled if you add or remove plugins. Currently, if you add or remove a plugin, the order of the links can change. This can be frustrating, especially if you've got used to their order/position. Sorting alphabetically would prevent this from happening.

I imagine the code to sort the links alphabetically should be relatively short and simple.

xscreach:
Fixes #631

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -205,6 +205,29 @@ function prepPluginsToLoad () {
   };
 }
 
+function setupToolboxSort() {
+  var toolboxElement = $('#toolbox')[0];
+
+  function sortToolbox() {
+    var children = Array.prototype.slice.call(toolboxElement.children);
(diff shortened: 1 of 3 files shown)


modos189 (approved):

MysticJay:
see comment in #664

PR #654, opened 2023-07-30
Fix: UserAgent is wrong on google.* if not .com

le-jeu:
Hardcode a white list for all `google.*` domain names
Note: this is more generic than the real exhaustive list from https://www.google.com/supported_domains

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
@@ -118,6 +118,7 @@ public class IITC_Mobile extends AppCompatActivity
     private String debugInputStore = "";
     private Map<String, String> mAllowedHostnames = new HashMap<>();
     private Set<String> mInternalHostnames = new HashSet<>();
+    private final Pattern mGoogleHostnamePattern = Pattern.compile("(^|\\.)google(\\.com|\\.co)?\\.\\w+$");
 
     private String mIITCDefaultUA;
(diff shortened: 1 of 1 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
```diff
@@ -118,6 +118,7 @@ public class IITC_Mobile extends AppCompatActivity
     private String debugInputStore = "";
     private Map<String, String> mAllowedHostnames = new HashMap<>();
     private Set<String> mInternalHostnames = new HashSet<>();
+    private final Pattern mGoogleHostnamePattern = Pattern.compile("google\\.(com\\.|co\\.).\\w+$");
```
xscreach:
```suggestion
    private final Pattern mGoogleHostnamePattern = Pattern.compile("google\\.(com\\.|co\\.)?\\w+$");
```
If the match should work as mentioned in comment (for google.com?.*), The pattern should look like this, right? 🤔

xscreach (approved):

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
le-jeu: fixup half fixed this, (`?.` instead of `.`), it's `?` now
I kept the `(com\.|co\.)?` for readability (instead of `(com?\.)?`, should I use `(com?\.)?` instead ?

xscreach: Imho it's good like this.
I'd maybe prefer having the dot on the other side like this: `...google(\\.com|\\.co)?\\.\\w+`
For me it reads a bit better, but it's really just a cosmetics 🤷

modos189 (approved):

PR #655, opened 2023-08-01
[Release] 0.36.0 (bump versions)

modos189:


plugins/player-activity-tracker.js
```diff
@@ -1,14 +1,14 @@
...
-// @version        0.12.2
+// @version        0.12.3
 // @description    Draw trails for the path a user took onto the map based on status messages in COMMs. Uses up to three hours of data. Does not request chat data on its own, even if that would be useful.
 
 /* exported setup, changelog --eslint */
 
 var changelog = [
   {
-    version: '0.12.2',
+    version: '0.12.3',
```
xscreach: Imho there's no need to change version here as only the changelog entry was added, nothing more

plugins/bookmarks.js
```diff
@@ -1,10 +1,18 @@
...
+// @version        0.5.0
 // @description    Save your favorite Maps and Portals and move the intel map with a click. Works with sync. Supports Multi-Project-Extension
 
 /* global L -- eslint */
+/* exported setup, changelog --eslint */
+
+var changelog = [
+  {
+    version: '0.5.0',
+    changes: ['Range display is now consistent with autodraw. Better precision for long links'],
```
xscreach: Here we only extracted functionality for use elsewhere, but the plugin functionality didn't change.

Maybe we should only bump minor version and write smth like 
```suggestion
    changes: ['(dev) extracted "formatDistance" function for global use'],
```

xscreach (commented):
Maybe we should also note in core changelog the addition of `window.formatDistance` function.

modos189:
@xscreach Thanks

PR #659, opened 2023-08-03
Machina marker icon

MysticJay:
new images in pNG, 2xPNG and SVG


xscreach (commented):
Looks good.
Just keep only files that are actually used (remove the 2 PNG)

xscreach:
Is the PNG file still used somewhere?
If not we should remove it

xscreach (approved):

plugins/images/marker-machina.svg
```diff
@@ -0,0 +1 @@
...8 24.672-3.3628 0.27662 2.2506 5.5933 4.4496-5e-6 -0.12716 10.424-5.9751-10.043 1.3984-1.3345-9.2163 4.4492-0.50864-5.848-2.7968-4.1313 5.5301 0.27698 2.2546-24.78 3.3705z"/><path d="m67.953 140.74 19.824 15.068-1.3958 1.787 0.80874 7.1014 0.08991-0.0899 11.596 3.1461-8.0905-6.1128 5.5733-7.4605 3.5057 2.7864-2.1575-6.1128-6.8316 0.98908-1.3984 1.7901-19.906-15.139z"/><path d="m67.395 139.84 3.3629 24.672-2.2506 0.27661-4.4496 5.5933 0.12715 1e-5 5.9751 10.424-1.3984-10.043 9.2163-1.3345 0.50864 4.4492 2.7968-5.848-5.5301-4.1313-2.2546 0.27697-3.3705-24.78z"/></g><g stroke-width=".265"><path d="m103.91 148.68-15.068 19.824-1.787-1.3958-7.1014 0.80874 0.08992 0.0899-3.1461 11.596 6.1128-8.0905 7.4605 5.5733-2.7864 3.5057 6.1128-2.1575-0.98909-6.8316-1.7901-1.3984 15.139-19.906z"/><path d="m104.82 148.13-24.672 3.3628-0.27662-2.2506-5.5933-4.4496 5e-6 0.12716-10.424 5.9751 10.043-1.3984 1.3345 9.2163-4.4492 0.50864 5.848 2.7968 4.1313-5.5301-0.27698-2.2546 24.78-3.3705z"/><path d="m105.85 148.37-19.824-15.068 1.3958-1.787-0.80874-7.1014-0.08991 0.0899-11.596-3.1461 8.0905 6.1128-5.5733 7.4605-3.5057-2.7864 2.1575 6.1128 6.8316-0.98908 1.3984-1.7901 19.906 15.139z"/></g></g></g></svg>
```
xscreach, line 1: Not sure why, but the SVG doesn't work on mobile and default png is displayed instead 😬





Edit: it's due to the missing `iconRetinaUrl` - see other coments

xscreach (changes requested):

plugins/machina-tracker.js
```diff
@@ -30,13 +30,16 @@ machinaTracker.events = [];
...
-    iconAnchor: [12, 32],
-  });
+  var iconImage = '@include_img:images/marker-machina.svg@';
+
+  machinaTracker.icon = new (L.Icon.Default.extend({
+    options: {
+      iconUrl: iconImage,
+      iconSize: [26, 32],
+      iconAnchor: [12, 32],
+      shadowAnchor: [12, 41],
```
xscreach, line 41: all the defaults are already defined - these values make machina marker smaller than markers from classic player tracker

plugins/machina-tracker.js
```diff
@@ -30,13 +30,16 @@ machinaTracker.events = [];
...
-  machinaTracker.icon = L.icon({
-    iconUrl: iconImage,
-    iconSize: [26, 32],
-    iconAnchor: [12, 32],
-  });
+  var iconImage = '@include_img:images/marker-machina.svg@';
+
+  machinaTracker.icon = new (L.Icon.Default.extend({
+    options: {
+      iconUrl: iconImage,
```
xscreach, line 37: iconRetinaUrl needs to be specified too otherwise theres default marker displayed on 'retina' revices

```
options: {
      iconUrl: iconImage,
      iconRetinaUrl: iconImage,
    },
```

plugins/images/marker-machina.svg
MysticJay: fixed

plugins/machina-tracker.js
MysticJay: fixed

plugins/machina-tracker.js
```diff
@@ -30,13 +30,17 @@ machinaTracker.events = [];
...
+      iconRetinaUrl: iconImage,
+      iconSize: [26, 32],
+      iconAnchor: [12, 32],
```
xscreach, line 40: Any comment on why you want to specify different sizes than defaults here?

It makes the icon smaller than the Enl/Res tracker.
And the values are now same as default, then remove them and keep the default in use - there's no need to specify it again and will result in more work later when we'll want to change the size of all these tracker markers.

xscreach (changes requested):

plugins/machina-tracker.js
McBen: > It makes the icon smaller than the Enl/Res tracker.

That was the intention behind these values. Machina was meant to be smaller and inconspicuous.

`same as default` , what defaults? The image size? 
Anchor + Shadow are depends on Image content and can hardly match any default.
So it's also good to have a fixed size here 'cause they are all related.

PR #664, opened 2023-08-07
Revert "Alphabetical sorting of all plugin links under the portal details panel"

modos189:
This reverts commit 747f876e (PR https://github.com/IITC-CE/ingress-intel-total-conversion/pull/653)

There was a conflict with some of the plugins. Let's improve the implementation, add a possibility to disable and bring back this feature

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -205,29 +205,6 @@ function prepPluginsToLoad () {
   };
 }
 
-function setupToolboxSort() {
-  var toolboxElement = $('#toolbox')[0];
-
-  function sortToolbox() {
-    var children = Array.prototype.slice.call(toolboxElement.children);
-    var sortedChildren = children.slice().sort(function (x, y) {
-      return x.innerText.localeCompare(y.innerText);
-    });
-    if (
-      sortedChildren.some(function (item, index) {
-        return item !== children[index];
-      })
-    ) {
-      sortedChildren.forEach(function (child) {
-        toolboxElement.removeChild(child);
-        toolboxElement.appendChild(child);
-      });
-    }
-  }
-  sortToolbox();
-  window.observeDOMChildren(toolboxElement, sortToolbox);
-}
-
 function boot() {
   log.log('loading done, booting. Built: '+'@build_date@');
   if (window.deviceID) {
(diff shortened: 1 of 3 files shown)


xscreach:
Provide more information about the conflict. Which plugin, what didn't worked, etc.

MysticJay:
Plugin is sorting in a different way, causing IITC and the plugin reordering the toolbox permanently. The user is no longer able to interact with IITC's UI.

MysticJay:
The plugin in question uses MutantObserver to rearrange (not only sort) the links in the toolbox. IITC does the same in the same way, so every time something changes that toolbox DIV, both routines get executed, causing the other routine to execute again, as there is a new change. 

The assumption that the links are always unsorted is wrong.

Yet I have not found a way to identify that an other plugin has intentionally arranged the links to allow IITC to ignore the change.

IMHO MutantObserver is a nice idea but counter productive here. There is no constant flow of new plugins loading after initial setup. OTOH there are tools to rearrange the links in a way that the user prefers. It would be sufficient to arrange the links once and done. 

Thus I see three options:
1. remove the MutantObserver and run the sort once all plugins are loaded.
2. expose the sort routine so that it can be removed from MutantObserver or replaced/hijacked
3. the built-in routine needs a switch that allows to skip the sorting.

MysticJay:
he plugin in question uses MutantObserver to rearrange (not only sort) the links in the toolbox. IITC does the same in the same way, so every time something changes that toolbox DIV, both routines get executed, causing the other routine to execute again, as there is a new change. 

The assumption that the links are always unsorted is wrong.

Yet I have not found a way to identify that an other plugin has intentionally arranged the links to allow IITC to ignore the change.

IMHO MutantObserver is a nice idea but counter productive here. There is no constant flow of new plugins loading after initial setup. OTOH there are tools to rearrange the links in a way that the user prefers. It would be sufficient to arrange the links once and done. 

Thus I see three options:
1. remove the MutantObserver and run the sort once all plugins are loaded.
2. expose the sort routine so that it can be removed from MutantObserver or replaced/hijacked
3. the built-in routine needs a switch that allows to skip the sorting.

PR #666, opened 2023-08-09
Toolbox sort

MysticJay:


core/code/boot.js
```diff
@@ -205,6 +205,29 @@ function prepPluginsToLoad () {
...
+  ) {
+    sortedChildren.forEach(function (child) {
+      toolboxElement.removeChild(child);
+      toolboxElement.appendChild(child);
+    });
+  }
+}
+function setupToolboxSort() {
+  toolboxElement = $('#toolbox')[0];
+  window.setTimeout( ()=> window.sortToolbox(), 1000);
```
xscreach: We don't use arrow functions in core.

```suggestion
  window.setTimeout(window.sortToolbox, 1000);
```

core/code/boot.js
```diff
@@ -205,6 +205,29 @@ function prepPluginsToLoad () {
...
+    })
+  ) {
+    sortedChildren.forEach(function (child) {
+      toolboxElement.removeChild(child);
+      toolboxElement.appendChild(child);
+    });
+  }
+}
+function setupToolboxSort() {
+  toolboxElement = $('#toolbox')[0];
```
xscreach: This and the variable itself should be moved into the sortToolbox function

core/code/utils_misc.js
```diff
@@ -427,3 +427,22 @@ if (!Element.prototype.closest) {
     return null;
   };
 }
+
+var MutObserver = window.MutationObserver || window.WebKitMutationObserver;
```
xscreach, line 431: This is not used so it should be here.

xscreach (changes requested):
Works for all mobile, button and tampermonkey (even with plugins loaded after iitc main script)

Needs code cleanup

Edit: doesn't always work with tampermonkey - large scripts might load after the sort is already done

core/code/boot.js
le-jeu:
```suggestion
```

core/code/boot.js
```diff
@@ -205,6 +205,29 @@ function prepPluginsToLoad () {
   };
 }
 
+var toolboxElement;
+window.sortToolbox = function () {
```
le-jeu:
```suggestion
window.sortToolbox = function () {
  var toolboxElement = document.getElementById('toolbox') ;
```

le-jeu:
since this is about improving toolbox readability, you can also display the link by column :
image

With  `column-count: 2` (and few tweak on margin/padding):
image

Same but with 3 columns:
image

It's possible to use a grid layout as well, giving the links another direction using 
`display: grid; grid-template-column: repeat(<number of column>, 1fr)` for instance
image

core/code/utils_misc.js
MysticJay: I just left it, as it is a nice extension to utils_misc. But maybe it is worth a different PR
@xscreach

xscreach: If it's not used then just remove it.
It can be simply added back when needed.

MysticJay:
> since this is about improving toolbox readability, you can also display the link by column :
> 
> With `column-count: 2` (and few tweak on margin/padding): 
> 
> Same but with 3 columns: 
> 
> It's possible to use a grid layout as well, giving the links another direction using `display: grid; grid-template-column: repeat(<number of column>, 1fr)` for instance 

looks nice and easy to implement, but I'd rather put that to a new issue/PR.
Many users have complained in the past about  the toolbox becoming too large. Putting the links into two columns adds even more size. 

My proposal here would be a mall checkbox that switches between the old (now sorted) and the 2-columnsview (vertically sorted). Three colums seem to overdo, but maybe the checkbox could also be a 3-state-toggle between 1, 2 and 3.

core/code/boot.js
```diff
@@ -205,6 +205,28 @@ function prepPluginsToLoad () {
...
+    })
+  ) {
+    sortedChildren.forEach(function (child) {
+      toolboxElement.removeChild(child);
+      toolboxElement.appendChild(child);
+    });
+  }
+}
+function setupToolboxSort() {
+  window.setTimeout(window.sortToolbox, 1000);
```
xscreach, line 227: The timeout won't be enough to have all the links always sorted - there'll always be possibility for some plugins to load later and add unsorted links.

Imho we should keep the observer and allow plugins to override the sorting function - that way we ensure it's always sorted (100%) and we also allow extensions by plugins 🤷

MysticJay: If a plugin adds to the toolbox dynamically, it should also call sortToolbxo() after that.
If we find slow plugins that have not finished initializing, we should look into this again.

PR #667, opened 2023-08-11
Update .prettierrc.json

MysticJay:
GIT converts LineEndings to the locally used standard while downloading.
In windows this is CRLF.
Editors that suport linters/prettier complain about that for each line as the default for Prettier is LF.
by adding `"endOfLine": "auto"` to prettierrc.json, prettier will accept the OS's default.

Please check if the setting causes any problems in your environment (it should not, but maybe you tweaked all your windows tools to work with LF too?)

(Please avoid discussions on IDEs/Windows/Linux/LF/CRFL.)

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -3,5 +3,6 @@
   "trailingComma": "es5",
   "tabWidth": 2,
   "semi": true,
-  "singleQuote": true
+  "singleQuote": true,
+  "endOfLine": "auto"
 }
--- a/EOL-Test-File.js
+++ b/EOL-Test-File.js
@@ -1,4 +1,0 @@
-// This is just a test file for EoL
-function test() {
-
-}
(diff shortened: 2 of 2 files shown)


xscreach:
Shouldn't we rather set gitattributes for js files to be always checked out with LF?
We do similar for other files so why not for js aswell?

MysticJay:
defaulting to LF causes many windows tools to fail.
I have been trying a lot to overcome this problem, but nothing worked out.
This setting does not harm anything, (which I just want you to confirm).
GIT is already downloading in the best way (auto) for your environment.

modos189:
I checked that I do indeed have the EOL-Test-File.js file loading as LF. Please remove this file for merge

modos189 (approved):

le-jeu (approved):
Looks good to me

PR #669, opened 2023-09-01
Update OSM tile server

le-jeu:
+ Expose configuration to allow customization by other plugins
+ Add CyclOSM tiles

### DIFF
--- a/plugins/basemap-openstreetmap.js
+++ b/plugins/basemap-openstreetmap.js
@@ -4,10 +4,18 @@
 // @version        0.1.2
 // @description    Add the native OpenStreetMap.org map tiles as an optional layer.
 
-/* exported setup --eslint */
+/* exported setup, changelog --eslint */
 /* global L, layerChooser */
+
 // use own namespace for plugin
 var mapOpenStreetMap = {};
+
+var changelog = [
+  {
+    version: '0.1.3',
+    changes: ['Update OSM tile provider', 'Add CyclOSM tiles'],
+  },
+];
 
 mapOpenStreetMap.addLayer = function () {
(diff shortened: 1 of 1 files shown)


plugins/basemap-openstreetmap.js
```diff
@@ -4,32 +4,53 @@
...
+  },
+  {
+    name: 'Humanitarian',
+    url: 'https://{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png',
+    options: osmOpt,
+  },
+  {
+    name: 'CyclOSM',
+    url: 'https://{s}.tile-cyclosm.openstreetmap.fr/cyclosm/{z}/{x}/{y}.png',
+    options: osmOpt,
```
xscreach:
```suggestion
    options: Object.assign({}, osmOpt),
```

... otherwise it's not possible to override only one of the layer settings as they share the same "options" instance

xscreach (changes requested):
Looks good except one little thing 👍

plugins/basemap-openstreetmap.js
le-jeu: it's possible by overriding the option object the same way

xscreach: That isn't very straightforward or user friendly, is it?

le-jeu: I assume this is dev friendly enough
also, `Object.assign` is not ES5

xscreach: Es6 is supported by all browsers since 2017 (some even earlier).

Also, Object.assign is already used in our code (utils_file.js#135)

le-jeu: grmpf, then coding style need to be updated

PR #675, opened 2023-10-12
Fix: player tracker reports wrong player team and/or wrong location

le-jeu:
Following recent changes in the COMM data, the player activity tracker no longer ignore destroyed link/field messages, therefor players may be displayed at wrong locations. 

This is due to comm message having reporting inconsistent data.
Each log message has a `team` field, and each player tag in a message has a `team` field too.
Usually, the message `team` is the same as the player `team`, but it differs on destroyed link message (at least). It looks like the link `team` is used instead.

The PR uses the player `team` instead of the message `team` to determine the faction of a player.

```json
[
    "f58d75500e1f4ae099912d6664f26066.d",
    1697053409760,
    {
      "plext": {
        "text": "Agent jorodfr destroyed the Enlightened Link Colonne Publi (2 Avenue de la Porte Chaumont, 75019 Paris, France) to Stade Marcel Cerdan (170 Avenue Jean Jaurès, 93500 Pantin, France)",
        "team": "ENLIGHTENED",
        "markup": [
          [
            "TEXT",
            {
              "plain": "Agent "
            }
          ],
          [
            "PLAYER",
            {
              "plain": "jorodfr",
              "team": "RESISTANCE"
            }
          ],
         ...
 ```

### DIFF
--- a/plugins/player-activity-tracker.js
+++ b/plugins/player-activity-tracker.js
@@ -153,6 +153,8 @@ window.plugin.playerTracker.processNewData = function(data) {
         // portal the player is at, so ignore it.
         if(markup[1].plain.indexOf('destroyed the Link') !== -1
           || markup[1].plain.indexOf('destroyed a Control Field') !== -1
+          // COMM messages changed a bit, keep old rules ↑ in case of rollback
+          || markup[1].plain.indexOf('destroyed the') !== -1
           || markup[1].plain.indexOf('Your Link') !== -1) {
           skipThisMessage = true;
           return false;
--- a/plugins/player-activity-tracker.js
+++ b/plugins/player-activity-tracker.js
@@ -186,7 +186,7 @@ window.plugin.playerTracker.processNewData = function(data) {
     });
 
     // skip unusable events
-    if (!plrname || !lat || !lng || !id || skipThisMessage || ![window.TEAM_RES, window.TEAM_ENL].includes(plrteam)) {
+    if (!plrname || !lat || !lng || !id || skipThisMessage || ![window.TEAM_RES, window.TEAM_ENL].includes(window.teamStringToId(plrteam))) {
       return true;
     }
 
--- a/plugins/player-activity-tracker.js
+++ b/plugins/player-activity-tracker.js
@@ -153,23 +153,25 @@ window.plugin.playerTracker.processNewData = function(data) {
     var skipThisMessage = false;
     $.each(json[2].plext.markup, function(ind, markup) {
       switch(markup[0]) {
-      case 'TEXT':
-        // Destroy link and field messages depend on where the link or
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugins/player-activity-tracker.js [hunk 1]
```diff
@@ -153,13 +159,16 @@ window.plugin.playerTracker.processNewData = function(data) {
         // portal the player is at, so ignore it.
         if(markup[1].plain.indexOf('destroyed the Link') !== -1
           || markup[1].plain.indexOf('destroyed a Control Field') !== -1
+          // COMM messages changed a bit, keep old rules ↑ in case of rollback
+          || markup[1].plain.indexOf('destroyed the') !== -1
```
xscreach: That won't ignore fields as they have "a" instead of "the" there
This will ignore both.
```suggestion
          || markup[1].plain.indexOf('destroyed ') !== -1
```

Haven't tested, not sure if it doesn't causes some problems 🤷

xscreach (approved):
Looks good, suggested small change

plugins/player-activity-tracker.js
le-jeu: `destroyed a resonator` is reliable

le-jeu: and for field, it's also a `destroyed the`

McBen:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/1fa34fe00dfe6adb392770b06e43bac1aef923a0/plugins/player-activity-tracker.js#L189

Shouldn't this replaced too?

McBen:
NOTE if the "destroy" messages are the only ones with wrong message team flag then this missleading data is ignored anyway :)

but I absolutly agree to use PLAYER.team instead of MSG.team here.

plugins/player-activity-tracker.js
```diff
@@ -177,7 +186,7 @@ window.plugin.playerTracker.processNewData = function(data) {
     });
 
     // skip unusable events
-    if (!plrname || !lat || !lng || !id || skipThisMessage || ![window.TEAM_RES, window.TEAM_ENL].includes(window.teamStringToId(json[2].plext.team))) {
+    if (!plrname || !lat || !lng || !id || skipThisMessage || ![window.TEAM_RES, window.TEAM_ENL].includes(plrteam)) {
```
modos189:
```suggestion
    if (!plrname || !lat || !lng || !id || skipThisMessage || ![window.TEAM_RES, window.TEAM_ENL].includes(window.teamStringToId(plrteam))) {
```

The `plrteam` contains the faction name, but TEAM_RES and TEAM_ENL contain numbers.
Without this fix, no trackers appear on the map

modos189 (approved):

McBen (approved):

PR #676, opened 2023-10-12
chat: convert new messages to old style

McBen:
convert from
Screenshot 2023-10-12 205721
to
Screenshot 2023-10-12 205531

and from
Screenshot 2023-10-12 205642
to
Screenshot 2023-10-12 205840

### DIFF
--- a/core/code/chat.js
+++ b/core/code/chat.js
@@ -432,6 +432,15 @@ window.chat.writeDataToHash = function(newData, storageHash, isPublicChannel, is
 //
 
 window.chat.renderText = function (text) {
+  if (text.team) {
+    var teamId = window.teamStringToId(text.team);
+    if (teamId === TEAM_NONE) teamId = TEAM_MAC;
+    var spanClass = window.TEAM_TO_CSS[teamId];
+    return $('<div>')
+      .append($('<span>', { class: spanClass, text: text.plain }))
+      .html();
+  }
+
   return $('<div>').text(text.plain).html().autoLink();
 };
 
@@ -501,7 +510,10 @@ window.chat.renderMarkupEntity = function (ent) {
 
 window.chat.renderMarkup = function (markup) {
   var msg = '';
-  markup.forEach(function(ent, ind) {
+
+  transformMessage(markup);
+
+  markup.forEach(function (ent, ind) {
     switch (ent[0]) {
     case 'SENDER':
     case 'SECURE':
@@ -521,7 +533,31 @@ window.chat.renderMarkup = function (markup) {
   return msg;
 };
 
-window.chat.renderTimeCell = function(time, classNames) {
+function transformMessage(markup) {
+  // Collapse <faction> + "Link"/"Field". example: "Agent <player> destroyed the <faction> Link ..."
+  if (markup.length > 4) {
+    if (markup[3][0] === 'FACTION' && markup[4][0] === 'TEXT' && (markup[4][1].plain === ' Link ' || markup[4][1].plain === ' Control Field @')) {
+      markup[4][1].team = markup[3][1].team;
+      markup.splice(3, 1);
+    }
+  }
+
+  // skip "Agent <player>"  at beginning
+  if (markup.length > 1) {
+    if (markup[0][0] === 'TEXT' && markup[0][1].plain === 'Agent ' && markup[1][0] === 'PLAYER') {
+      markup.splice(0, 2);
+    }
+  }
+
+  // skip "<faction> agent <player>" at beginning
+  if (markup.length > 2) {
+    if (markup[0][0] === 'FACTION' && markup[1][0] === 'TEXT' && markup[1][1].plain === ' agent ' && markup[2][0] === 'PLAYER') {
+      markup.splice(0, 3);
+    }
+  }
+}
+
+window.chat.renderTimeCell = function (time, classNames) {
   var ta = unixTimeToHHmm(time);
   var tb = unixTimeToDateTimeString(time, true);
   // add <small> tags around the milliseconds


core/code/chat.js
```diff
@@ -521,7 +533,31 @@ window.chat.renderMarkup = function (markup) {
   return msg;
 };
 
-window.chat.renderTimeCell = function(time, classNames) {
+function transformMessage(markup) {
```
xscreach, line 536: If we want to transform data, we should do it as close to server as possible - entity_something.js

Imho we should update the rendering instead.

core/code/chat.js
```diff
@@ -432,6 +432,15 @@ window.chat.writeDataToHash = function(newData, storageHash, isPublicChannel, is
...
+  if (text.team) {
+    var teamId = window.teamStringToId(text.team);
+    if (teamId === TEAM_NONE) teamId = TEAM_MAC;
+    var spanClass = window.TEAM_TO_CSS[teamId];
+    return $('<div>')
+      .append($('<span>', { class: spanClass, text: text.plain }))
+      .html();
+  }
+
   return $('<div>').text(text.plain).html().autoLink();
```
xscreach, line 444: Do we really need two returns here?
It should be enough to fill the spanClass conditionally and always return the same construct, yes?

misuo:
To be consistent with the first from/to example, the second from/to example with the "<agent> created a Control Field ...", the text "Control Field" should be colored to the faction color.

McBen:
> To be consistent with the first from/to example, the second from/to example with the " created a Control Field ...", the text "Control Field" should be colored to the faction color.

no, `destroy` messages contains additional informations (killed a MAC,RES or ENL link/field)
`create` messages would only gain more distracting colors.

core/code/chat.js
McBen: yes, except if we want to add useless `<span>` to all text parts.
or remove the jquery codes and use plain text.

atleast I don't see a better and clearer solution here.

(note: the `<div>`  tag here is removed the `.html()` call )

core/code/chat.js
McBen: chat data is completly processed in chat.js
`renderMarkup` is used to process new received data and store it in a local cache.

so here I changed the server data right after it is received and as close as possible to the point were it is rendered. (not displayed)

Other possible locations for a "transformMessage" call would be: 'renderMsgRow', 'parseMsgData'

le-jeu: > Other possible locations for a "transformMessage" call would be: 'renderMsgRow', 'parseMsgData'

I prefer in `renderMsgRow`, exposing the function as `window.chat.transformMarkup`, something like
```js
var markup = window.chat.transformMarkup(data.markup);
var msg = window.chat.renderMarkup(markup);
...
// or
var markup = window.chat.transformMarkup(data); // give parsed data as well
var msg = window.chat.renderMarkup(markup);
```
both without modifying the original `markup` data (no strong opinion about this, but I find it cleaner)

core/code/chat.js
xscreach: And why add span into empty div? Just for coloring? That's also useless span, isn't it? 🤷

McBen: yes, just for coloring. how do you apply colors to text parts? 😕

core/code/chat.js
McBen: Wouldn't it be cool if IITC would do the whole text parsing? 
And plugins could rely on always receiving the same format?

Let say nia changes the text from "destroyed" to "eliminated". Wouldn't it be cool if it just requires an IITC update instead of patching 10 plugins?  --- okay, if we can agree that IITC updated shouldn't take a year.

---
'exposing' .. I prefer private data, private functions and strict interfaces. Exposed functions and data entice to manipulate them. Making code harder to change or leads to unpredicted results.

core/code/chat.js
McBen: what I tried to say:
`$("<div>").text("Hello").html()` = `"Hello"`
`$("<div>").text(" destroyed ").html()` + `$("<div class='enl'>").text("link").html()` = `" destroyed link"`

core/code/chat.js
le-jeu: plugins based on chat data use raw data directly via the chat hooks, they don't use the one parsed by the core plugin as far as I know
also, it's about rendering, the markup is not the place to check and rewrite what the data is about, it's better to leave it in a specific places.
Then, you may have a plugin that see : this message is a "destroyed field" message, about this "agent" on this "portal", and then create a markup for it from scratch.

EDIT: about extracting data, that was one of my objective with the comm-filter plugin for instance
https://github.com/le-jeu/iitc-plugins/blob/d2aec89da0a6bd949e5938e45825725d6cca67e5/plugins/comm-filter-tab.js#L307
https://github.com/le-jeu/iitc-plugins/blob/d2aec89da0a6bd949e5938e45825725d6cca67e5/plugins/comm-filter-tab.js#L414

McBen: > raw data 

not "result" ? atleast all stock plugins use "result" instead of "raw"

> EDIT: about extracting data, that was one of my objective with the comm-filter plugin for instance

isn't this excatly what I tried to say? Why doing this in every plugin when IITC can standarized it?

McBen: yeah, sure, the 'transformMessage' function doesn't fit well in the chat.js code style.

McBen: but in the end I don't care.
This is just a feature request with code. Do what ever you like with it.
I don't wanna waste our time with endless discussions.

le-jeu: > not "result" ? atleast all stock plugins use "result" instead of "raw"

`.result == .raw.result` is what the intel is sending, unprocessed

> isn't this excatly what I tried to say? Why doing this in every plugin when IITC can standarized it?

I was agreeing to your point, but stating that markup isn't the place for it

core/code/chat.js
xscreach: Exactly, through the div 🤷

You now have `<div><span class="enl">text</></>`
But it's really enough to create only `<div class="enl">text</>`

McBen: there is no div.
`.hmtl() ` will return the innerhtml code. not the tag itself.

like 
html: `<div id="test">Hello</div>`
Js:  `$("#test").html()  => "Hello"`  <- no div, no  "test",  just the inner html

le-jeu: the function returns the string `<span class="enl">text</span>`, it doesn't return `<div><span class="enl">text</></>`

xscreach: Then why the hell create the div in the first place? 🙈

le-jeu: bc it's a common pattern in iitc codebase and jquery codes. You can replace them by `.prop('outerHTML')` since jquery doesn't provide a direct method for outer html

xscreach: So you are basically saying that this
```
$('<div>') 
       .append($('<span>', { class: spanClass, text: text.plain })) 
       .prop("innerHTML");
```

Is better than

```
return $('<span>', { class: spanClass, text: text.plain })
       .prop("outterHTML");
```

And this

```
return $('<div>').text(text.plain).prop("innerHTML").autoLink();
```

Is better than this

```
return text.plain.autoLink();
```
?

McBen: The reason for the code is to convert a string into html code.
like: `$("<div>", {text:"<br>"}).html()` => `&lt;br&gt;`

When using `.prop("innerHTML")` you'll leave JQuery and start using DOMElement code.
I wouldn't mix it.

xscreach: Ok. I'll let it slide for some actual html, but the later case is just nonsense.

Why surround string with div just to get that string out later?
That should be fixed to only `text.plain.autolink()`

McBen: it's just the JQuery way to do:
```
function escapeHTML(str) {
    return str.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;') ;
}
```
or
```
var escape = document.createElement('textarea'); // or div 
function escapeHTML(html) {
    escape.textContent = html;
    return escape.innerHTML;
}
```
( examples taken from: https://stackoverflow.com/questions/5499078/fastest-method-to-escape-html-tags-as-html-entities )
if I recall correclty there is also an IITC helper function in utils.js. This could be used as alternative here.

PR #682, opened 2023-11-20
max machina link indicator

xscreach:
Added max link length table to both cluster dialogs

image
image

Also new layer with max lengths accumulated while moving map (reset only with reload)
image

### DIFF
--- a/plugins/machina-tools.js
+++ b/plugins/machina-tools.js
@@ -252,6 +252,31 @@ function createChildListItem(parent, childData, childPortal) {
   return childListItem;
 }
 
+function getLinkLengths(clusterPortals, initialValue = Array(9).fill(0)) {
+  return Object.values(clusterPortals)
+    .filter((p) => !!p.children)
+    .reduce((previousValue, currentValue) => {
+      previousValue[currentValue.level] = Math.max(previousValue[currentValue.level], ...currentValue.children.map((c) => c.length));
+      return previousValue;
+    }, initialValue);
+}
+
+function renderLinkLen(len, i) {
+  var lenElement = $('<div>', { class: 'machina-link-length' }).text(`L${i}: ${window.formatDistance(len)}`);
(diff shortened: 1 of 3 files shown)


plugins/machina-tools.js
```diff
@@ -646,10 +686,75 @@ machinaTools.showClustersDialog = function () {
...
+        var origin = window.portals[link.options.data.oGuid];
+        if (origin) {
+          var level = origin.options.level;
+          var points = link.getLatLngs();
+          var linkLength = points[0].distanceTo(points[1]);
+          previousValue[level] = Math.max(previousValue[level], linkLength);
+        }
+        return previousValue;
+      }, machinaTools._maxLinks);
+    // TODO: remove 0 element and display warning in case it's bigger than 0
```
xscreach:
```suggestion
```

MysticJay:
The logic should NOT consider any links that originate from "degraded" portals i.e. missing resonators. The link length in that case is neither "exceeding expected length" nor should that length be used for max link length summary, as it would relate to the wrong level.

We could deduct the "real" level of the portal from the resonator's level, even if some are missing, as all have the same level, but this might change. So it seems more safe to just record valid lengths for fully deployed portals.

modos189 (approved):
Looks good. Ready to merge?

xscreach:
@modos189 there were some concers from Johtaja about portals not in current map view.
IMHO those concerns are related to everything in machina-tools as we ignore loaded portals not visible in current map view (behind screen borders) and should be at least investigated, but that could be done in different PR

also, i'm strugling to find some time to work on that, so maybe it will be better to merge this as is and check the "portals not in view" bugs later 🤷‍♀️

and... because the code do not reset the recognised max link lengths, the portals outside of view can be cumulatively added to the max link table/report by moving the map, it should update the table with new lengths but still remember the old ones (if no link in current view is longer that those from previous search)

PR #690, opened 2024-01-20
SEARCH with scanner link or GUID

MysticJay:
Modification to allow pasting of a scanner link into IITC's search. ('%2C' substitution)
Search fro GUID by @le-jeu

### DIFF
--- a/core/code/search.js
+++ b/core/code/search.js
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-  var locations = query.term.replaceAll('%2C',',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
+var locations = query.term.replaceAll(/%2[cC]/,',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
   var added = {};
   if(!locations) return;
   locations.forEach(function(location) {
@@ -438,11 +438,11 @@ addHook('search', function (query) {
   const res = query.term.match(guid_re);
   if (res) {
     const guid = res[0];
-    const data = portalDetail.get(guid);
+    const data = window.portalDetail.get(guid);
     if (data) window.search.addResult(query, data);
     else {
-      portalDetail.request(guid).then(function (data) {
-        addResult(query, data);
+      window.portalDetail.request(guid).then(function (data) {
+        window.search.addResult(query, data);
       });
     }
   }
--- a/core/code/search.js
+++ b/core/code/search.js
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-var locations = query.term.replaceAll(/%2C/gi, ',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
+  var locations = query.term.replaceAll(/%2C/gi, ',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
   var added = {};
   if(!locations) return;
   locations.forEach(function(location) {
--- a/core/code/search.js
+++ b/core/code/search.js
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-  var locations = query.term.match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
+  var locations = query.term.replaceAll('%2C',',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
   var added = {};
   if(!locations) return;
   locations.forEach(function(location) {
--- a/core/code/search.js
+++ b/core/code/search.js
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-  var locations = query.term.replaceAll('%2C',',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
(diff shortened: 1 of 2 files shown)


core/code/search.js
```diff
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-  var locations = query.term.match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
+  var locations = query.term.replaceAll('%2C',',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
```
xscreach:
```suggestion
  var locations = query.term.replaceAll('%2C', ',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
```

xscreach (approved):

core/code/search.js
```diff
@@ -401,3 +401,50 @@ addHook('search', function(query) {
   $.getJSON(NOMINATIM + encodeURIComponent(query.term) + viewbox + bounded, onQueryResult.bind(null, true));
 });
 
+// search for guid
+window.search.addResult = function (query, data) {
```
xscreach: This function is very similar to code starting at line 270...
We should extract the part from line 270 to a function and call it instead to get the same visuals for both "portal results"

xscreach (approved):

core/code/search.js
```diff
@@ -297,7 +297,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
-  var locations = query.term.match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
+var locations = query.term.replaceAll(/%2[cC]/, ',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
```
xscreach: Mhmm... The replacement doesn't work now 😬

Maybe try adding gi into the pattern 🤷
/%2c/gi (global+case insensitive search)


```suggestion
var locations = query.term.replaceAll(/%2c/gi, ',').match(/[+-]?\d+\.\d+, ?[+-]?\d+\.\d+/g);
```

xscreach (changes requested):

core/code/search.js
MysticJay: No longer. Function needed GUID as third parameter, as guid is not part of data.

xscreach: The code is still duplicate and we could have one method instead of 2 slightly different (one is addResilult and second is the part from line 270 - that's not method but it's very much the same as the fubction we added... With some special formatting...

If we don't mind different formatting for the same results, then it's fine 😉

core/code/search.js
```diff
@@ -393,11 +393,57 @@ addHook('search', function(query) {
...
   // http://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases/EN
   var bounded = '&bounded=1';
-  
+
   $.getJSON(NOMINATIM + encodeURIComponent(query.term) + viewbox + bounded, onQueryResult.bind(null, true));
 });
 
+// search for guid
+window.search.addGuidResult = function (query, data, guid) {
+  const teams = ['NEU', 'RES', 'ENL'];
```
le-jeu: this part doesn't know about machina

xscreach (approved):

core/code/search.js
```diff
@@ -393,11 +393,57 @@ addHook('search', function(query) {
...
   // Bounded search allows amenity-only searches (e.g. "amenity=toilet") via special phrases
   // http://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases/EN
   var bounded = '&bounded=1';
-  
+
   $.getJSON(NOMINATIM + encodeURIComponent(query.term) + viewbox + bounded, onQueryResult.bind(null, true));
 });
 
+// search for guid
+window.search.addGuidResult = function (query, data, guid) {
```
xscreach: Yup, this method is duplication of code around line 270...
The only difference that it uses new iitc apis..

Removed duplication and added machina support (as per @le-jeu comment) here - https://github.com/MysticJay/ingress-intel-total-conversion/pull/12

xscreach (changes requested):
Duplicated code and no machina support - merge the second PR before this one 😉

core/code/search.js
MysticJay: fixed

core/code/search.js
MysticJay: merged proposal from @xscreach

core/code/search.js
```diff
@@ -25,6 +25,8 @@ addHook('search', function(query) {});
   selected or the search was cancelled by the user).
 */
 
+/* global L, addHook -- eslint */
```
xscreach:
```suggestion
/* global L -- eslint */
```

`L` is a global type and should be specified here, but `addHook` is our function defined on the `window` object - we should reference it that way (use `window.addHook` instead)

core/code/search.js
```diff
@@ -393,11 +399,26 @@ addHook('search', function(query) {
...
   // Bounded search allows amenity-only searches (e.g. "amenity=toilet") via special phrases
   // http://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases/EN
   var bounded = '&bounded=1';
-  
+
   $.getJSON(NOMINATIM + encodeURIComponent(query.term) + viewbox + bounded, onQueryResult.bind(null, true));
 });
 
+// search on guid
+addHook('search', function (query) {
```
xscreach:
```suggestion
window.addHook('search', function (query) {
```

core/code/search.js
```diff
@@ -297,7 +303,7 @@ addHook('search', function(query) {
 // search for locations
 // TODO: recognize 50°31'03.8"N 7°59'05.3"E and similar formats
 addHook('search', function(query) {
```
xscreach:
```suggestion
 window.addHook('search', function(query) {
```

xscreach (approved):

core/code/search.js
MysticJay: done

core/code/search.js
MysticJay: done

core/code/search.js
MysticJay: done

xscreach (approved):

PR #692, opened 2024-01-22
API for toolbox buttons

modos189:
Created an API for adding buttons to a toolbox with backward compatibility with the old method. For testing purposes, the old toolbox is shown as semi-transparent

Using the new API to add buttons to Toolbox in core and plugins

```
@example
const buttonId = IITC.toolbox.addButton({
  label: 'AboutIITC',
  action: window.AboutIITC
});

@example
IITC.toolbox.addButton({
  id: 'mybtn',
  label: 'Test Button',
  action: () => alert('Clicked!')
});

@example
const buttonId = 'mybtn';
const isUpdated = IITC.toolbox.updateButton(buttonId, { label: 'Updated Button', action: () => console.log('New Action') });
```

closes #666


core/code/toolbox.js
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * Internal method to apply sorting to the buttons.
+   *
+   * @private
+   */
+  _applySort() {
+    const toolbox_component = document.querySelector('#toolbox_component');
+    const buttonElements = Array.from(toolbox_component.children);
+
+    buttonElements.sort((a, b) => this.sortMethod(this.buttons[a.id], this.buttons[b.id]));
```
xscreach: imho we should do smth like this to avoid broken sortMethods from 3rd party plugins. Console log should be enough, but we could also add some warning to toolbox that custom sorting was attemted but failed 🤷‍♀️
```suggestion
    try {
      buttonElements.sort((a, b) => this.sortMethod(this.buttons[a.id], this.buttons[b.id]));
    } catch (e) {
      console.error("Sorting function produced error", e);
      buttonElements.sort((a, b) => this._defaultSortMethod(this.buttons[a.id], this.buttons[b.id]));
    }
```

xscreach (approved):
Very nice work

PR #697, opened 2024-01-30
fix region-names (closes #678)

### LINKED ISSUE #678: Wrong region naming (and probably indexing) in the AS face
IITC currently gives such naming scheme for AS regions:
```
S1 . . . S16
.
.
.
A1
```
but the actual names given by Niantic are:
```
S16 . . . A16
.
.
.
S1
```
I'm guessing that it could be a problem imported from the S2 library, probably [line 74](https://github.com/jonatkins/s2-geometry-javascript/blob/27d91107450fbdb5f600d7f9a576da1ca31b77f2/src/s2geometry.js#L74) and [line 106](https://github.com/jonatkins/s2-geometry-javascript/blob/27d91107450fbdb5f600d7f9a576da1ca31b77f2/src/s2geometry.js#L106).

McBen:
ID and codeword are swapped for some faces (Zones starting with `AS` `PA` or `ST` )

While comparing cell-id calculated by iitc-s2-lib and the original google code.
I wasn't able to find any differences.

So I assume this swap was done on purpose by Niantic.

### DIFF
--- a/plugins/regions.js
+++ b/plugins/regions.js
@@ -68,6 +68,11 @@ window.plugin.regions.regionName = function (cell) {
     var regionI = cell.ij[0] >> (cell.level-4);
     var regionJ = cell.ij[1] >> (cell.level-4);
 
+    // for Odd faces Nia swaps id & codename
+    if (cell.face & 1) {
+      [regionI, regionJ] = [regionJ, regionI];
+    }
+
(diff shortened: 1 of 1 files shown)


xscreach (approved):
Tested with as02-november-13 and nr02-foxtrot-05, both display correct names on map and in the score screen.
Only tested APK, but should work the same way on PC

McBen:
fixed white spaces

PR #701, opened 2024-02-20
Added support for authorization via AppleID, fix Facebook

modos189:
In addition, with a fixed popup size the login page looks more correct: with a smaller font size and without exceeding the page width boundaries

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
@@ -13,6 +13,7 @@ import android.webkit.WebChromeClient;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
+import android.util.DisplayMetrics;
 
 @SuppressLint("SetJavaScriptEnabled")
 public class IITC_WebViewPopup extends WebView {
@@ -100,6 +101,11 @@ public class IITC_WebViewPopup extends WebView {
                             uriHost.startsWith("myaccount.google.") ||
                             uriHost.startsWith("gds.google.")) {
                         Log.d("popup: Google login");
+                        openDialogPopup();
+                        return false;
+                    }
+                    if (uriHost.equals("appleid.apple.com")) {
+                        Log.d("popup: AppleID login");
                         openDialogPopup();
                         return false;
                     }
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
@@ -90,7 +90,7 @@ public class IITC_WebViewPopup extends WebView {
                         return shouldOverrideUrlLoading(view, uriQuery);
                     }
                     if (uriHost.endsWith("facebook.com")
-                            && (uriPath.contains("oauth") || uriPath.equals("/login.php") || uriPath.equals("/checkpoint/")
+                            && (uriPath.contains("oauth") || uriPath.startsWith("/login") || uriPath.equals("/checkpoint/")
                             || uriPath.equals("/cookie/consent_prompt/"))) {
                         Log.d("popup: Facebook login");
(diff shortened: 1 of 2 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
```diff
@@ -103,6 +104,11 @@ public boolean shouldOverrideUrlLoading(final WebView view, final String url) {
                         openDialogPopup();
                         return false;
                     }
+                    if (uriHost.equals("appleid.apple.com")) {
+                        Log.d("popup: AppleID login");
+                        openDialogPopup();
+                        return false;
+                    }
```
xscreach: image

Do i have to comment or is the image enough? 🙈

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
```diff
@@ -145,9 +151,28 @@ public void onDismiss(DialogInterface dialog) {
...
+
+        // Calculate 80% of screen height and 90% of width
+        int height = (int) (displayMetrics.heightPixels * 0.8);
+        int width = (int) (displayMetrics.widthPixels * 0.9);
+
+        // Apply the calculated height and width to the dialog
+        WindowManager.LayoutParams layoutParams = new WindowManager.LayoutParams();
+        layoutParams.copyFrom(mDialog.getWindow().getAttributes());
+        layoutParams.height = height;
+        layoutParams.width = width;
```
xscreach: Yeah, this will make the popups look a bit better, but we should imho use

a) height/width with `match_parent`
b) use horizontalMargin/verticalMargin to specify "The horizontal margin, as a percentage of the container's width, between the container and the widget." - https://www.dre.vanderbilt.edu/~schmidt/android/android-4.0/out/target/common/docs/doc-comment-check/reference/android/view/WindowManager.LayoutParams.html

xscreach (changes requested):

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
modos189: You mean to combine these comditions into one? I think it's better to leave it as it is, because it makes the code more readable

xscreach: Well, this is horrible.
It can be done much better.
What about enclosing the conditions into methods and using one if like

```
If (isFacebook(url) || isGoogle(URL) || isNiantic(URL)) {
  Console.log("login URL encountered")
  Popup stuff....
}
```

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
```diff
@@ -145,9 +152,40 @@ public void onDismiss(DialogInterface dialog) {
...
+
+        // Set horizontal and vertical margins
+        WindowManager.LayoutParams layoutParams = new WindowManager.LayoutParams();
+        layoutParams.copyFrom(mDialog.getWindow().getAttributes());
+
+        // Set width and height to match_parent
+        layoutParams.width = WindowManager.LayoutParams.MATCH_PARENT;
+        layoutParams.height = WindowManager.LayoutParams.MATCH_PARENT;
+
+        // Calculate the margin size
```
xscreach, line 172: This is what we should get rid of.

That's what I meant by the last comment.

Set width/height with constants
Set margins as percentage
Do not calculate margins on our own, let the framework do that thanks to the widths and margins configured.

That's not possible? 🤷

xscreach (changes requested):
Facebook login works but the code needs cleanup

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
modos189: No, I don't know how to do that

PR #702, opened 2024-02-20
Remove option to disable web popups

modos189:


mobile/app/src/main/java/org/exarhteam/iitc_mobile/prefs/ShareDebugInfoPreference.java
```diff
@@ -60,7 +60,7 @@ private String createDialogText() {
         int version = Build.VERSION.SDK_INT;
         String versionRelease = Build.VERSION.RELEASE;
 
-        return getContext().getString(R.string.debug_info_dialog_text, manufacturer, model, version, versionRelease, buildVersion, BuildConfig.BUILD_TYPE, iitcVersion, originalUserAgent, userAgent, getBooleanDescription("pref_fake_user_agent", false), getBooleanDescription("pref_popup", true), getUserPluginCount());
+        return getContext().getString(R.string.debug_info_dialog_text, manufacturer, model, version, versionRelease, buildVersion, BuildConfig.BUILD_TYPE, iitcVersion, originalUserAgent, userAgent, getBooleanDescription("pref_fake_user_agent", false), getUserPluginCount());
```
xscreach, line 63: The text under the key "debug_info_dialog_text" should also be updated - removed the line with webpopup

xscreach (changes requested):
One minot thing to be fixed - without it the debug info causes application crash.

PR #709, opened 2024-02-25
fix: crash on google links

le-jeu:


McBen:
to give a little bit more details of the Bug.

will be opened inside iitc::
`window.open("https://www.google.com")` 

will crash the app:
`window.open("https://www.google.com", "", "noopener")`

will be correclty opened in google map but IITC gets 'shaded' (like showing an invisible popup)
`window.open("https://maps.google.com/maps?layer=c&cbll=48.859139,2.293676")`


my test plugin: https://gist.github.com/McBen/c1ea7abb46ce373103513990f29f227f

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
```diff
@@ -45,18 +45,22 @@ public void onCloseWindow(WebView view) {
...
-            private boolean reloadWithUserAgent(final WebView view, final String url) {
+            public boolean setUserAgentForUrl(final WebView view, final String url) {
                 final Uri uri = Uri.parse(url);
                 final String uriHost = uri.getHost();
 
                 final String currentUA = view.getSettings().getUserAgentString();
                 final String targetUA = mIitc.getUserAgentForHostname(uriHost);
                 if (targetUA == null || currentUA.equals(targetUA))
                     return false;
+                view.getSettings().setUserAgentString(targetUA);
```
xscreach: I can understand how the code became like this, but I'd say it needs a bit... Well, just...

What about...
Having method getTargetUA returning the target UA... And then just..
If (target !== current) setUA... 🤷

The readability of this part will imho improve a lot 😉

xscreach (changes requested):
Clean up 🙏

McBen:
cannot say much about the code style.
but I can confirm: it works

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_WebViewPopup.java
le-jeu: this is not javascript, afaik `.equals` is how to compare two strings in Java, and I definitely don't want to call `setUserAgentString` with a `null` argument

xscreach: oh, i thought you are experienced enough to understand... ok, this is what i meant the result should look like

```java
           private boolean reloadWithUserAgent(final WebView view, final String url) {
                final Uri uri = Uri.parse(url);
                final String uriHost = uri.getHost();
                final String currentUA = view.getSettings().getUserAgentString();
                final String targetUA = mIitc.getUserAgentForHostname(uriHost);
                if (targetUA != null && currentUA.equals(targetUA)) {
                     view.getSettings().setUserAgentString(targetUA);
                     view.loadUrl(url);
                     return true;
                }
                return false;
            }
```

le-jeu: so you want me to change
`if (test) return false; code; return true;` by `if (!test) { code; return true; } return false;`
but why ? this is about taste isn't it ?

McBen: IMHO the "return early pattern" by le-jeu is okay here.

Also xscreach example expression contains an error. That's excatly the reason why we should avoid negative expressions if possible.

The other point here is: you can merge both functions into one.
IMHO both ways have it's pro & con.

le-jeu: the other point was not clear to me but I agree with it, I will change

PR #716, opened 2024-03-12
fix dialogs on iitc boot

McBen:
DIALOGS is not initialized
when "checkingIntelURL" or "extractFromStock" function fails and tries to open a dialog.

example:
error"can't access property "dialog-anon-0", window.DIALOGS is undefined" when not using "intel.ingress.com"

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -255,11 +255,11 @@ function boot() {
   var loadPlugins = prepPluginsToLoad();
   loadPlugins('boot');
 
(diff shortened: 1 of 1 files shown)


McBen:
to reproduce the problem: change the `intel.ingress.com`
in boot.js:98
`if (window.location.hostname !== 'intel.ingress.com' && localStorage['pass-checking-intel-url'] !== 'true') {`
to trigger the warning

xscreach (approved):

PR #723, opened 2024-03-30
Update intel link to match stock.

### LINKED ISSUE #684: `Permalink` and `Portal link` are different than stock intel
Stock intel generates URLs like this:
https://intel.ingress.com/intel?ll=37.423521,-122.089649&z=17
https://intel.ingress.com/intel?ll=37.423521,-122.089649&z=17&pll=37.423521,-122.089649
That is WITH `/intel` in the path name.

IITC generates URLs like this:
https://intel.ingress.com/?ll=37.423521,-122.089649&z=17
https://intel.ingress.com/?pll=37.423521,-122.089649
This is, with OUT the `/intel` portion.

Also, the game's own deep links include `/intel`.

While stock works without the `/intel` portion, it seems like a good way to track those using IITC.

Also, at least one Android helper app, [Portal Intel URL to GPS](https://play.google.com/store/apps/details?id=com.marcosdiez.ingressurltogps) does not recognize the links without `/intel` in them.  I filed https://github.com/marcosdiez/IngressIntelToGps/issues/4 for that.  (Though considering it was last updated Nov 15, 2018, and the fact that the author has a more generic app that handles IITC generated URLs, I'm not sure it would ever be updated.)

nexushoratio:
It looks like when `/intel` was removed with e5fe5d8fe, stock was not adding the string to the URL.  Currently it does in both the web and scanner apps.

Closes #684.

### DIFF
--- a/core/code/utils_misc.js
+++ b/core/code/utils_misc.js
@@ -529,7 +529,7 @@ window.makePermalink = function (latlng, options) {
   if (options.fullURL) {
     url += new URL(document.baseURI).origin;
   }
-  url += '/';
+  url += '/intel';
   return url + '?' + args.join('&');
 };
 


xscreach (changes requested):
This is something niantic didn't do properly... 
We know they won't fix it but do we really want the same "bug" in iitc?
No need to argue,  just asking 😉🤷‍♂️

nexushoratio:
> This is something niantic didn't do properly... We know they won't fix it but do we really want the same "bug" in iitc? No need to argue, just asking 😉🤷‍♂️

While annoying, repeating strings in the hostname and path is not uncommon, and I wouldn't consider it a bug.

Tabs I have open now that do this:
mail.google.com/mail
calendar.google.com/calendar

The fact both of those are Google is not surprising.  They've always been inconsistent about which way they want to go (which probably bled over into Niantic, particularly since maps.google.com/maps was a thing).

xscreach:
> > This is something niantic didn't do properly... We know they won't fix it but do we really want the same "bug" in iitc? No need to argue, just asking 😉🤷‍♂️
> 
> While annoying, repeating strings in the hostname and path is not uncommon, and I wouldn't consider it a bug.
> 
> Tabs I have open now that do this:
> mail.google.com/mail
> calendar.google.com/calendar
> 
> The fact both of those are Google is not surprising.  They've always been inconsistent about which way they want to go (which probably bled over into Niantic, particularly since maps.google.com/maps was a thing).

Niantic changed from something.ingress.com/intel to intel.ingress.com without context path. This is clearly remained after this change and is clearly a bug or smth they didn't finish. Period.
Last month some old urls stoped working even though niantic supported them for so long.
I just don't want be changing this again when nia finds some people to work on ingress (haha)

modos189:
I agree with @xscreach that most likely /intel is an outdated url and at some point Niantic will finally update the url by making the default url without /intel 
For the same reason this change was made to commit https://github.com/IITC-CE/ingress-intel-total-conversion/commit/e5fe5d8feef0ccd9130b2811a54fceb17c15ccc7

nexushoratio:
They are just as likely to someday require `/intel` as they are to do away with it.

Meanwhile, their official software (both web and scanner) includes it.

And IITC includes it when it generates a deep (scanner) link.

So, this change removes an unnecessary difference between stock and IITC.

nexushoratio:
Just checking to see if this has any hope of moving forward, or if I should just withdraw it.

xscreach:
We can keep discussing this for years... and fiddling with it back and forth..or just wait for niantic to do something that would need an update here and change it then 😉

Issue #123, opened 2019-02-27
Issues from abandoned iitc-project


johnd0e:
Feature Requests:
- [ ] Draw without creating crosslinks [#1267](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1267)
- [ ] Add open location code search [#1218](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1218)
- [ ] Show vanilla intel plans on temporary layer [#1197](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1197)
- [ ] Language Support for Arabic and Hebrew [#1200](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1200)
- [ ] Optional portals/function parameter for window.plugin.portalcounts.getPortals [#1171](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1171)
- [ ] Sort "Missions in view" list [#1130](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1130)
- [ ] Mission Filtering [#1082](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1082)
- [ ] Dialog location isn't remembered [#1104](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1104)
- [ ] sync drawn items [#1030](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1030)
- [ ] Add tooltip to player name in portal details view [#1078](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1078)
- [ ] Cache Key Locations [#1039](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1039)
- [ ] Undo for "Snap to Portas" [#1052](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1052)
- [ ] chat and chatcontrols toggle [#1016](https://github.com/iitc-project/ingress-intel-total-conversion/issues/1016)
- [ ] Letting distance to portal plugin use Lat/lng searched for [#996](https://github.com/iitc-project/ingress-intel-total-conversion/issues/996)
- [ ] New color on markedStarterPortals in missions plugin [#989](https://github.com/iitc-project/ingress-intel-total-conversion/issues/989)
- [ ] add-kml: load file from url (web) /improvement request/ [#988](https://github.com/iitc-project/ingress-intel-total-conversion/issues/988)
[...]

PR #724, opened 2024-03-31
Add new bulk star/unstar/toggle bookmarks feature.

nexushoratio:
Scroll to an area and use the layer filter to hide certain classes of portals (level, faction, etc).  Then use the bulk commands to add/remove portals from the bookmarks.

### DIFF
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -743,12 +743,12 @@ window.plugin.bookmarks.loadStorageBox = function() {
     const folders = window.plugin.bookmarks.bkmrksObj['portals'];
 
     const counts = {skip: 0, add: 0, delete: 0};
-    let total = 0;
+    let visible = 0;
 
     for (const [guid, portal] of Object.entries(portals)) {
       // The check for _map restricts to portals actually shown currently
       if (displayBounds.contains(portal.getLatLng()) && portal._map) {
-        total += 1;
+        visible += 1;
 
         // First, figure out what to do with the portal.
         let op = 'skip';
@@ -793,7 +793,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
       window.runHooks('pluginBkmrksEdit',
                       {target: 'all', action: 'import'});
     }
-    console.log('BOOKMARKS:', total, counts);
+    console.log('BOOKMARKS:', `visible: ${visible}`, 'summary:', counts);
   }
 
   window.plugin.bookmarks.dialogLoadListFolders = function(idBox, clickAction, showOthersF, scanType/*0 = maps&portals; 1 = maps; 2 = portals*/) {
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -743,7 +743,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
     const folders = window.plugin.bookmarks.bkmrksObj['portals'];
     const idsInUse = new Set();
 
-    const counts = {'skip': 0, 'add': 0, 'delete': 0};
+    const counts = {skip: 0, add: 0, delete: 0};
     let total = 0;
 
     // Find existing ids in use so we don't accidentally reuse them.
@@ -773,7 +773,7 @@ window.plugin.bookmarks.loadStorageBox = function() {
           case 'skip':
             break;
 
-          case 'add':
+          case 'add': {
             const label = portal.options.data.title;
             const ll = portal.getLatLng();
             const latlng = `${ll.lat},${ll.lng}`;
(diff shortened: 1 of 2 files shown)


nexushoratio:
I first implemented this about 10 years ago.

See https://github.com/iitc-project/ingress-intel-total-conversion/pull/881/files

Also mentioned in #123 in the bookmarks section.

Not the fastest beast.  Processes about 100 portals/sec on my machine.  So, if one has a large map in the browser (via large monitor, or zooming out in the browser while staying zoomed in with the map), it can take 20s to process 2,000 portals.  During that time there is currently no feedback, and, occasionally, the browser, at least Firefox, may give the "the script it slow" message.

plugins/bookmarks.js
```diff
@@ -734,6 +734,82 @@ window.plugin.bookmarks.loadStorageBox = function() {
...
+          op = 'add';
+        }
+        counts[op]++;
+
+        // Then do it.
+        switch (op) {
+          case 'skip':
+            break;
+
+          case 'add':
```
xscreach: Big NO on this.
please envelope each case within own method and call them directly from the IFs above instead of assigning special "action" variable.
It will make the code much better readable

McBen:
What's the use case?

xscreach (changes requested):
Works nicely but needs reimplementation.
Those double nested loops (for-in-for, while-in-for) should be simplified if possible and the "action" variable `op` should be removed in favor of direct function calls.

Summary:

Functionality: working ✅️
Code/implementation style: need rework ❌️

nexushoratio:
> What's the use case?

* Restrict visible portals to only the other faction, bookmark, plan attack
* Restrict visible portals to lower levels, bookmark, plan bumps
* Bookmark all portals in an area, export, compare to a previous export to see what has come/gone/moved

While all of the above are possible now, when dealing with a view that has hundreds of portals on it (or, more likely, panning around to do the same), clicking a single button is a lot easier than clicking all of the ones in the portals list.

plugins/bookmarks.js
nexushoratio: When I first wrote this, there was a definite problem with collisions from `generateID()`.  Back then it was only `time+rand(0,99)`, and so bookmarking more than a handful of portals/second would cause problems.

It looks like that was fixed with https://github.com/iitc-project/ingress-intel-total-conversion/commit/3527cf1acb0e1e816ee49ac9eaa9d4cb948f5c50 .  I'd noticed in my testing I was not seeing any collisions logged.  Now I know why!

Addressing that should remove some of those nested loops.  They were just collecting existing IDs.

nexushoratio: Oh, as a bonus, removing the collision checking seems to have improved throughput from about 100 portals/sec to 500 portals/sec.

nexushoratio: Ugh.  Ignore all comments about speed.

Almost all of that time is spent in `jquerySortableScript()`.

nexushoratio:
Please take another look.

I created a couple of new functions to handle populating the bookmark object.  I would like to eventually have a couple of other places in the file make use of them, effectively replacing/streamlining existing calls to `addPortalBookmark()`.  Hence the `*ByGuid()` variant, currently unused.

The *reviewdog* checks are much cleaner now.  Unsure if I did something on my end, or if you all tweaked something.  If so, thanks!

I'm not well versed in PRs that have updates in it.  Do I need to do anything to *squash*?  (going off to read docs now.)

DanielOndiordna:
Screenshot_20240401-090825_IITC Mobile
My Bookmarks add on plugin adds exactly this functionality. And it works fast (instantly).
[https://softspot.nl/ingress/plugindetails/iitc-plugin-bookmarks-addon](https://softspot.nl/ingress/plugindetails/iitc-plugin-bookmarks-addon)

nexushoratio:
> And it works fast (instantly). https://softspot.nl/ingress/plugindetails/iitc-plugin-bookmarks-addon

How does it update the bookmark widget quickly when there are hundreds of bookmarks?

DanielOndiordna:
> How does it update the bookmark widget quickly when there are hundreds of bookmarks?

By creating a new json formatted bookmarks string and using the steps of the import function of the bookmarks plugin. Actually just as fast if you start iitc and all bookmarks are drawn.

nexushoratio:
> Actually just as fast if you start iitc and all bookmarks are drawn.

Yeah, that isn't so fast for me either.

With only 505 portals bookmarked, it took  just shy of 24 seconds to execute the line `$(".bookmarkList ul li ul").sortable(`.  (Same after restarting the browser.)
```
10:35:44.505 entered jquerySortableScript [intel.ingress.com:663:13](https://intel.ingress.com/?ll=37.477991,-122.303625&z=15%20line%202244%20%3E%20injectedScript)
10:35:44.509 finished .bookmarksAnchor [intel.ingress.com:675:13](https://intel.ingress.com/?ll=37.477991,-122.303625&z=15%20line%202244%20%3E%20injectedScript)
10:36:08.298 finished .bookmarksLink [intel.ingress.com:688:13](https://intel.ingress.com/?ll=37.477991,-122.303625&z=15%20line%202244%20%3E%20injectedScript)
```

This may be getting into a Firefox-on-Linux-using-ViolentMonkey performance issue, and is independent of my code.

nexushoratio:
Just checking to see if this has any hope of moving forward, or if I should just withdraw it.

xscreach:
IMHO since there's maintained plugin to handle exactly this and more, we don't need that functionality in core plugins 🤷‍♂️

PR #725, opened 2024-04-09
Bookmarks refactor: simplying API for adding a new portal

nexushoratio:
Also adds jsdoc for the API being replaced.

### DIFF
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -367,89 +367,89 @@ window.plugin.bookmarks.loadStorageBox = function() {
     }
     // If portal isn't saved in bookmarks: Add this bookmark
     else{
-      window.plugin.bookmarks.addPortalBookmarkByGuid(guid, true);
+    window.plugin.bookmarks.addPortalBookmarkByGuid(guid, true);
     }
   }
 
-  /**
-   * Adds a portal to the default bookmark folder.
-   *
-   * @param {L.circleMarker} marker - As enhanced when added to
-   * window.portals.
-   * @param {boolean} doPostProcess - Whether additional post processing
-   * should be done after the bookmark was added.  E.g., saving to local
-   * storage, refreshing the widget, and running hooks.  If part of a batch
-   * update, this should probably be false.
-   */
-  window.plugin.bookmarks.addPortalBookmarkByMarker = function(marker, doPostProcess) {
-    const guid = marker.options.guid;
-    const label = marker.options.data.title;
-    const ll = marker.getLatLng();
-    const latlng = `${ll.lat},${ll.lng}`;
-    const ID = window.plugin.bookmarks.generateID();
+/**
+ * Adds a portal to the default bookmark folder.
+ *
+ * @param {L.circleMarker} marker - As enhanced when added to
+ * window.portals.
+ * @param {boolean} doPostProcess - Whether additional post-processing
+ * should be done after the bookmark was added.  E.g., saving to local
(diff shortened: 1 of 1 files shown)


nexushoratio:
While, net code wise, there are now more lines, I think this new API would be easier for other plugins to use, rather than remembering how to properly build the latlng string and extract the portal name.

nexushoratio:
Hmmm... apparently rebasing to pick up the recent build fixes was not the right thing to do history wise?  It looks like it mixed in @modos189 's changes with mine, which was not my intent.

I've not done a lot of work with git in a collaborative setting like this, so it is not surprising if I messed something up.

Regardless of the final disposition of this PR, guidance on git protocol in this case would be appreciated.  :->

xscreach:
> Hmmm... apparently rebasing to pick up the recent build fixes was not the right thing to do history wise?  It looks like it mixed in @modos189 's changes with mine, which was not my intent.
> 
> I've not done a lot of work with git in a collaborative setting like this, so it is not surprising if I messed something up.
> 
> Regardless of the final disposition of this PR, guidance on git protocol in this case would be appreciated.  :->

You need to do rebase with "onto" to select the starting point...

Should be smth like this: `git rebase 141e2d1b2 --onto master`

plugins/bookmarks.js
```diff
@@ -367,13 +367,75 @@ window.plugin.bookmarks.loadStorageBox = function() {
...
+   *
+   * @param {string} guid - The GUID of the portal.
+   * @param {boolean} doPostProcess - Whether additional post processing
+   * should be done after the bookmark was added.  E.g., saving to local
+   * storage, refreshing the widget, and running hooks.  If part of a batch
+   * update, this should probably be false.
+   */
+  window.plugin.bookmarks.addPortalBookmarkByGuid = function(guid, doPostProcess) {
+    const marker = window.portals[guid];
+    if (marker) {
```
modos189: So it is possible that the portal will not be found and adding a bookmark will be quietly ignored?
In the old function the addition happens anyway. But whether it is good or bad is a question

nexushoratio: In the previous paths, the portal was looked prior to building the details.

In my edit, that would be lines 370 and 1180 in the original code.  The first is the same lookup, unchecked.  And the second is the `format` callback for the *Portals list* dialog, in which case, the marker is actually available in that context (you can tell it is a marker because it has the `getLatLng()` method).  It could actually make sense to replace that with a call to `addPortalBookmarkByMarker(portal, true);`

By definition, for something to be displayed on the map, there must be an existing `L.circleMarker()` in `window.portals`.  And both the sidebar and *Portals list* use that as a source to populate themselves.  So lack of checking in existing code is not an issue.

But, for any future uses, who knows where the guid may come from?  Say, someone wrote something to turn all saved key counts into bookmarks.  It is likely some of those would not be currently loaded because they're not in view.  So, yes, in that case, the request would be silently dropped.

The only other piece of code I found that does anything interesting when a portal cannot be found by guid is https://github.com/IITC-CE/ingress-intel-total-conversion/blob/2b47ad49356ab8da489d3bbff5d4b029f542f335/plugins/debug-raw-portal-data.js#L29 where all it does is a `console.warn`.

Best I can tell, nothing else uses this information to update user data.

Would something like:
```javascript
else {
  raise new Error(`Could not find portal information for guid "${guid}"`);
}
```
be appropriate/sufficient?

modos189: Yes, let's do that:
1. Replace what you mentioned with addPortalBookmarkByMarker
2. Add raise new Error

PR #746, opened 2024-07-19
Plugin "Scoreboard": Include information on Machina

NineBerry:
This adds information on Machina portals and links  currently visible on the screen to the Scoreboard plugin window. (No to be confused with the regional scoring plugin)


plugins/scoreboard.js
```diff
@@ -153,6 +170,7 @@ function displayScoreboard () {
   } else {
     dialog({
       html: html,
+      width: 400,
```
xscreach: This feels very wrong.
Not all displays will handle that.
It's too static to look good at few places (we'd like it to look good everywhere/on every device)

But not sure what to suggest otherwise

plugins/scoreboard.js
```diff
@@ -86,41 +87,57 @@ function getEntitiesCount (entities,bounds) {
...
 function makeTable (portals,linksCount,fieldsCount) {
 
   var html = '';
   html += '<table>'
-  + '<colgroup><col><col class="enl"><col class="res"></colgroup>'
+  + '<colgroup><col><col class="enl"><col class="res"><col class="mac"></colgroup>'
   + '<tr>'
   + '<th>Metrics</th>'
   + '<th class="enl">Enlightened</th>'
   + '<th class="res">Resistance</th>'
```
xscreach: These names, and css classes used above are saved in constants similar to e.g. TEAM_MAC
Please use those instead of repeating the value

plugins/scoreboard.js
```diff
@@ -86,41 +87,57 @@ function getEntitiesCount (entities,bounds) {
...
   +'<tr><td>Links</td>'
     +'<td>'+linksCount.enl+'</td>'
-    +'<td>'+linksCount.res+'</td></tr>'
+    +'<td>'+linksCount.res+'</td>'
+    +'<td>'+linksCount.mac+'</td>'
+    +'</tr>'
   +'<tr><td>Fields</td>'
     +'<td>'+fieldsCount.enl+'</td>'
-    +'<td>'+fieldsCount.res+'</td></tr>';
+    +'<td>'+fieldsCount.res+'</td>'
```
xscreach: Imho this whole table could be better rendered using some loops. 
I know it was like this for years... but if its not too much trouble, could you rewrite it that way?

Just if you feel up to it and have time for it
Otherwise your changes are as good as the original so... 👍

xscreach (changes requested):
Imho the css width is wrong and should be styles some other way than with width in pixels - thats the only comment i thing should really be considered 

Otherwise just some minor comments, but the code was bad long before this PR so... if it would be possible to improve the old code, would be nice, but that's imho nothing to block this PR for

plugins/scoreboard.js
NineBerry: I have copied this syntax from other default plugins where a constant integer number without any unit name is also passed as width property to the dialog() function. For mobile, this part of the code is not relevant (handled via useAppPanes instead)

NineBerry: I have added a commit that uses "auto" instead of a concrete width. That works good in all tested browsers.

NineBerry:
I have improved the dialog width CSS issue. I might further improve the code quality and function of the plugin (e.g. auto update content when zooming or panning view) but only later on in another PR.

NineBerry:
Look of modified dialog: 

Additionally to Res and Enl portals and links Machina information is also shown. This is very useful for the currently "Shared Memory Global Challange" and will probably be useful for future challenges too to find out number of portals and/or links of a specific faction in a specific area.

image

xscreach (approved):

plugins/scoreboard.js
```diff
@@ -86,41 +91,101 @@ function getEntitiesCount (entities,bounds) {
...
+    '<td>' +
+    fieldsCount.enl +
+    '</td>' +
+    '<td>' +
+    fieldsCount.res +
+    '</td>' +
+    '<td>' +
+    fieldsCount.mac +
+    '</td>' +
+    '</tr>';
```
McBen:
```suggestion
  const lines = [
    ['Portals', portals.enl.total, portals.res.total, portals.mac.total],
    ['avg Level', portals.enl.levels, portals.res.levels, portals.mac.levels],
    ['avg Health', portals.enl.health, portals.res.health, portals.mac.health],
    ['Level 8', portals.enl.level8, portals.res.level8, portals.mac.level8],
    ['Max Level', portals.enl.maxLevel, portals.res.maxLevel, portals.mac.maxLevel],
    ['Links', linksCount.enl, linksCount.res, linksCount.mac],
    ['Fields', fieldsCount.enl, fieldsCount.res, fieldsCount.mac],
  ]

  html += lines.map(line => {
    const cells = line.map(cell => `<td>${cell}</td>`);
    return `<tr>${cells}</tr>`;
  });
```
example for a loop solution 
https://github.com/IITC-CE/ingress-intel-total-conversion/pull/746#discussion_r1684349094

we can even go further. but in my option this is a good compromise between complexity and readability

PR #755, opened 2024-08-28
Fix eslint for all files

modos189:
Fixed the style, in some cases minor refactoring or bug fixes have been made, in such cases I tried to mention it in the commit message

- [X] Fix eslint for core files
- [x] Fix eslint for plugins


xscreach:
Frankly, i was saying we should do this long time ago, but the concensus from TG was that it will mess with history making it harder to find the real "owner" of the changes thus "we" don't want this to happen...

Did this change? I'm personally all for this, but we agreed differently... so... 🤷‍♂️

Are "we" really ok with it? 🤷‍♂️

modos189:
Really, I didn't want to do it before. But it turned out that so much time has passed, and we still have a lot of code that does not conform to the style, which makes it difficult to understand and complicates the preparation of new changes.
So I think we need to bring the files to the style. I want to know what other people think

xscreach:
> Really, I didn't want to do it before. But it turned out that so much time has passed, and we still have a lot of code that does not conform to the style, which makes it difficult to understand and complicates the preparation of new changes.
> So I think we need to bring the files to the style. I want to know what other people think

Cool, cool 👍
I'm all for/pro this thing. As you said, it will make new changes easier in terms of preparing "green" PR, lot fewer people pissed from the constant eslint complains and code better readable (at least not all red in IDE cause of broken codestyle 🤪), etc...

I'll just note that we should test this PR before merging as it's changing every single file (or like 99% of them). It should be all good... it's just eslint fix... but... it could also be, for some stupid reason, very bad 😬

modos189:
> I'll just note that we should test this PR before merging as it's changing every single file (or like 99% of them). It should be all good... it's just eslint fix... but... it could also be, for some stupid reason, very bad 😬

Yes, it's true. For example, I had to add some eslint-disable-next-line in boot.js. After the fixes, the build completed successfully and I didn't notice any errors, but I might have missed something

PR #756, opened 2024-09-05
Add favicon.ico

Konano:
In the original intel, there is `favicon.ico` in the head section.

image

But after the IITC plugin, this line disappeared.

### DIFF
--- a/core/total-conversion-build.js
+++ b/core/total-conversion-build.js
@@ -122,6 +122,7 @@ if (!window.PLAYER || !PLAYER.nickname) {
 // possible without requiring scripts.
 document.head.innerHTML =
   '<title>Ingress Intel Map</title>' +
+  '<link rel="shortcut icon" href="/img/favicon.ico" />' +
   '<style>' +
   '@include_string:style.css@' +
   '</style>' +


mxxcon:
But there's no actual icon file, unless I missed it...?

Konano:
> But there's no actual icon file, unless I missed it...?

I didn't add the ico file, I just added back the ["shortcut icon"](https://intel.ingress.com/img/favicon.ico) that was overwritten by IITC.

modos189 (approved):

modos189:
In what cases is it required? When the icon is not displayed?

xscreach:
> In what cases is it required? When the icon is not displayed?

This is just how to tell browser where to get icon for the site from. Browsers then use the icon for bookmarks and stuff like that.
The default url is iirc /favicon.ico...

This line is also present in the original intel html code.

But imho browser reads this info from the original html before it's rewritten by our stuff, we are not changing the value... so it might be kinda redundant, but maybe better save than sorry 🤷‍♂️

Konano:
> But imho browser reads this info from the original html before it's rewritten by our stuff, we are not changing the value... so it might be kinda redundant

Actually, browsers don't do that, from what I've seen... So if there is no icon url in the rewritten content, the browser will not load the icon.

PR #760, opened 2024-09-17
Handle errors when serializing logged objects in console plugin

### LINKED ISSUE #759: debug plugin causes console statements to throw
In multiple circumstances, the debug plugin throws errors.

Example: `console.log(window)`
```
Uncaught RangeError: Maximum call stack size exceeded
    at JSON.stringify (<anonymous>)
    at <anonymous>:60:16
    at Array.forEach (<anonymous>)
    at debugTab.console.renderLine (<anonymous>:57:8)
    at debugTab.console.log (<anonymous>:86:20)
    at window.console.<computed> [as log] (<anonymous>:110:31)
    at <anonymous>:1:9
```

See also bannergress/iitc-plugin#34

In general, the contract of console.something() is that it must **never** throw.

ewoerner:
When serialization of a value for console output using `JSON.stringify()` fails for any reason, use its `toString()` implementation as a fallback instead of throwing.

Closes #759

### DIFF
--- a/plugins/debug-console.js
+++ b/plugins/debug-console.js
@@ -51,19 +51,22 @@ debugTab.renderLine = function (errorType, args) {
   var text = [];
   args.forEach(function (v) {
     if (typeof v !== 'string' && typeof v !== 'number') {
-      var cache = [];
-      v = JSON.stringify(v, function (key, value) {
-        if (typeof value === 'object' && value !== null) {
-          if (cache.indexOf(value) !== -1) {
-            // Circular reference found, discard key
-            return;
+      try {
+        var cache = [];
+        v = JSON.stringify(v, function (key, value) {
+          if (typeof value === 'object' && value !== null) {
+            if (cache.indexOf(value) !== -1) {
+              // Circular reference found, discard key
+              return;
+            }
+            // Store value in our collection
+            cache.push(value);
           }
-          // Store value in our collection
-          cache.push(value);
-        }
-        return value;
-      });
-      cache = null;
+          return value;
+        });
+      } finally {
+        cache = null;
+      }
     }
     text.push(v);
   });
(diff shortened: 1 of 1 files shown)


le-jeu (approved):

modos189:
I observe that typing `console.log(window)` without and with a fix causes the browser to freeze.

After fixing in Chrome, after waiting for some time, the output is displayed correctly.
In Firefox in both cases the exception “Uncaught InternalError: too much recursion” occurs and the output is incorrect.

xscreach:
> I observe that typing `console.log(window)` without and with a fix causes the browser to freeze.
> 
> After fixing in Chrome, after waiting for some time, the output is displayed correctly.
> In Firefox in both cases the exception “Uncaught InternalError: too much recursion” occurs and the output is incorrect.

That's what i thought... the code is missing a catch... will comment where...

plugins/debug-console.js
```diff
@@ -51,19 +51,22 @@ debugTab.renderLine = function (errorType, args) {
...
           }
-          // Store value in our collection
-          cache.push(value);
-        }
-        return value;
-      });
-      cache = null;
+          return value;
+        });
+      } finally {
```
xscreach: There/here also needs to be a `catch`, otherwise the error thrown will continue until browser handles it.
The catch should also return something for the output...
Maybe smth like

```
} catch (e) {
  return "error rendering: " + String.valueOf(value);
}
```

... the idea is to output the message and string representation of that object which should output some nonsense like `[[Object object]]` - which is still less confusing that either error or empty/undefined value

xscreach (changes requested):

plugins/debug-console.js
le-jeu: indeed, but more likely `v = "error rendering: " + String.valueOf(v);` just in case the dev used `console.log(window, 42)`

PR #770, opened 2024-10-27
Update timestamps each time a build is done.

### LINKED ISSUE #740: Stale timestamps with `build.py --watch`
I run `build.py` with `'version_timestamp': True,` and usually with `--watch`.

The other day I noticed that *About IITC* was showing a build time that was several days old.  It was confusing because I was seeing code changes in the plugin I was working on.  On closer inspection, I noticed that the `// @version` metadata and `plugin_info.dateTimeVersion =` in all userscripts were "wrong".  Or at least not the values I expected.  They would be correct on the first build after restarting `build.py`.

I just did some digging and saw that `build_date` and `build_timestamp` are only set once during the lifetime of the python process, in the call to `settings.load()`.

Since I use ViolentMonkey as my userscript manager, and I happened to use it to explicitly load and track the plugin I was working on, it was loading that particular file fine.  Only once in a very rare time do I want to load all of the rest of the files.  To that extent, I would use VM's *Update scripts* feature.  But, that would often fail because it would use the standard `@version` comparison and see the old versions.  (Alternatively, I could manually force a load of specific user.js, but that requires many more clicks.)

Anyway, a simple change:
```
diff --git a/build.py b/build.py
index a02d77a6..c5a1b0c6 100755
--- a/build.py
+++ b/build.py
@@ -118,7 +118,7 @@ def watch(build_cb, *args, interval=1, **kwargs):
     Every found dependancy (to be watched for changes) build_cb must append
     to deps_list, which is passed to it as additional keyword argument.
     """
-    from time import ctime, sleep, time
+    from time import ctime, gmtime, sleep, strftime, time
     from traceback import print_exc
 
     basetime = None
@@ -134,6 +134,9 @@ def watch(build_cb, *args, interval=1, **kwargs):
             else:
                 continue
         basetime = time()
+        utc = gmtime(basetime)
+        settings.build_date = strftime('%Y-%m-%d-%H%M%S', utc)
+        settings.build_timestamp = strftime('%Y%m%d.%H%M%S', utc)
         print('\nrebuild started [{}]'.format(ctime(basetime)))
         watch_list = []
         try:
```

seems "good enough for me".

Is there anything wrong with that approach in a general case?  As commented in another issue from 2019, around when this was implemented, this would cause userscript managers to reload the built files more often.  At least for VM, this means once/day automatically, or upon explicit request (except in the case of the specific plugin I configure to track).

I don't know how it would affect other userscript managers.  NB: Since I use VM on Firefox, I have to serve up the userscripts via a webserver rather than it tracking the files locally.  Again, I have no idea what that experience would be like.

nexushoratio:
Closes #740.

### DIFF
--- a/build.py
+++ b/build.py
@@ -118,7 +118,7 @@ def watch(build_cb, *args, interval=1, **kwargs):
     Every found dependancy (to be watched for changes) build_cb must append
     to deps_list, which is passed to it as additional keyword argument.
     """
-    from time import ctime, sleep, time
+    from time import ctime, gmtime, sleep, strftime, time
     from traceback import print_exc
 
     basetime = None
@@ -134,6 +134,10 @@ def watch(build_cb, *args, interval=1, **kwargs):
             else:
                 continue
         basetime = time()
+        utc = gmtime(basetime)
+        # Keep formatting in sync with settings.load()
+        settings.build_date = strftime('%Y-%m-%d-%H%M%S', utc)
+        settings.build_timestamp = strftime('%Y%m%d.%H%M%S', utc)
         print('\nrebuild started [{}]'.format(ctime(basetime)))
         watch_list = []
         try:
--- a/settings.py
+++ b/settings.py
@@ -40,6 +40,7 @@ def load(build_name, localfile=None):
     mod.pop('load')
     mod['build_name'] = build_name
     utc = time.gmtime()
+    # Keep formatting in sync with build.watch()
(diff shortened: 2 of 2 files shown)


nexushoratio:
I couldn't figure out how to unify setting the timestamps, so just put in comments.

I would prefer a shared function, but simple as they are properties in one location and a dict in another, I gave up.

PR #772, opened 2024-10-28
Update timestamps each time a build is done

### LINKED ISSUE #740: Stale timestamps with `build.py --watch`

modos189:
Found a way to avoid code duplication

Closes https://github.com/IITC-CE/ingress-intel-total-conversion/issues/740 https://github.com/IITC-CE/ingress-intel-total-conversion/pull/770

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -8,26 +8,17 @@ values from localbuildsettings.py (if exists).
 
 import time
 
+_build_date = None
+_build_timestamp = None
 
-def get_current_timestamp():
-    """Get current build date and timestamp in desired formats."""
+
+def generate_timestamps():
+    """Generate build date and timestamp in desired formats."""
+    global _build_date, _build_timestamp
+
     utc = time.gmtime()
-    return (
-        time.strftime('%Y-%m-%d-%H%M%S', utc),  # build_date
-        time.strftime('%Y%m%d.%H%M%S', utc)     # build_timestamp
-    )
-
-
-def build_timestamp():
-    """Dynamically generate current build timestamp."""
-    _, timestamp = get_current_timestamp()
-    return timestamp
-
-
-def build_date():
-    """Dynamically generate current build date."""
-    date, _ = get_current_timestamp()
-    return date
+    _build_date = time.strftime('%Y-%m-%d-%H%M%S', utc)
+    _build_timestamp = time.strftime('%Y%m%d.%H%M%S', utc)
 
 
 def load(build_name, localfile=None):
@@ -61,6 +52,8 @@ def load(build_name, localfile=None):
     mod = vars(__import__(__name__))
     mod.pop('load')
     mod['build_name'] = build_name
+    mod['build_date'] = lambda: _build_date
+    mod['build_timestamp'] = lambda: _build_timestamp
     base = Path(localfile or __file__).parent
     mod['build_source_dir'] = base
     mod['build_target_dir'] = base / 'build' / build_name
--- a/settings.py
+++ b/settings.py
@@ -6,11 +6,33 @@ Defaults are taken from buildsettings.py, and extended with
 values from localbuildsettings.py (if exists).
 """
 
+import time
+
+
+def get_current_timestamp():
+    """Get current build date and timestamp in desired formats."""
+    utc = time.gmtime()
+    return (
+        time.strftime('%Y-%m-%d-%H%M%S', utc),  # build_date
+        time.strftime('%Y%m%d.%H%M%S', utc)     # build_timestamp
+    )
+
+
+def build_timestamp():
+    """Dynamically generate current build timestamp."""
+    _, timestamp = get_current_timestamp()
+    return timestamp
+
+
+def build_date():
+    """Dynamically generate current build date."""
+    date, _ = get_current_timestamp()
+    return date
+
 
 def load(build_name, localfile=None):
     """Load settings for given iitc build name."""
     import buildsettings as config
-    import time
     from pathlib import Path
     from runpy import run_path
 
@@ -39,9 +61,6 @@ def load(build_name, localfile=None):
     mod = vars(__import__(__name__))
     mod.pop('load')
     mod['build_name'] = build_name
-    utc = time.gmtime()
-    mod['build_date'] = time.strftime('%Y-%m-%d-%H%M%S', utc)
-    mod['build_timestamp'] = time.strftime('%Y%m%d.%H%M%S', utc)
     base = Path(localfile or __file__).parent
     mod['build_source_dir'] = base
     mod['build_target_dir'] = base / 'build' / build_name
--- a/build.py
+++ b/build.py
@@ -48,6 +48,7 @@ def run_cmds(cmds, source, target):
 
 
 def iitc_build(source, outdir, deps_list=None):
+    settings.generate_timestamps()
     run_cmds(settings.pre_build, source, outdir)
 
     iitc_script = 'core/total-conversion-build.js'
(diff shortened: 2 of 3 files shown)


nexushoratio:
What needs to be done to move this forward?

modos189:
> What needs to be done to move this forward?

If you confirm that this PR works properly, I will merge it

nexushoratio:
> If you confirm that this PR works properly, I will merge it

Ahh, sorry.  Didn't realize this was up to me.  :->

nexushoratio:
Embarrassing git novice question (I've not done this step before):

How do I get this locally using `git` instead of `gh` ?

From reading https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally, I thought I could something like:
`git fetch origin pull/772/head:testing`

But I just git an error:
> fatal: couldn't find remote ref pull/772/head

modos189:
> How do I get this locally using `git` instead of `gh` ?

This should work:

```
git fetch origin pull/772/head:fix/build-timestamps
git switch fix/build-timestamps
```

xscreach:
> Embarrassing git novice question (I've not done this step before):
> 
> How do I get this locally using `git` instead of `gh` ?
> 
> From reading https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally, I thought I could something like:
> `git fetch origin pull/772/head:testing`
> 
> But I just git an error:
> > fatal: couldn't find remote ref pull/772/head

You skiped step 5 where you should have `git fetch origin pull/ID/head:BRANCH_NAME`

But thats usually not necessary, just checkout the original branch and test on that...

nexushoratio:
> `git fetch origin pull/772/head:fix/build-timestamps`

That's what I was trying but got the `couldn't find remove ref` error.  At least I was on the right track.  I must have something misconfigured locally.

xscreach:
> > `git fetch origin pull/772/head:fix/build-timestamps`
> 
> That's what I was trying but got the `couldn't find remove ref` error.  At least I was on the right track.  I must have something misconfigured locally.

Thats because you are fetching from your fork, not the original repository.
For that you'd need to add the original remote and then do the fetching from that remote instead of "origin"

I'd really only suggest checking out the fix/build-timestamp branch.
Just make sure your branch is based on latest master

nexushoratio:
> Thats because you are fetching from your fork, not the original repository.

I'd *just* figured that out.  Like I said, novice.  :-/

xscreach:
Maybe, catch us on telegram, we can discuss these things better/faster there...

https://t.me/screach

nexushoratio:
> > What needs to be done to move this forward?
> 
> If you confirm that this PR works properly, I will merge it

Looks good to me.

Thanks!

settings.py
```diff
@@ -6,11 +6,33 @@
...
+def get_current_timestamp():
+    """Get current build date and timestamp in desired formats."""
+    utc = time.gmtime()
+    return (
+        time.strftime('%Y-%m-%d-%H%M%S', utc),  # build_date
+        time.strftime('%Y%m%d.%H%M%S', utc)     # build_timestamp
+    )
+
+
+def build_timestamp():
```
xscreach: Would it be better or more readable to call these functions from the get_current_timestamp method and have the actual building here?

modos189: Do you propose to get rid of `get_current_timestamp()`, move the code to `build_timestamp()` and `build_date()`?
Maybe. I like both options, there are pros and cons to both

nexushoratio: Both have a race condition.  If the second ticks over between the calls to `gmtime()`, the results will be different.

I don't know if there is any code that depends on those values being identical.

xscreach: > Do you propose to get rid of `get_current_timestamp()`, move the code to `build_timestamp()` and `build_date()`?
> Maybe. I like both options, there are pros and cons to both

Well, no... do not remove the get_current_timestamp.
Rather to have the `time.strftime(...)` inside of the "build" methods which would get called from get_current_timestamp with the same "time" parameter and have this parameter default to time.gmtime() in case its called separately.

The point is to not calculate both when we need only one of those.

(This is just about polishing the code, I'm quite ok with the current state, it just feels weird to call method to calculate several values and then use only one 🤷‍♂️)

xscreach: No need to block the PR for this, but if it could be improved I'd be glad 😉

modos189:
I've updated the code to get rid of a possible time unsynchronization within the same build

settings.py
```diff
@@ -6,11 +6,26 @@
...
+
+def generate_timestamps():
+    """Generate build date and timestamp in desired formats."""
+    global _build_date, _build_timestamp
+
+    utc = time.gmtime()
+    _build_date = time.strftime('%Y-%m-%d-%H%M%S', utc)
+    _build_timestamp = time.strftime('%Y%m%d.%H%M%S', utc)
+
+    return _build_date, _build_timestamp
```
xscreach: Is the return needed now since all is "just" saved to the globals?

xscreach (approved):
Maybe just remove the return from generate timestamp, otherwise ok

settings.py
modos189: No, you're right

xscreach (approved):

nexushoratio:
Thanks!

PR #775, opened 2024-10-31
Disable RTL for Portalnames

McBen:
in right-to-left languages some message are screwed up.

like:` x created field @11+ portal mus`
instead of:` x created field @portal  +11 mus`

... not sure if this should be applied to player names too?!?
or better moved to an extra css class?!?

### DIFF
--- a/core/code/comm.js
+++ b/core/code/comm.js
@@ -75,7 +75,7 @@ function _initChannelData(id) {
  * @memberof IITC.comm
  */
 let portalTemplate =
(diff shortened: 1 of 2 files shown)


modos189:
If I'm not mistaken, only Latin characters are allowed in the username

But it is better to use css class for such formatting, it may be useful in other parts of the interface as well

McBen:
to tell the truth I'm not pretty sure how,who set this property in vanilla intel
grafik

Issue #742, opened 2024-06-20
FR: Clear field `(X)` button for `Search location`


nexushoratio:
It is a common feature to have a button to clear a field.

* GMail's internal search has one
* MDN's search bar
* GitHub's search bar

As an X Window's user, I am used to its standard *double-click to select/middle-click to paste* (and the last time I used anything else, I configured it to do the same).

Since the search text sticks around, to do a second search (assuming the next search is already selected), left-click the field, use `Control`+`A` to select all the text, then `Delete` to erase the text, finally middle-click pasting.  It would be faster if I could left-click on a button to clear the field, then paste.

Related. there is no "clear results" button.  To accomplish the same, one has to erase the current search text then hit `Enter` to erase the results.  Against, a clear field button would make that process easier (either click button followed by `Enter` or having the click button clear the results).

None of the plugins that I have installed offer a clear-field feature, so I don't know if there is any prior-art in the code base or not (in particular, whether is is some simple thing that jQuery offers out of the box).

nexushoratio:
I just learned that `type="search"` enables this feature on most browsers, but not yet on [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1654288).

Can be enabled by setting `layout.forms.input-type-search.enabled` to `true` in about:config, however that seems to confuse the UX in search box.  ~~The geolocation button moves down into the search results, and clicking the (x) does not clear the results.~~ The geolocation button moving is intended.  I just never noticed it before!

Guess maybe I'll grudgingly install chrome to compare.

Still, I personally would prefer that clicking on the clear button would remove the results as well.  But I can live with it as is.

PR #779, opened 2024-11-03
Search API

### LINKED ISSUE #771: The `keypress` event is deprecated
While looking into how `search.js` works, I noticed the following bit from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event):

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

And, at least the search box uses this.

modos189:
- Search moved to namespace IITC.search
- The code is rewritten in modern JS, without jQuery and with separation of logic and display
- IITC.search is divided into separate files: IITC.search, IITC.search.Query, IITC.search.QueryResultsView.
- The geolocation button is always placed in the search field
- Added an accordion to expand and collapse search results without jQuery
- Updated icons, added search icon and search field cleanup icon https://github.com/IITC-CE/ingress-intel-total-conversion/issues/742
- The window.search.addSearchResult method has been moved to IITC.search.Query.addPortalResult, just like before plugins could call `query.addResult` to add a search result
- Search supports input in both decimal format (e.g., 51.5074, -0.1278) and DMS format (e.g., 50°31'03.8"N 7°59'05.3"E)
- Backward compatibility with older function calls is provided

screenshot

TODO:

- [x] I think we should mark the methods as private except for the ones that are suggested to be used. For example, query.addResult and query.addPortalResult are public, while the others in IITC.search.Query are private
- [x] Write tests
- [x] Write documentation on migration
- [ ] In a separate PR to use the new API in our codebase

close https://github.com/IITC-CE/ingress-intel-total-conversion/issues/771


nexushoratio:
Just been wanting to comment on this for a while.

I *really* appreciate the clear field enhancement, particularly since it also cancels the current search.  This is something I do dozens of times a day, and this particular change has greatly enhanced my experience here.

Thank you!

PR #784, opened 2024-11-23
Positioning Leaflet elements using CSS Grid

### LINKED ISSUE #781: Unable to close "info screen" on Firefox on Android
Maybe this isn't a supported configuration, but I thought I'd give it a try.

Rather than installing the standalone IITCm app, I would prefer to use to Firefox on Android, particularly since it is now more amenable to extensions.  This particular problem happens with both the IITC button and Violentmonkey US managers.

It starts out looking good:
Screenshot_20241110-122625

When I select a portal it looks good, and I can dismiss the view using the `[X]` button:
Screenshot_20241110-122717

But, if I just use the `tap here for info screen` button, well, I have not yet figured out how to dismiss the dialog:
Screenshot_20241110-122205

Is this simple user error, a bug, or simply an unsupported configuration?  If the last, would you be willing to accept patches to support it better?

modos189:
- The element for selecting highlighting has been moved to `leaflet-control-container`. This allowed to apply the same indentation for all elements. For example, this fixes accessing the element on smartphones without using the app https://github.com/IITC-CE/ingress-intel-total-conversion/issues/781#issuecomment-2467247554 close #781
- Leaflet controls now use CSS Grid instead of complex indentation calculations
- Use Leaflet's native attribution control for Google Maps layer
- Always position scale bar in bottom right corner
-  Disabled closing the Info panel when closing the portal in the mobile version

### DIFF
--- a/core/code/portal_highlighter.js
+++ b/core/code/portal_highlighter.js
@@ -61,12 +61,10 @@ window.updatePortalHighlighterControl = function () {
 
   if (window._highlighters !== null) {
     if ($('#portal_highlight_select').length === 0) {
-      $('body').append("<select id='portal_highlight_select'></select>");
+      $('.leaflet-top.leaflet-left').first().append("<select id='portal_highlight_select' class='leaflet-control'></select>");
       $('#portal_highlight_select').change(function () {
         window.changePortalHighlights($(this).val());
       });
-      $('.leaflet-top.leaflet-left').css('padding-top', '20px');
-      $('.leaflet-control-scale-line').css('margin-top', '25px');
     }
     $('#portal_highlight_select').html('');
     $('#portal_highlight_select').append($('<option>').attr('value', window._no_highlighter).text(window._no_highlighter));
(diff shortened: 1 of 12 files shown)


xscreach:
Tested mobile app, seems to have no effect which is expected 👍

modos189:
I've added another commit now. I forgot to send it earlier.
This also affects IITC Mobile, pressing “x” only closes the portal and does not go to the map. I noticed that I don't use this button in the mobile version and it would be useful to close the active portal if required. In the unofficial IITC for android this button is not visible at all.

core/code/portal_highlighter.js
le-jeu, line 1: could we benefit from making this a leaflet control ?

modos189: For `.leaflet-top, .leaflet-bottom`, `pointer-events: none;` is set in the Leaflet styles
Except for `.leaflet-control`, it is set to `pointer-events: auto;` for clickability

Another option is to remove the `leaflet-control` class from the highlighting selection and set `pointer-events: auto;` for it.

PR #785, opened 2024-11-25
feat: layers: add a 20m range around portals, to aid Wayfarer submissions

sandrotosi:
Currently submitting a new portals that's closer than 20 meters to an existing one, will not make it appear on the map even if accepted.

In order to facilitate the submission of new portals, let's draw a 20m range around portals, making the un-shaded area available for new submissions.

### DIFF
--- a/plugins/wayfarer-range.js
+++ b/plugins/wayfarer-range.js
@@ -1,0 +1,88 @@
+// @author         morph
+// @name           Wayfarer portal submission range
+// @category       Layer
+// @version        0.1.0
+// @description    Add a 20m range around portals, to aid Wayfarer portals submissions
+
+/* exported setup, changelog --eslint */
+/* global L -- eslint */
+
+var changelog = [
+  {
+    version: '0.1.0',
+    changes: ['Initial release (code heavily based on zaprange)'],
+  },
+];
+
+// use own namespace for plugin
+window.plugin.wayfarerrange = function () {};
+window.plugin.wayfarerrange.wayfarerLayers = {};
+window.plugin.wayfarerrange.MIN_MAP_ZOOM = 16;
+
+window.plugin.wayfarerrange.portalAdded = function (data) {
+  data.portal.on('add', function () {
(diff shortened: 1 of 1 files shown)


modos189 (approved):

xscreach:
Is this true? Afaik new portals respect S2 cell of some level (can't remember which) but thanks to that it's possible to submit portals way closer than 20m, they just need to NOT share the S2 cell 🤔

xscreach:
> Is this true? Afaik new portals respect S2 cell of some level (can't remember which) but thanks to that it's possible to submit portals way closer than 20m, they just need to NOT share the S2 cell 🤔

Ok, edits don't respect this, but new nominations do 👍

xscreach (approved):

sandrotosi:
> > Is this true? Afaik new portals respect S2 cell of some level (can't remember which) but thanks to that it's possible to submit portals way closer than 20m, they just need to NOT share the S2 cell 🤔
> 
> Ok, edits don't respect this, but new nominations do 👍

yeah that's the information i have: for a new POI you need to stay away 20m from others, while you can move an existing POI closer than that.

im curious what an S2 cell is tho, can you share some info? thanks!

xscreach:
@sandrotosi well, there are these cells... these red ones are for ingress, the blue ones for pogo... and it was said that only one portal (or pokestop/gym) can be inside of a cell... but I'm not sure how that works these days with the 20m rule 🤷‍♂️



https://raw.githubusercontent.com/IITC-CE/Community-plugins/master/dist/DanielOnDiordna/cells.user.js

sandrotosi:
do you know how to access objects created by another plugin? namely the draw tools in this case?

it would be interesting to be able to "Draw a marker" on the map and get the same 20m radius around it, when "wayfarer range" is selected

modos189:
@sandrotosi I think such this plugin is a good example. It also takes data from Draw tools:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/plugins/done-links.js#L69

PR #789, opened 2024-12-15
Apply a `filter` to `glob` results to verify a file exists.

### LINKED ISSUE #788: `build.py` watch mode regularly crashes due to dangling symlinks
I'm regularly seeing the following due to disappearing emacs autosave files.
```
Traceback (most recent call last):
  File "/home/nexus/src/ingress-intel-total-conversion-git/./build.py", line 141, in watch
    build_cb(*args, deps_list=watch_list, **kwargs)
  File "/home/nexus/src/ingress-intel-total-conversion-git/./build.py", line 88, in backup_and_run
    iitc_build(source, workdir, deps_list=deps_list)
  File "/home/nexus/src/ingress-intel-total-conversion-git/./build.py", line 55, in iitc_build
    build_plugin.process_file(source / iitc_script, outdir, deps_list=deps_list)
  File "/home/nexus/src/ingress-intel-total-conversion-git/build_plugin.py", line 184, in process_file
    script = re.sub(r"'@bundle_code@';", partial(bundle_code, path=path), script)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/__init__.py", line 185, in sub
    return _compile(pattern, flags).sub(repl, string, count)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nexus/src/ingress-intel-total-conversion-git/build_plugin.py", line 133, in bundle_code
    return '\n'.join(map(wrap_iife, sorted(files)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nexus/src/ingress-intel-total-conversion-git/build_plugin.py", line 128, in wrap_iife
    """.format(filename, filename, content=readtext(filename))
                                           ^^^^^^^^^^^^^^^^^^
  File "/home/nexus/src/ingress-intel-total-conversion-git/build_plugin.py", line 103, in readtext
    return filename.read_text(encoding='utf-8-sig')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/pathlib.py", line 1059, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/pathlib.py", line 1045, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'core/code/.#map_data_render.js'
```

nexushoratio:
Also replace a `joinpath` with `/` for consistency.

Closes #788 .

### DIFF
--- a/build.py
+++ b/build.py
@@ -55,7 +55,7 @@ def iitc_build(source, outdir, deps_list=None):
     build_plugin.process_file(source / iitc_script, outdir, deps_list=deps_list)
 
     outdir.joinpath('plugins').mkdir(parents=True, exist_ok=True)
-    for filename in source.joinpath('plugins').glob('*.js'):
+    for filename in filter(lambda p: p.exists(), (source / 'plugins').glob('*.js')):
         build_plugin.process_file(
             filename,
             outdir / 'plugins',
--- a/build_plugin.py
+++ b/build_plugin.py
@@ -129,7 +129,7 @@ var log = ulog('{.stem}');
 
 
 def bundle_code(_, path=None):
-    files = (path / 'code').glob('*.js')
+    files = filter(lambda p: p.exists(), (path / 'code').glob('*.js'))
     return '\n'.join(map(wrap_iife, sorted(files)))
 
 


nexushoratio:
FWIW, file processing in these two modules is a bit inconsistent.

Here I used `filter()` because `bundle_code` used `map()`.

However, in `timestamp()`, a comprehension is used instead.

At my job, `map()` and `filter()` are disallowed in preference to always using comprehensions for consistency.

I also noticed that a similar pattern is used in `web_meta_gen.py`, but since I don't know anything about that file, I did not update it.

This also suggests that this approach should probably be factored out into a standalone function, but I wasn't sure where to put it.  The only existing module used in all cases is `settings.py`, and it seemed strange to toss a helper like this into there.

modos189:
I'm curious as to why it works that way. Why doesn't glob only return existing files? And maybe instead of this implementation, the filter should only return files that don't start with a dot?

McBen:
> I'm curious as to why it works that way. Why doesn't glob only return existing files? And maybe instead of this implementation, the filter should only return files that don't start with a dot?

in #788 he said: `glob returns dangeling symlinks. `(file-links which points to nowhere)
BTW how about files without read-permission ? 

I suggest to check if the file was read correctly before processing (in process_file).

Shouldn't "readtext" throw an exception?

nexushoratio:
> I'm curious as to why it works that way. Why doesn't glob only return existing files? And maybe instead of this implementation, the filter should only return files that don't start with a dot?

The glob returns existing file entries.  The symlink is an entry.  And while emacs happens to use a dot, not all do.

I suspect a similar issue is why @johnd0e added a similar `exists()` check in `timestamp()` with https://github.com/IITC-CE/ingress-intel-total-conversion/commit/4162de90712f24d4484e59779abad026e0830c74

> Shouldn't "readtext" throw an exception?

It is throwing the exception that shows up in the stack trace in my original report (well, actually `io.open`).  However, it does so after the iterator has already been made, so `bundle_code` would need to become slightly more complicated.

And regarding permissions... that's a separate issue.  Should the build fail when a file has no permissions vs a dangling symlink?  Even now, perhaps it should at least warn on the dangling symlink?

However, if there are going to be design questions, perhaps that conversation should happen in the report rather than this PR which could end up being tossed in favor of a cleaner solution?  (Just to make stuff easier to find in the future.)

johnd0e:
> I suspect a similar issue is why @johnd0e added a similar exists() check in timestamp() with https://github.com/IITC-CE/ingress-intel-total-conversion/commit/4162de90712f24d4484e59779abad026e0830c74

Actually not. That was meant to cover the case when some file exists on watch start, but then gets removed.

McBen:
> And regarding permissions... that's a separate issue.

its all the same: no permission, invalid symlink, broken harddrive, out-of-memory ......
it's all "File can't be read"

> would need to become slightly more complicated.
catch the exception. print a user friendly message 'file x couldn't be read' and terminate the build script.

IMHO it's not the job of a build script to workaround user generated problem.
(or explain why borken symlinks are a regular issue on mac)

also it does it without telling the user there was a problem.

nexushoratio:
> IMHO it's not the job of a build script to workaround user generated problem.

A build script that relies on globs should generally be more resilient to such environmental issues.

A more principled solution would be to explicitly list all inputs, not "just whatever file happens to show up in the directory."  But I have a feeling such a change would not fly.

nexushoratio:
I like #791 .  I am going to withdraw this PR.

PR #791, opened 2024-12-18
Skip unreadable files during build

modos189:
I propose this solution with skipping files that could not be read. It is less specific and can be useful not only for bypassing emacs links

replaces #789
closes #789 #788


nexushoratio:
Thank you.

As a silly test, I symlinked `/dev/zero` to a fake plugin.  That was fun.  Took my machine quite a while to page my normal working set back into physical memory!

nexushoratio:
Because I'm an impatient jerk sometimes...  what needs to be done to make this happen?

PR #794, opened 2024-12-27
Fix/cache render details

le-jeu:
Clarify `renderPortalDetails` : its selects, refresh and render portal details.
Split it in a select/refresh part and render part.

Revert part of fa3dddd66914537b9db748bec480bbb4cfd9d225

Invalidate cache entry if map data is inconsistent with portal details. (from map_data_render)

Hook portal details rendering on `portalAdded` and `portalDetailLoaded` event

Use hook data in unique/machina-tools plugins instead of using `portalDetail.get`.

I'm not satisfied with the data flow I've made for the permanent portal markers, as data is store in two places (portal marker and cache) and `createPortalEntity` is used to parse data for map data and portal details.
This fix a bit this issue by parsing portal details twice but it feels wrong to me.

### DIFF
--- a/core/code/portal_detail.js
+++ b/core/code/portal_detail.js
@@ -64,14 +64,19 @@ var handleResponse = function (deferred, guid, data, success) {
   }
 
   if (success) {
+    // Parse portal details
+    var dict = window.decodeArray.portal(data.result, 'detailed');
+    cache.store(guid, dict);
+
     // entity format, as used in map data
     var ent = [guid, data.result[13], data.result];
     var portal = window.mapDataRequest.render.createPortalEntity(ent, 'detailed');
 
+    // Update cache with from current map
     cache.store(guid, portal.options.data);
 
     deferred.resolve(portal.options.data);
-    window.runHooks('portalDetailLoaded', { guid: guid, success: success, details: portal.options.data, ent: ent });
+    window.runHooks('portalDetailLoaded', { guid: guid, success: success, details: portal.options.data, ent: ent, portal: portal });
   } else {
     if (data && data.error === 'RETRY') {
       // server asked us to try again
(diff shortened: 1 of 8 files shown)


core/code/portal_detail.js
```diff
@@ -64,14 +64,19 @@ var handleResponse = function (deferred, guid, data, success) {
   }
 
   if (success) {
+    // Parse portal details
+    var dict = window.decodeArray.portal(data.result, 'detailed');
+    cache.store(guid, dict);
```
modos189, line 69: The data is cached twice, is this done on purpose?

plugins/machina-tools.js
```diff
@@ -359,13 +359,13 @@ function createInfoLink(text, title, clickCallback) {
...
 
-machinaTools.onPortalDetailsUpdated = function () {
+machinaTools.onPortalDetailsUpdated = function (data) {
   var portalData;
 
   // If the portal was cleared then exit.
   if (window.selectedPortal === null) return;
 
-  portalData = window.portalDetail.get(window.selectedPortal);
+  portalData = data.portalData;
```
modos189, line 368: Is this change about code improvement (decoupling plugins from cache) rather than fixing a bug?

le-jeu: both, this code triggers the issue but it could have used the hook data instead

core/code/portal_detail.js
le-jeu: It is. Both for consistency wrt hooks.
First cache is make sure the data is there when the portal entity is created (for `portalAdded` hook)
Second cache updates it (from map data) for `portalDetailLoaded` hook functions. (including renderPortalDetails)

plugins/machina-tools.js
modos189: I think I've found a solution to make it backwards compatible. I'll create a separate PR for this

Issue #792, opened 2024-12-19
Should `deleted` and `experimental` plugins be linted?


nexushoratio:
While manually playing around with some stuff, I noticed that some non-mainstream plugins are flagged with lint issues.

For example:
`eqeqeq`:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/c64ca0e64e19f59788c13e6d2c40308bdc9ae4c6/plugins/deleted/highlight-infrastructure.js#L31

`spaced-comment`:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/c64ca0e64e19f59788c13e6d2c40308bdc9ae4c6/plugins/experimental/marker-canvas-icon-data.js#L20

I looked for, and could not find, a setting that explicitly disabled those directories.

Granted, I am using an **old** version of `eslint` for $REASON, but I was still surprised by a few things.

Maybe a document on how to run `eslint` from the CLI?  I am looking at adding it to `post_build` in my `localbuildsettings.py`.

McBen:
> Should deleted and experimental plugins be linted

in fact they should be removed or moved to a experimental branch.
its dead code.

> how to run eslint from the CLI

`npx eslint <filename>`  
example:  `npx eslint ./core/total-conversion-build.js`
(replace npx by yarn if you're using yarn package manager)

nexushoratio:
I've been doing `eslint <directory>`.  As in `eslint plugins`, or more generally `eslint .`

I take it that is not how others typically run it?

PR #795, opened 2024-12-31
user-location: lint (semi)

nexushoratio:


modos189:
The `mobile` folder is ignored in eslint/prettier. Besides this fix, there are many other style inconsistencies in this file.

I think it's better to move the plugin to the main plugins folder and make changes to make it work in both the mobile app and browsers

nexushoratio:
That ignoring explains why I stumbled upon it.  (Referencing https://github.com/IITC-CE/ingress-intel-total-conversion/issues/792, when I set up my local configs do deal with my old tools, I missed the `mobile/` directory.)

Personally, I tend to address one lint issue at a time rather than a whole file at a time.  Usually the diffs are smaller and easier to review, and easier to rollback if something goes amiss.

Anyway, considering this is already available, I think it would make sense to accept it and the moving could happen afterwards.

PR #805, opened 2025-03-28
fix: missing portal indicators (access+range)

### LINKED ISSUE #797: Range circle no longer shows up when loading a portal via URL
This appears to be another fallout from the recent side panel work.

When visiting a portal directly, it would get the range circle automatically (just like when clicking on the portal):
https://intel.ingress.com/?pll=37.423521,-122.089649&z=19

Image

However, now it does not:

Image

Again, clicking on the portal will show it.

I often open a bunch of portal URLs in separate tabs for various reasons (say, from Portal Counts or Inventory) and often use that range circle to see where I need to be to reach a particular portal.

le-jeu:
this fix a regression from the portal marker factoring `setPortalIndicators` is called from `selectPortal` that was called by `renderPortalDetails`.

Now, the function is called:
 - on click
 - on portal added if selected

It's called twice on click when portal defail is not fresh

Fix #797

### DIFF
--- a/core/code/map_data_render.js
+++ b/core/code/map_data_render.js
@@ -431,6 +431,9 @@ window.Render.prototype.createPortalEntity = function (ent, details) {
     window.portals[data.guid] = marker;
   }
 
+  if (guid === window.selectedPortal) {
+    window.setPortalIndicators(marker);
+  }
 
   window.ornaments.addPortal(marker);
 


nexushoratio (changes requested):
Hah!  That is nearly the same change I came up with last night.

Should `==` be `===`?

Also, by putting in a logging statement, I noticed that this is actually called twice.  Once when `createPortalEntry` is called with `details === 'extended'` and once with `details === 'detailed'`.

The value `detailed` is what is passed whenever a portal is explicitly clicked on.

Anyway, I wasn't sure if it made sense to continue looking to reduce to a single call upon page load, or not bother.

nexushoratio:
Ping?

Kind of miss this.  :->

PR #809, opened 2025-04-18
Fix for new login URL

vykend:
Niantic changed login URL to signin.nianticpatial.com

(properly this time 😅 )

### DIFF
--- a/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
+++ b/mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
@@ -188,7 +188,7 @@ public class IITC_Mobile extends AppCompatActivity
         mAllowedHostnames.put("youtube.com", googleUA);
         mAllowedHostnames.put("facebook.com", mDesktopUA);
         mAllowedHostnames.put("appleid.apple.com", mIITCDefaultUA);
-        mAllowedHostnames.put("signin.nianticlabs.com", mIITCDefaultUA);
+        mAllowedHostnames.put("signin.nianticspatial.com", mIITCDefaultUA);
 
         // enable progress bar above action bar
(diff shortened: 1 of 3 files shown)


mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
```diff
@@ -188,7 +188,7 @@ protected void onCreate(final Bundle savedInstanceState) {
         mAllowedHostnames.put("youtube.com", googleUA);
         mAllowedHostnames.put("facebook.com", mDesktopUA);
         mAllowedHostnames.put("appleid.apple.com", mIITCDefaultUA);
-        mAllowedHostnames.put("signin.nianticlabs.com", mIITCDefaultUA);
+        mAllowedHostnames.put("signin.nianticspatial.com", mIITCDefaultUA);
```
xscreach, line 192: Could we add these new addresses and checks rather than replace them?
I worry they'll be switching it on and off for some time (it's still niantic) or idk... better safe than sorry 😉🤷‍♂️

xscreach (approved):
Imho it's good. I'd rather add the new URL, but it should be good like this too 🤞

Tested the login, works on my machine/phone 👍
Can't see the addresses, but Google asks for permission for nianticspatial.com

mobile/app/src/main/java/org/exarhteam/iitc_mobile/IITC_Mobile.java
vykend: I've added both options for now :)

nexushoratio:
Will new versions of the stable release be made with this change?

modos189:
> Will new versions of the stable release be made with this change?

Yeah, I'm planning to release it soon

PR #813, opened 2025-04-21
drawtools: add delete confirmation dialog

McBen:
show a classic "are you sure" before erasing all of your work

in other words: 
if you hit the "clear all" button in the drawtool bin button sub menu it will raise a confirmation dialog before removing all your drawings


modos189 (approved):

McBen:
changed the code to do the monkey patching the iitc way

modos189 (approved):

PR #814, opened 2025-04-21
Multi-column controlbar

McBen:
arrange control bar icons in multiple columns if screen size is to small

Screenshot 2025-04-21 101616

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -250,6 +250,16 @@ function prepPluginsToLoad() {
 }
 
 /**
+ * update Z-Index of plugins buttons in leaflet container
+ */
+function updateControlBarZIndex() {
+  $('.leaflet-control-container .leaflet-top.leaflet-left').children().each((index, element) => {
+    element.style.zIndex = (10100 - index).toString();
+  });
+}
+
+
+/**
  * The main boot function that initializes IITC. It is responsible for setting up the map,
  * loading plugins, and initializing various components of IITC.
  * @function boot
@@ -293,6 +303,8 @@ function boot() {
 
   window.iitcLoaded = true;
   window.runHooks('iitcLoaded');
+
+  updateControlBarZIndex();
 }
 
 try {
--- a/core/style.css
+++ b/core/style.css
@@ -1253,11 +1253,13 @@ td + td {
 }
 
 #portal_highlight_select {
+  position: absolute;
+  top:5px;
+  left:10px;
   z-index: 2500;
   font-size:11px;
   background-color:#0E3C46;
   color:#ffce00;
-  order: -100;
 }
 
 .leaflet-control-scale {
--- a/core/style.css
+++ b/core/style.css
@@ -128,6 +128,10 @@ a:hover {
   text-decoration: underline;
 }
 
+.leaflet-control {
+  user-select: none; 
+}
+
 .leaflet-control-layers-overlays label.disabled {
   text-decoration: line-through;
   cursor: help;
--- a/core/style.css
+++ b/core/style.css
@@ -132,6 +132,18 @@ a:hover {
   user-select: none; 
 }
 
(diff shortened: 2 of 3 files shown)


MysticJay:
Hi,
what is the basic difference to what we have with "Left SideBar Columns" ins ZASOITEMS?
https://github.com/MysticJay/ZasoItems.CE/blob/master/left-sidebar-columns.user.js

But if that is going to IITC we/I should remove that tweak.

MysticJay:
Hi,
what is the basic difference to what we have with "Left SideBar Columns" ins ZASOITEMS?
https://github.com/MysticJay/ZasoItems.CE/blob/master/left-sidebar-columns.user.js

But if that is going to IITC we/I should remove that tweak.

McBen:
> Hi, what is the basic difference to what we have with "Left SideBar Columns" ins ZASOITEMS? https://github.com/MysticJay/ZasoItems.CE/blob/master/left-sidebar-columns.user.js
> 
> But if that is going to IITC we/I should remove that tweak.

Zasos:
- doesn't work, atleast for me (firefox). looks like a conflict with`.leaflet-left {display:grid;}`
- it doesnt handle the highligher control ([c175c79](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/814/commits/c175c79dc3d6fa4efdeef93ced98b47c548b5519))
- it doesn't handle sub icons overlap (the text button of Drawtools are behind buttons)  ( [0ef9b5e](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/814/commits/0ef9b5e2b79997ebd312efe0ade0a8cfe8712138) )
- interesting: by code it has a fixed column count (2) but this doesn't matter in real life
 
but in the end it's the same just a different css style approach

core/style.css
```diff
@@ -128,6 +128,22 @@ a:hover {
...
+.leaflet-control-container .leaflet-top.leaflet-left {
+  width: 700px;  /* chat default width */
+  height: calc(100% - 126px); /* chat default start */
+  max-height: 90%;
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  align-content: flex-start;
+  align-items: flex-start;
+}
```
modos189, line 144: Can we keep just those lines?
It seems like a good idea to me to leave `#portal_highlight_select` in `.leaflet-top.leaflet-left` so that all the controls are in one place. Otherwise it becomes necessary to manually fix their positioning in different parts of the code.

McBen: When the highlight control is inside the same container as all buttons it will define the column width.
The second column would start on the right of it instead of a compact block below it.

McBen: grafik

McBen:
fyi [0ef9b5e](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/814/commits/0ef9b5e2b79997ebd312efe0ade0a8cfe8712138) it to counter  this:
grafik

core/style.css
modos189: Ok, I haven't found a better solution:(

core/code/boot.js
```diff
@@ -249,6 +249,16 @@ function prepPluginsToLoad() {
   };
 }
 
+/**
+ * update Z-Index of plugins buttons in leaflet container
+ */
+function updateControlBarZIndex() {
+  $('.leaflet-control-container .leaflet-top.leaflet-left').children().each((index, element) => {
+    element.style.zIndex = (10100 - index).toString();
```
modos189, line 257: Can we move this into CSS?

McBen: IDK, maybe there is a better solution ?!

it recalculates Z-Indexs to make sure first button is topmost and last button blow all others.
This way subelements, like the DrawTools "finsh"+"Cancel", are drawn above buttons in next columns.

modos189 (approved):

PR #817, opened 2025-04-28
Expand `makePrimeLink()` into the component parts.

nexushoratio:
There is no functional difference with this change; it just builds the URL on the fly.

Firebase is sunsetting this feature:
https://firebase.google.com/support/dynamic-links-faq

It is unknown if link.ingress.com still uses Firebase, or has implemented its own work-alike.  But by documenting it this way, it may help any future changes that come along.

Also, it opens up the possibility to customize links.  For example, I use a similar breakdown in a Python to create deep links that leaves off the `ios` portion, since I personally do not need them, and they are much shorter.

### DIFF
--- a/core/code/portal_detail_display.js
+++ b/core/code/portal_detail_display.js
@@ -380,6 +380,9 @@ window.rangeLinkClick = function () {
 /**
  * Creates a link to open a specific portal in Ingress Prime.
  *
+ * It is using Firebase's Dynamic Links feature.
+ * https://firebase.google.com/docs/dynamic-links/create-manually
+ *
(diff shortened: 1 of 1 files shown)


xscreach (approved):

modos189 (approved):

nexushoratio:
I'm still not up on how GH works in this case.

After being approved, does another button need to be pressed to actually merge this change?  Or is it still being considered?

modos189:
> After being approved, does another button need to be pressed to actually merge this change? Or is it still being considered?

All ok, I'll do the merge tomorrow

PR #819, opened 2025-04-29
DrawTools: fix snap by creating a new object (closes #818)

### LINKED ISSUE #818: In mobile, move a vertex in a drawn segment, snap to portals moves a portal
1. Using Drawtools in mobie IITC, any current version or beta, create a line with 2 segments connecting 3 portals and Finish 
2. Click on modify drawing 
3. Drag the vertex of the 2 segments off its portal 
4. Save
5. Open DrawTools Opt 
6. Click on Snap to portals
7. Ok
8. IITC will reflect no changes 
9. Click on the portal that was the vertex before Snap. The portal will jump to the line segment vertex.

If you kill IITC and come back in, you can see the portal has returned to its proper location and the vertex is unsnapped 

Submitted by bacon9000

McBen:
find snap point return the coordination object instead of a new L.LatLng object

leaflets L.latLng(A) will return A if A is a L.LatLng object itself

### DIFF
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -659,7 +659,7 @@ window.plugin.drawTools.snapToPortals = function () {
     if (minGuid) {
       var pll = window.portals[minGuid].getLatLng();
       if (pll.lat !== latlng.lat || pll.lng !== latlng.lng) {
-        return L.latLng(pll);
+        return new L.LatLng(pll.lat, pll.lng);
       }
     }
   };


plugins/draw-tools.js
```diff
@@ -659,7 +659,7 @@ window.plugin.drawTools.snapToPortals = function () {
     if (minGuid) {
       var pll = window.portals[minGuid].getLatLng();
       if (pll.lat !== latlng.lat || pll.lng !== latlng.lng) {
-        return L.latLng(pll);
+        return new L.LatLng(pll.lat, pll.lng);
```
xscreach, line 662: Interesting... can you pls explain how is this fixing the issue?
Those two lines should have the same effect... from the first look 🤷‍♂️

xscreach (approved):

plugins/draw-tools.js
xscreach: Ah... we need to make a copy...

Anyways, tested on my S25 and it works with and without this fix 🤷‍♂️

McBen: L.latLng is a shortcut for L.toLatLng. which is used to convert a bunch of formats to a LatLng object.
But if the parameter is already a LatLng object it simply return that object.


`a = L.latLng(1,1);
b= L latLng(a);
b lat =2;

console.log(a);
// { lat: 2, lng: 1} 
// a and b are the same object reference`

In our case instead of setting the coordinates, snap sets the location reference to the portal marker location object.
So when the drawtool marker is moved it also alters the portal marker position.

modos189 (approved):

Issue #554, opened 2021-12-14
Requires IITC Mobile update to be able to send to Google Play


modos189:
Starting from November 1, 2021, API level requirements for app updates on Google Play have been increased, now the minimum supported API level is 30.

A barrier to IITC Mobile updates is the need to switch to scoped storage. As you know, we now use full internal storage access to access the IITC_Mobile folder to store plugins in it.

Perhaps the https://github.com/anggrayudi/SimpleStorage library can help with the transition to scoped storage, but I have not been able to implement it. I also found through experimentation that simply raising the Android API Level to 30 without switching to scoped storage breaks file system access on Android 11 and later versions.

Issue #487, opened 2021-04-18
Compatibility with legacy android versions


johnd0e:
IITC mobile application currently is built for 4.2 (`minSdkVersion 17`), but functionality may be limited.

This issue is created to track known problems.

---
Related: #300

johnd0e:
**4.4 KitKat**

In order to pass Google login - `useragent` string should be changed (it is done by default for KitKat).
After successful login useragent could be set to standard.

johnd0e:
**4.2 Jelly Bean**

- Issues with map baselayers - ssl errors preventing tiles to load. Could be fixed by changing `https` -> `http` in urls (CartoDB, Humanitarian)
- GoogleMutant-based maps are not working: `uncaught referenceError: MutationObserver is not defined`.
  See also #365 notes.
- Search location doesn't work
- Open file doesn't seem to work (draw opt import, kml/gpx)
  Perhaps after #243.

(thanks to @le-jeu for investigation)

PR #828, opened 2025-06-01
IITC Mobile: migrate to Storage Access Framework

### LINKED ISSUE #300: Android KitKat support

modos189:
- Migrate plugin system to use Storage Access Framework (SAF)
- Add PluginManager for unified plugin handling
- When user is prompted to select a folder for plugins storage, the folder picker dialog now attempts to open at IITC_Mobile folder if it exists in primary storage
- Show dialog requesting folder access when enabled user plugins are detected
- Remove legacy storage and pre-Lollipop support (min SDK 21, targetSdkVersion 34)
- Remove Android Beam NFC sharing (Android Beam API was deprecated and completely removed in Android API 29)

After installing the application, go to plugins and click on add new plugin. Previously the system asked for permission to access the files, but now you need to select a folder to store the plugins. You can select the existing `IITC_Mobile` folder from past versions - then already installed plugins will be added as before. New plugins are also saved in this folder. 

fix https://github.com/IITC-CE/ingress-intel-total-conversion/issues/300 https://github.com/IITC-CE/ingress-intel-total-conversion/issues/554 
close https://github.com/IITC-CE/ingress-intel-total-conversion/pull/304 https://github.com/IITC-CE/ingress-intel-total-conversion/issues/487


xscreach:
The folder selection works nice.
It would be better if it could suggest the folder not just by text, but by somehow preselecting the folder in the android folder selector (like somehow telling android which folder we'd like 🤷‍♂️)

PR #829, opened 2025-06-06
Fix：Hack Cooldown Time in Portal info panel enhanced

### LINKED ISSUE #534: Hacktime is not shown correctly
The cooldown timer within IITC is still set to 5 minutes in portal view,even though it's been 2,5 minutes for a while

2021-08-27_22h46_59

ODBLK:
According to "https://ingress.com/news/q4-2022-tempchanges/":
> The following temporary change will return to its original value, effective starting Friday, September 30:
> Portal cooldown between hacks will be reverted to its original value of 300 seconds on neutral Portals and Portals aligned to the opposite Faction, up from 90 seconds.
> Portal cooldown between hacks will be 180 seconds on Portals aligned to the Agents’ Faction, up from 90 seconds

Modify the basic calculation value of the relevant information in the Portal info panel

Fix: #534

### DIFF
--- a/core/code/portal_info.js
+++ b/core/code/portal_info.js
@@ -349,7 +349,8 @@ window.getPortalHackDetails = function (d) {
   // first mod of type is fully effective, the others are only 50% effective
   var effectivenessReduction = [1, 0.5, 0.5, 0.5];
 
-  var cooldownTime = window.BASE_HACK_COOLDOWN;
+  var isFriendly = window.teamStringToId(d.team) === window.teamStringToId(window.PLAYER.team); 
+  var cooldownTime = isFriendly ? FACTION_HACK_COOLDOWN : window.BASE_HACK_COOLDOWN; 
 
   $.each(heatsinks, function (index, mod) {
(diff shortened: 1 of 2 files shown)


modos189 (approved):

core/code/portal_info.js
```diff
@@ -349,7 +349,8 @@ window.getPortalHackDetails = function (d) {
   // first mod of type is fully effective, the others are only 50% effective
   var effectivenessReduction = [1, 0.5, 0.5, 0.5];
 
-  var cooldownTime = window.BASE_HACK_COOLDOWN;
+  var isFriendly = window.teamStringToId(d.team) === window.teamStringToId(window.PLAYER.team);
```
johnd0e, line 352: teamStringToId calls are redundant.

McBen: > teamStringToId calls are redundant.

`portailDetails.team = "R"`

`window.PLAYER.team = "RESISTANCE"`

no, it's not redundant

PR #830, opened 2025-06-07
Add browser geolocation to user-location plugin

modos189:
This PR migrates the `user-location` plugin from IITC Mobile-specific plugins to the general plugin repository and adds browser support for GPS-enabled devices.

Plugin appears as regular plugin in IITC Mobile and its enabled/disabled state syncs with the user location display setting.

### Browser Compass Support

| Browser/Platform | Method Used | Notes |
|------------------|-------------|-------|
| **Modern browsers** | `deviceorientationabsolute` | Chrome, Firefox, Edge, Samsung Internet |
| **Safari iOS** | `webkitCompassHeading` | iOS-specific implementation |
| **Legacy/Alternative browsers** | `absolute === true` fallback | When deviceorientationabsolute unavailable |

on top of https://github.com/IITC-CE/ingress-intel-total-conversion/pull/828


McBen:
Well, while the idea is funny. But I usually not moving my desktop around. The user location plugin only makes sense for the mobile edition.

modos189:
Yes, to run on mobile browsers. Also, mobile apps now come with a plugin that is embedded in the app

nexushoratio:
I do move my laptop around.  I would find this handy at times.

McBen:
I still don't see the traget group of this feature.

> I do move my laptop around. I would find this handy at times.

Browser Geolocation doesn't provide your real location. It's okay for moving the map to your area. 
But the 'you are here' pin will be usually 2-10km off.  ([example](https://browserleaks.com/geo))
/edit (for devices without gps)


anyway, you already did the job. I just hope that it won't require extra steps for the casual user.

modos189:
> Browser Geolocation doesn't provide your real location

It's more for mobile devices, but I have accurate geolocation displayed (by wifi location) on my macbook as well.
No additional actions are required from users, just the possibility to use the plugin in the browser as well

PR #846, opened 2025-07-28
IITC Mobile: Remove custom tile cache

le-jeu:
Currently, IITC uses a custom cache mechanism for map tiles.

On each tile, iitc intercept the request and looks for the corresponding tile in the app directory.
If the tile is found, IITC crafts a http response with the tile.
Otherwise, IITC lets the webview behaving normally AND does an additional HTTP request to the tile server for the tile and store it for further requests.

That means that IITC requests each tile twice first, then relies on its own cache.

While having two requests for each tile is not great, the caching works great and we didn't have visible issues so far.
But it seems like the second HTTP requests from our custom cache doesn't respect the [Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/) from OpenStreetMap.
Indeed it's a simple HTTP requests without proper headers.

As I didn't intercept those second requests, this PR tests if removing the custom cache fix the symptom.

Another solution would be to add proper headers to the second requests to make it match the OSM requirement.


McBen:
that looks good. but wasn't able too test ATM.

Boby360:
Can confirm that while the main release continues to throw access blocked, this release loads the tiles successfully.

le-jeu:
related commits fa89fc67d85f94a7e7c5c79e4b9e7223621f1dac and 3e5124022d58e457313455382c1a8f9dd2f00a28

PR #847, opened 2025-07-30
Fix mobile tile caching with proper HTTP headers

modos189:
This PR fixes IITC Mobile tile caching mechanism to comply with OpenStreetMap's Tile Usage Policy while preserving the benefits of local caching. Instead of removing the custom cache entirely (as proposed in #846), this solution implements proper HTTP headers and eliminates duplicate requests.

1. Single-request architecture: Implemented LazyTileResponse that intercepts WebView requests and handles downloading with proper headers
2. Proper HTTP headers: All tile requests now include:
   - User-Agent
   - Referer: https://intel.ingress.com/
   - Modern Accept: `image/avif,image/webp,image/apng,image/svg+xml,image/*,*/*;q=0.8`
3. Background validation: TileValidationChecker performs HEAD requests to validate cached tiles without blocking UI

### Architecture Comparison

**Before (Double Requests):**
User opens map
      ↓
WebView requests tile
      ↓
IITC_TileManager checks cache → File not found
      ↓
Return null to WebView + Launch DownloadTile in background
      ↓
WebView makes HTTP request (proper headers) ← REQUEST 1
      +
DownloadTile makes HTTP request (missing headers) ← REQUEST 2
      ↓
Result: 2 requests, OSM blocks request 2 due to missing headers

**After (Single Request):**
User opens map
      ↓
WebView requests tile
      ↓
IITC_TileManager checks cache → File not found
      ↓
Return LazyTileResponse
      ↓
LazyTileResponse makes HTTP request (proper headers) ← SINGLE REQUEST
      ↓
Save to cache + Return stream to WebView
      ↓
Background TileValidationChecker validates cache freshness

---

close #846


mobile/app/src/main/java/org/exarhteam/iitc_mobile/TileHttpHelper.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public static void setTileRequestHeaders(URLConnection conn, IITC_Mobile iitc, String url) throws IOException {
+        URL tileUrl = conn.getURL();
+        String host = tileUrl.getHost();
+        
+        // Get User-Agent - use specific one for known hosts, fallback to IITC default for others
+        String userAgent = iitc.getUserAgentForHostname(host);
+        if (userAgent == null) {
+            userAgent = iitc.getDefaultUserAgent();
+        }
```
iandees, line 34: I recommend updating `getUserAgentForHostname()` to add a mapping for `tile.openstreetmap.org` that uses a unique and identifying user agent for requests to OSM's tile services.

iandees (commented):
Hi, volunteer from the OSM team here. Some feedback below as it relates to the tile.openstreetmap.org blocking.

mobile/app/src/main/java/org/exarhteam/iitc_mobile/TileHttpHelper.java
modos189: Ok, thanks!

PR #851, opened 2025-08-09
feat: draw a range around draw-tools markers too

sandrotosi:
This is a bit of a complete rewrite, mostly done by Gemini (hence the tag at the bottom). 

I'm considering renaming this plugin one Niantic Spatial releases the new submissions tool, but im a bit concerned about current users (if any, lol) and how the transition will go; i guess we still have more time to think about it

Co-Authored-By: Google Gemini

### DIFF
--- a/plugins/wayfarer-range.js
+++ b/plugins/wayfarer-range.js
@@ -132,20 +132,29 @@ var setup = function () {
 
   // Hook for draw-tools plugin events
   if (window.plugin.drawTools) {
-    // This function will redraw circles for all markers currently on the map.
-    // It's used for initial load and for events that change multiple markers.
-    const syncAllDrawnMarkers = function () {
-      // 1. Remove all existing circles that belong to markers
-      for (const layerId in window.plugin.wayfarerrange.markerLayers) {
-        window.plugin.wayfarerrange.removeMarker(layerId);
-      }
+    // Sync function that only updates changed markers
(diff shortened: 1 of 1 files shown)


sandrotosi:
@modos189 since you reviewed #785

modos189:
> I'm considering renaming this plugin one Niantic Spacial releases the new submissions tool, but im a bit concerned about current users (if any, lol) and how the transition will go; i guess we still have more time to think about it

We can change the plugin name and description, but keep the same file name - then there won't be any problems when updating

sandrotosi:
thanks for improving the code, it's quite clear i'm not really familiar with js and its code style :)

> > I'm considering renaming this plugin one Niantic Spacial releases the new submissions tool, but im a bit concerned about current users (if any, lol) and how the transition will go; i guess we still have more time to think about it
> 
> We can change the plugin name and description, but keep the same file name - then there won't be any problems when updating

im mostly concerned about the plugin name in the hamburger menu on the top-right: if we change its name, will users that currently have selected "Wayfarer range" get the new plugin selected upon upgrade?

nexushoratio:
Typo in title:  s/arond/around/

nexushoratio:
Oh, also s/Spacial/Spatial/ in body.

sandrotosi:
do you think this is ready to be merged?

PR #860, opened 2025-09-01
Traveling salesman

yavidor:
A plugin to get the best route between the player's location to a bookmark folder of portals


modos189:
Interesting idea. But I think it's better to have it be a third party plugin, add a link to it in the Community plugins collection

PR #861, opened 2025-09-01
Plugin/traveling agent

yavidor:
A plugin to get the best route between the player's location to a bookmark folder of portals


modos189:
Interesting idea. But I think it's better to have it be a third party plugin, add a link to it in the Community plugins collection

yavidor:
@modos189 Ok! thanks

nexushoratio:
In addition to using GMaps API, you might also consider a pure JS fallback.

In the past, I ported such a pure JS implementation (though quite old) at https://code.google.com/archive/p/google-maps-tsp-solver/ to Python to do it offline.  I noticed at least one fork on github as well.

A source of true travel distances (they usually asymmetric) helps, my research back in the day demonstrated that if you have a mix of travel distances and straight-line distances, the ratio (around my area) was generally 1.1:1.

PR #867, opened 2025-09-22
Range display isn't updated on detail load

McBen:
When selecting a portal, the link range is only displayed when details are already loaded.
But it obviously loads afterward.

This updates the circles when the details are loading:

### DIFF
--- a/core/code/portal_detail_display.js
+++ b/core/code/portal_detail_display.js
@@ -190,11 +190,12 @@ window.renderPortalToSideBar = function (portal) {
 
   window.renderPortalUrl(lat, lng, title, guid);
 
-  // compatibility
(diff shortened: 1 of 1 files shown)


McBen:
alternative somewhere a:

`window.addHook('portalDetailLoaded', (details) => window.setPortalIndicators(details.portal));`

and (for the field/link count:
`window.addHook('requestFinished', ()=> window.renderPortalDetails(selectedPortal));`

McBen:
same as #805 just a different approach

#867 will draw indicators on portal sidebar update, after portal detail load is finished
#805 will draw indicators when portal marker are redrawn because of updated data

modos189:
Thanks. In the future we may switch to a hook-based solution

PR #871, opened 2025-10-05
LayerChooser: Restore saved base map reliably when plugins register late

sir1ke:
**Problem**
IITC currently restores the last selected base map (localStorage['iitc-base-map']) during the iitcLoaded hook.
If a base map provider (e.g., a plugin) registers its layer after that hook, the restore fails and IITC falls back to the default (CartoDB Dark Matter). As a result, users see their selected map can reset on load. 
-> Happened to me especially when the window was saved from the last session and gets restored on system start up.

**Root cause**
- The saved base layer name is read only once at startup
- Base map plugins that register after initialization are not re-checked against the saved name
- Thus, the user’s preferred layer is never re-applied, even though it (was) saved in localStorage

**Proposed Solution**
- Keep a stable reference to the last saved base layer name (this.lastBaseLayerName)
- When a new base layer is registered, check whether its name matches the saved one
- If it matches and no user-selected base layer is active yet, automatically activate that layer using showLayer()
- This ensures the correct base map is restored even if its provider loads later

### DIFF
--- a/core/code/layerchooser.js
+++ b/core/code/layerchooser.js
@@ -53,7 +53,8 @@ var LayerChooser = L.Control.Layers.extend({
       }
     }
     this._mapToAdd = options && options.map;
-    this.lastBaseLayerName = localStorage['iitc-base-map'];
+    this.lastBaseLayerName = localStorage['iitc-base-map'] || null;
+    this._earlyRestoreDone = false;
     this._lastPriority = -1000; // initial layers get priority <0
     L.Control.Layers.prototype.initialize.apply(this, arguments);
     this._lastPriority = 0; // any following gets >0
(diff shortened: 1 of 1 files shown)


McBen:
why would a basemap provider register itself after load?
AFAIK the load order problem was fixed long ago and all Basemaps are initialized in "setup" - before "IITCLoaded" is triggered.

Beside of this the code looks okay. I think the "_earlyRestoreDone" is not needed but thats not important.

sir1ke:
I haven't found the reason chrome/tampermonkey loads them in a wierd order.
I have noticed only happens the first time i load chrome after system start - it may be related to the amount of tabs I have saved or they way "inactive" tabs are loaded.

Everytime I open the browser the first time after system start, my selected map tile (OSM) is overwritten. I have pinned the issue to that the map tiles arent available at the moment iitc loads - therefore defaults back to cartodb dark.

This change fixed the issue for me.

modos189 (approved):

PR #872, opened 2025-10-07
enhance unit test

McBen:
- moved mocked functions to extra file `_mocks.js`
- use 'Sinon' to spy/mock functions call

I was not able to import Leaflet, JQuery or jsDom to have a complete the base function setup.
The different environments makes it real hard (node, browser, modul, commonjs,... whatever)

In other projects I (tried jest and then) switched to vitest but I can't say if it would be a solution here too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,6 @@
     "build:local": "./build.py local",
     "build:mobile": "./build.py mobile",
     "fileserver": "./web_server_local.py local",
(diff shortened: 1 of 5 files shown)


modos189 (approved):

McBen:
fixed style
(my local code formater and prettier were fighting)

PR #873, opened 2025-10-07
Preparing Leaflet v2 - remove factory function calls

McBen:
while Leaflet v2 is still in development phase (alpha.1) the breaking changes are already announced.

One of these changes is
- remove all factory functions ( use `x = new L.Marker(..) ` instead of  `x = L.marker(..) `)

IITC uses a lot of these. This can already be changed in current code.

There is also a polyfill lib available to catch these.
IMHO the main code should not rely on this and should only include the polyfill to keep old 3rd-Party plugins running.

### DIFF
--- a/plugins/draw-tools.js
+++ b/plugins/draw-tools.js
@@ -55,7 +55,7 @@ window.plugin.drawTools.getMarkerIcon = function (color) {
     console.warn('Color is not set (default #a24ac3 will be used)');
   }
   // todo: refactor to get rid of getMarkerIcon
-  return L.divIcon.coloredSvg(color);
+  return new L.DivIcon.ColoredSvg(color);
 };
 
 window.plugin.drawTools.currentColor = '#a24ac3';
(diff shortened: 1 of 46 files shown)


McBen:
done

changes from initial PR
- ornaments needed the factory methode
- geodesic circle used factory methode to clone latlng object
- bookmark used latLng conversation done by factory method
- removed some autoformated code

plugins/draw-tools.js
```diff
@@ -1001,19 +1000,19 @@ function loadExternals() {
   try {
     // https://github.com/Leaflet/Leaflet.draw
     // eslint-disable-next-line
-    '@include_raw:external/leaflet.draw-src.js@';
+    ('@include_raw:external/leaflet.draw-src.js@');
```
modos189: Here and in other similar cases, you need to remove the parentheses. This breaks the import

`[Violentmonkey][IITC plugin: Draw tools] SyntaxError: missing ) in parenthetical`

modos189 (changes requested):

plugins/draw-tools.js
McBen: ups,
no idea how this happend
fixed  [replace L.divIcon factory](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/873/commits/1ef57a5d60cf870ffd6842e274b23bc04875e40b) and [replace L.latLng & latLngBounds factories](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/873/commits/7ea4e4b52cd7c005dcc4bcdbf3bae4fa20bb5b85)

modos189: Yes, some code formatters make such replacements, but usually `eslint-disable-next-line` is sufficient to prevent this

modos189 (approved):

PR #883, opened 2026-02-22
Portal Names: Add configurable label offsets

Heistergand:
## Summary
This PR adds positioning controls to the existing Portal Names plugin so portal labels can be moved away from other portal overlays.

## What’s included
- New “Portal Names Opt” entry in the toolbox
- X/Y offset sliders (with persistence via localStorage)
- Optional dashed leader line from label to portal
- Optional highlight of the docking edge (only enabled when the leader line is on)
- Updated overlap/collision logic to account for the new label positioning

## Why
When other plugins render information on/near portals, portal names can become hard to read or visually merge with those overlays. Being able to offset labels makes the map much clearer while keeping the Portal Names layer simple and familiar.

## Notes
- Implemented as an update to the existing Portal Names plugin (not a separate plugin), since it’s a general usability feature.
- Backwards-compatible defaults: if offsets are left at 0 and the leader line is off, behavior matches the original plugin.

### DIFF
--- a/plugins/portal-names.js
+++ b/plugins/portal-names.js
@@ -1,13 +1,17 @@
 // @author         ZasoGD
 // @name           Portal Names
 // @category       Layer
-// @version        0.2.4
+// @version        0.3.0
 // @description    Show portal names on the map.
 
 /* exported setup, changelog --eslint */
-/* global L -- eslint */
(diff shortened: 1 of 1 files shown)


Heistergand:
Closed because it's shit.

PR #905, opened 2026-04-24
Prefetch portal details (desktop only)

McBen:
Make portal details appear instant.

In reality, this does not speed up the actual request itself.   
Instead, it preloads portal data when the user hovers over a marker.

This gives IITC a head start when the user selects a portal, making the details feel instant.  
The trade-off is **increased network traffic** in cases where the user hovers over a marker but does not click it.

note: mouseover won't work on mobile, so this is desktop only

### DIFF
--- a/core/code/portal_detail.js
+++ b/core/code/portal_detail.js
@@ -58,30 +58,31 @@ window.portalDetail.remove = function (guid) {
   return cache.remove(guid);
 };
 
-var handleResponse = function (deferred, guid, data, success) {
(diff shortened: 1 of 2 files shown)


McBen:
here as plugin with statistics:
https://github.com/McBen/IITCDev

direct plugin link https://github.com/McBen/IITCDev/raw/refs/heads/main/dist/iitc_plugin_DevTools.dev.user.js

( dev version, includes soucemap, so you can also find the code in f12->debugger->webpack->iitc_plugin_devtools )

PR #909, opened 2026-04-26
url-parameter: move to own function

McBen:
IITC (and Intel) support two URL parameters:

- `pll` — selects a portal by latitude/longitude when opening the map  
- `pguid` — selects a portal by GUID when opening the map  

Currently, these parameters are parsed at startup and then checked on every new portal creation.

This PR moves that logic into a dedicated hook, which is only active until the portal selection has been fulfilled.

The functions `selectPortalWhenLoadedByLatLng` and `selectPortalWhenLoadedByGuid` now register these hooks, replacing the corresponding logic previously located in `createPortalEntity`.

(this separated the responsibilty and saves a nanosecond in the render loop)

### DIFF
--- a/core/code/portal_data.js
+++ b/core/code/portal_data.js
@@ -1,4 +1,4 @@
-/* global L -- eslint */
+/* global L, log -- eslint */
 
 /**
  * @file Contain misc functions to get portal info
@@ -87,7 +87,7 @@ window.zoomToAndShowPortal = function (guid, latlng) {
   // if the data is available, render it immediately. Otherwise defer
   // until it becomes available.
   if (window.portals[guid]) window.renderPortalDetails(guid);
-  else window.urlPortal = guid;
+  else window.selectPortalWhenLoadedByGuid(guid);
 };
 
 /**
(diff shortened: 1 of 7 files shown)


le-jeu (approved):

core/total-conversion-build.js
```diff
@@ -685,20 +685,6 @@ window.TEAM_CODE_MAC = window.TEAM_CODES[window.TEAM_MAC];
...
- */
-window.urlPortal = null;
-
-/**
- * Portal lng/lat if the orignial URL had it.
- * @type {object|null}
- * @memberof storage_variables
- */
-window.urlPortalLL = null;
-
```
modos189, line 701: move to _deprecated.js

core/code/portal_data.js
```diff
@@ -115,8 +115,66 @@ window.selectPortalByLatLng = function (lat, lng) {
...
+window.selectPortalWhenLoadedByLatLng = (latLng) => {
+  if (urlPortalLL === undefined) {
+    window.addHook('portalAdded', testPortalLagLng);
+  }
+
+  urlPortalLL = latLng;
+  window.urlPortalLL = latLng;
+};
+
+const testPortalLagLng = (data) => {
```
modos189: typo
```
- testPortalLagLng
+ testPortalLatLng
```

core/code/portal_data.js
```diff
@@ -115,8 +115,66 @@ window.selectPortalByLatLng = function (lat, lng) {
...
+ * @function
+ * @name selectPortalWhenLoadedByGuid
+ * @param {string} guid - the guid of the portal
+ */
+window.selectPortalWhenLoadedByGuid = (guid) => {
+  if (urlPortal === undefined) {
+    window.addHook('portalAdded', testPortalGuid);
+  }
+
+  urlPortal = guid;
```
modos189, line 165: add
`window.urlPortal = guid;`
like in `window.selectPortalWhenLoadedByLatLng`

core/code/portal_data.js
```diff
@@ -115,8 +115,66 @@ window.selectPortalByLatLng = function (lat, lng) {
...
+  urlPortalLL = latLng;
+  window.urlPortalLL = latLng;
+};
+
+const testPortalLagLng = (data) => {
+  if (data.portal.getLatLng().equals(urlPortalLL)) {
+    log.log(`urlPortalLL ${urlPortalLL.toString()} matches portal GUID ${data.portal.options.guid}`);
+    window.selectedPortal = data.portal.options.guid;
+    console.assert(window.portals[window.selectedPortal], 'no portal data');
+    window.renderPortalDetails(window.selectedPortal, true);
```
modos189: I clicked “Portal link” and loaded the page. The console showed the error `Assertion failed: no portal data`, but the portal had been selected and its data had loaded. It seems these two lines are redundant.

core/code/portal_data.js
```diff
@@ -115,8 +115,66 @@ window.selectPortalByLatLng = function (lat, lng) {
...
+    window.addHook('portalAdded', testPortalLagLng);
+  }
+
+  urlPortalLL = latLng;
+  window.urlPortalLL = latLng;
+};
+
+const testPortalLagLng = (data) => {
+  if (data.portal.getLatLng().equals(urlPortalLL)) {
+    log.log(`urlPortalLL ${urlPortalLL.toString()} matches portal GUID ${data.portal.options.guid}`);
```
modos189: It seems that `log.log` isn't supported in the `ulog` library - the message isn't displayed in the console. In any case, it's best to remove it.

modos189 (changes requested):

core/code/portal_data.js
McBen: supporting deprecated data/functions won't help to remove them.
anyway.

core/code/portal_data.js
McBen: yes, the assert is wrong here. 
because "portalAdded" is triggered before storing the portal data.
(BTW: it won't happen on portals with links)

console.assert doesn't interrupt the execution. so "renderPortalDetails" is executed anyway and required.

core/code/portal_data.js
```diff
@@ -115,8 +115,67 @@ window.selectPortalByLatLng = function (lat, lng) {
...
+
+  urlPortal = guid;
+  window.urlPortal = guid; // @deprecated
+};
+
+const testPortalGuid = (data) => {
+  if (data.portal.options.guid === urlPortal) {
+    log.log(`urlPortal GUID ${window.urlPortal} found - selecting...`);
+    window.selectedPortal = urlPortal;
+    console.assert(window.portals[urlPortal], 'no portal data');
```
modos189: the log and console.assert are still here

modos189 (changes requested):

modos189:
Actually, that's a separate issue: why doesn't `ulog` output `log.log` or `log.debug`?
We could have just left the log as `log.debug`

modos189 (approved):

McBen:
> Actually, that's a separate issue: why doesn't `ulog` output `log.log` or `log.debug`? We could have just left the log as `log.debug`

use https://intel.ingress.com/?log=debug

grafik



/edit
I'm sure there was another option to set log-level. but I can't remember.
I usually prefer classic console commands.

McBen:
re-added the log entries

PR #910, opened 2026-04-26
uniques: move GUID cache to plugin

McBen:
The position cache is currently only used by the Unique plugin, where it is required to resolve portal GUIDs from chat messages.

This PR moves cache management into the plugin itself.  
As a result, the cache is only populated and maintained when the plugin is active, instead of being updated on every portal render.


le-jeu (approved):

modos189:
I suggest not replacing these functions with no-ops, but moving them as-is into "_deprecated.js", in case some plugins still rely on them. They can be fully removed later.

In the uniques plugin itself, it would be better to keep this logic internal and not expose "findPortalGuidByPositionE6" / "pushPortalGuidPositionCache" on "window".

McBen:
the main reason for this PR was to remove the `pushPortalGuidPositionCache` from main render routine.

I thing just removing the caching part should be good enough.

And I believe the Field+Links checks are redundant.

modos189:
By doing that, we're not deprecating findPortalGuidByPositionE6 / pushPortalGuidPositionCache - we're effectively removing them, since replacing them with no-ops is functionally the same as deletion. Even if no known community plugins use these functions, we can't be sure about private plugins in the wild.

I think we should move the full implementations into _deprecated.js, using the same portalAdded hook approach you're introducing in uniques - if that's possible. If not, we could at least keep findPortalGuidByPositionE6 functional without the cache (just the live lookups through window.portals), so plugins relying on it don't silently break. The cache part is less critical - the important thing is that the functions actually work.

McBen:
reworked:
- moved cache stuff to unique plugin
- removed cache from main
- keeping the findPortalByPositionE6

refactored `findPortalByPositionE6` to increase preformance.
there is no need for searching in links and field list because all portals are in the portal list.

plugins/uniques.js
```diff
@@ -597,6 +597,7 @@ var setup = function () {
   window.addPortalHighlighter('Uniques', window.plugin.uniques.highlighter);
   window.addHook('portalDetailsUpdated', window.plugin.uniques.onPortalDetailsUpdated);
   window.addHook('publicChatDataAvailable', window.plugin.uniques.onPublicChatDataAvailable);
+  window.addHook('publicChatDataAvailable', storeGUID);
```
modos189:
```
error running hook publicChatDataAvailable 
TypeError: can't access property "options", hookData.portal is undefined 

callback:  
function storeGUID(hookData)
 data:  
Object { raw: {…}, result: (50) […], processed: {…} }
```

Possibly, the correct hook is `portalAdded`

core/code/portal_data.js
```diff
@@ -178,78 +178,17 @@ const testPortalGuid = (data) => {
...
-        .forEach(function (item) {
-          delete cache[item[0]];
-        }); // delete the rest
-      cache_level = Object.keys(cache).length;
-    }
-  };
-})();
+/**
+ * Finds a portal GUID by its position. Searches through currently rendered portals.
+ * Note: this includes fields and links.
```
modos189, line 183: Old comment. Not includes fields and links

core/code/_deprecated.js
```diff
@@ -275,3 +275,16 @@ window.urlPortal = null;
...
+/**
+ * Pushes a portal GUID and its position into a cache.
+ *
+ * @function
+ * @name pushPortalGuidPositionCache
+ * @param {string} guid - The GUID of the portal.
+ * @param {number} latE6 - The latitude in E6 format.
+ * @param {number} lngE6 - The longitude in E6 format.
+ *
+ * @deprecated GUID are nolonger cached.
```
modos189:
```suggestion
 * @deprecated GUID are no longer cached.
```

plugins/uniques.js
```diff
@@ -607,3 +608,58 @@ var setup = function () {
...
+
+  if (cache_level > GC_LIMIT) {
+    Object.keys(cache) // get all latlngs
+      .map(function (latlng) {
+        return [latlng, cache[latlng][1]];
+      }) // map them to [latlng, timestamp]
+      .sort(function (a, b) {
+        return b[1] - a[1];
+      }) // sort them
+      .slice(GC_KEEP) // drop the MRU
```
modos189:
```suggestion
      .slice(GC_KEEP) // drop the LRU
```

plugins/uniques.js
McBen: copy&paste error - it has to be the hook `portalAdded`

core/code/_deprecated.js
McBen: fixed

plugins/uniques.js
McBen: copy&pasted from original source.
I had no idea what this abbreviations stands for. .. replaced it

core/code/portal_data.js
McBen: This is intentional.
The code obscures the fact that fields and links are still being checked.
(because all link+field portals are included in portal list)

McBen:
fixed requested changes  (except  `Note: this includes fields and links.` )

plugins/uniques.js
modos189: Most Recently Used (MRU)
Least Recently Used (LRU)

So, there was a typo in the original. We're deleting the rarely used records from cache

modos189 (approved):

PR #916, opened 2026-05-04
sync: add sign-out button and display logged-in account email

KrzysztofHajdamowicz:
Previously users had no way to sign out of Google Drive within the plugin, making it impossible to switch accounts without revoking access through Google Account settings. This adds a Sign Out button and displays the connected account's email address in the Sync dialog.

Key changes:
- Add Sign Out button (uses gapi signOut, not disconnect, to preserve granted scopes for smoother re-authorization)
- Display logged-in Google account email via getBasicProfile()
- Guard auto-sign-in with _userInitiatedAuth flag so signing out doesn't immediately re-trigger the sign-in prompt
- Clear all sync intervals and reset registered map state on sign-out
- Force file search on next auth to avoid reusing cached file IDs from a different account

### DIFF
--- a/plugins/sync.js
+++ b/plugins/sync.js
@@ -1,13 +1,17 @@
 // @author         xelio
 // @name           Sync
 // @category       Misc
-// @version        0.5.3
+// @version        0.5.4
 // @description    Sync data between clients via Google Drive API. Only syncs data from specific plugins (currently: Keys, Bookmarks, Uniques). Sign in via the 'Sync' link. Data is synchronized every 3 minutes.
 
 /* exported setup, changelog --eslint */
 /* global IITC -- eslint */
 
 var changelog = [
+  {
+    version: '0.5.4',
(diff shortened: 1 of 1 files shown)


modos189:
Text editors usually have built-in support for ESLint and automatically format the code. Which one do you use?
And please change the plugin version to 0.6.0

KrzysztofHajdamowicz:
> Text editors usually have built-in support for ESLint and automatically format the code. Which one do you use? And please change the plugin version to 0.6.0

vscode, but that's my first JavaScript-related pull request.

modos189:
I think all you need to do is run `npm install` in the root of the repository and install ESLint from the VS Code extensions

KrzysztofHajdamowicz:
Okay, formatting fixed. Thanks for the tip!

modos189 (approved):

PR #917, opened 2026-05-04
draw-tools: sync drawn items across devices via Sync plugin

KrzysztofHajdamowicz:
## Summary

Wires `plugins/draw-tools.js` into the existing `plugins/sync.js` infrastructure so drawn items (polylines, polygons, circles, markers) are kept in sync across browsers and devices through the user's Google Drive. Previously the only cross-device option was manual JSON export/import.

## How it works

- Each drawn item now carries a stable UUID that round-trips through `save()` / `load()` and the share/import flow.
- The synced field `window.plugin.drawTools.itemMap` is a flat map keyed by `"<projectStorageKey>::<uuid>"`. This keeps the per-item incremental granularity Sync expects while letting a **single registered field cover every Multi-Projects-Extension project** — including projects that arrive purely from a peer device.
- MPE project deletions are propagated by listening to the `mpe` hook and dropping orphaned entries from the synced map.
- Legacy id-less drawings are migrated once at boot: ids are minted, the canonicalized JSON is written back to localStorage, and (if the active project was rewritten) its on-map layers are reloaded so layer ids match the persisted data.
- Local edits are debounced (5s) before pushing through `plugin.sync.updateMap`, mirroring the Keys/Uniques pattern.
- On a remote `fullUpdated` (offline catch-up from a peer), every project's localStorage is rewritten from `itemMap`, MPE is asked to rescan, and the active project's layers are rebuilt — firing the existing `pluginDrawTools` `import` hook so dependent plugins (Cross-Links, Done-Links, Wayfarer-Range, etc.) stay coherent.

## Compatibility

Verified compatibility with every plugin that touches `window.plugin.drawTools` or the `pluginDrawTools` hook — no consumer changes required:

- `cross-links.js`, `done-links.js`, `wayfarer-range.js` — already handle non-`layerCreated` events by re-checking, so they react correctly to the `import` event fired after a remote rebuild.
- `bookmarks.js` — fires `draw:created` on the map; our existing handler routes it through `save()`, which mints the id.
- `layer-count.js`, `link-show-direction.js`, `tidy-links.js`, `fly-links.js` — only iterate `drawnItems` / use `filterEvents`; both contracts preserved.

`sync.js` description was updated to include Draw Tools in the list of synced plugins (patch bump to push the metadata refresh to existing installs).

## Versions

- `draw-tools` 0.12.1 → **0.13.0** (new feature)
- `sync` 0.6.0 → **0.6.1** (description update)

## Test plan

- [ ] Two browsers signed into the same Google account, both with Draw Tools and Sync enabled. Draw a polyline in A → after the 3-min poll the polyline appears in B.
- [ ] Edit / delete the item in A → propagates to B.
- [ ] Install MPE in both. Create a second project in A, draw a marker only in that project, switch back to default → marker is hidden locally, present in `itemMap`, syncs to B, and appears in B when B switches to that project.
- [ ] Pre-populate `localStorage['plugin-draw-tools-layer']` with id-less legacy data → confirm load mints UUIDs, persists them back, and seeds `itemMap` correctly.
- [ ] Disable network mid-edit, draw items, re-enable → confirm queue flushes after re-auth.
- [ ] Confirm Cross-Links re-evaluates after a remote rebuild and Wayfarer-Range circles re-attach to incoming markers.

### DIFF
--- a/plugins/bookmarks.js
+++ b/plugins/bookmarks.js
@@ -1,13 +1,14 @@
 // @author         ZasoGD
 // @name           Bookmarks for maps and portals
 // @category       Controls
-// @version        0.4.7
+// @version        0.4.8
 // @description    Save your favorite Maps and Portals and move the intel map with a click. Works with sync. Supports Multi-Project-Extension
 
 /* exported setup, changelog --eslint */
 /* global IITC, L -- eslint */
 
 var changelog = [
+  { version: '0.4.8', changes: ['Register with the Sync plugin regardless of plugin load order'] },
   { version: '0.4.7', changes: ['Fix data reset issue'] },
   {
     version: '0.4.6',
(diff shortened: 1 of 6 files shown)


McBen:
it includes PR #916.
Only do this if they are truly interdependent.

PR #918, opened 2026-05-05
mobile: remove obsolete iitc-update handling

McBen:
This java bridge was used by removed (and obsolete) update-plugin.


McBen:
This bug fixes a serious security issue that was reported in the old IITC-ME version.

PR #921, opened 2026-06-01
Fix incorrect "resonator levels needed for next portal level"

### LINKED ISSUE #734: Level 1 portals show incorrect "resonator level(s) needed for next portal level"
I have a portal that's level 1 from a single level 5 resonator. It should require 11 more resonator levels to become level 2. However, IITC says 8, probably because it doesn't quite understand how level 1 works differently from all other levels:

Screenshot from 2024-05-17 23-08-44

jimsug:
`getPortalLevel()` clamped its result to a minimum of 1, so a portal whose resonator levels sum to less than 8 (e.g. a single level-5 resonator, sum 5) reported its fractional level as exactly 1.0. The sidebar tooltip then treated that integer value as sitting on a level boundary and showed "8 resonator level(s) needed for next portal level" instead of the correct 11 (next level 2 requires a resonator-level sum of 16).

Remove the clamp from `getPortalLevel()` so it returns the true resonator-based fractional level, and reapply the in-game level-1 minimum in getPortalRange() where it is actually needed for link range. Rewrite the tooltip calculation in terms of the displayed integer level and the resonator sum (8 * (level + 1) - sum), which handles the sub-level-1 case and every other case uniformly.

Fixes #734


core/code/portal_detail_display.js
```diff
@@ -129,11 +129,12 @@ window.renderPortalToSideBar = function (portal) {
...
-    if (levelDetails !== 8) {
-      if (levelDetails === Math.ceil(levelDetails)) levelDetails += '\n8';
-      else levelDetails += '\n' + (Math.ceil(levelDetails) - levelDetails) * 8;
-      levelDetails += ' resonator level(s) needed for next portal level';
+    var portalLevel = window.getPortalLevel(details); // resonator-based fractional level
+    levelDetails = portalLevel;
+    if (levelInt !== 8) {
+      // resonator levels still needed to reach the next displayed portal level
+      var needed = Math.round(8 * (levelInt + 1) - 8 * portalLevel);
+      levelDetails += '\n' + needed + ' resonator level(s) needed for next portal level';
```
McBen: works and is fine.

possible improvement:

```suggestion
      const req_reso_levels = 8 * (levelInt + 1- portalLevel);
      levelDetails += '\n' + req_reso_levels + ' resonator level(s) needed for next portal level';
```
- replaced comment by better var. name
- remove "Round" (value will never have decimal places, because portalLevel is n/8). at first the "round" made me suspicious 'cause it suggest the value can go up or down.

core/code/portal_info.js
```diff
@@ -19,7 +21,7 @@ window.getPortalLevel = function (d) {
     lvl += parseInt(reso.level);
     hasReso = true;
   });
-  return hasReso ? Math.max(1, lvl / 8) : 0;
+  return hasReso ? lvl / 8 : 0;
 };
```
McBen: again, while its fine, 


```suggestion
 * Calculates the resonator-based level of a portal.
 * This includes a decimal part and is not clamped to the minimum level of 1
 *
 * @function getPortalLevel
 * @param {Object} d - The portal detail object containing resonator information.
 * @returns {number} The calculated portal level.
 */
window.getPortalLevel = function (d) {
    if (!d.resonators) return 0;
    return d.resonators.reduce( (sum,reso) => sum+reso.level,0) / 8;
};
```

- shorten comment
- modern, shorten and faster function code
- the if clause shouldnt be required anyway

core/code/portal_info.js
```diff
@@ -87,7 +89,7 @@ window.getPortalRange = function (d) {
   // formula by the great gals and guys at
   // http://decodeingress.me/2012/11/18/ingress-portal-levels-and-link-range/
   var range = {
-    base: window.teamStringToId(d.team) === window.TEAM_MAC ? window.LINK_RANGE_MAC[d.level + 1] : 160 * Math.pow(window.getPortalLevel(d), 4),
+    base: window.teamStringToId(d.team) === window.TEAM_MAC ? window.LINK_RANGE_MAC[d.level + 1] : 160 * Math.pow(Math.max(1, window.getPortalLevel(d)), 4),
```
McBen: I wonder if this is required. 
Noone ever can test if the link range of a single lvl 1 reso portal is 40cm. :)

McBen (approved):

jimsug:
All good notes, thank you! Was overly paranoid with some of the math I think 😂

modos189:
Please remove package-lock.json

core/code/portal_info.js
```diff
@@ -5,21 +5,16 @@
...
-  var lvl = 0;
-  var hasReso = false;
-  $.each(d.resonators, function (ind, reso) {
-    if (!reso) return true;
-    lvl += parseInt(reso.level);
-    hasReso = true;
-  });
-  return hasReso ? Math.max(1, lvl / 8) : 0;
+  if (!d.resonators) return 0;
+  return d.resonators.reduce((sum, reso) => sum + reso.level, 0) / 8;
```
modos189:
```suggestion
  return d.resonators.reduce((sum, reso) => sum + (parseInt(reso?.level) || 0), 0) / 8;
```
the current code is fine, but just to be safe, add this check

PR #922, opened 2026-06-14
keep dialog inside main window

McBen:
a) disable map scroll when dragging a dialog to the borders
b) disable moving dialogs outside the main window

I'm not sure what will happen with large, fixed size dialogs on small screens.
Even if dragging isn't possible on mobile, it could still have a negative impact in similar situations.

### DIFF
--- a/core/code/dialog.js
+++ b/core/code/dialog.js
@@ -242,8 +242,10 @@ window.dialog = function (options) {
     // ui-modal includes overrides for modal dialogs
     dialog.parent().addClass('ui-modal');
   } else {
-    // Enable snapping
-    dialog.dialog().parents('.ui-dialog').draggable('option', 'snap', true);
+    const baseElement = dialog.dialog().parents('.ui-dialog');
+    baseElement.draggable('option', 'snap', true); // Enable snapping
+    baseElement.draggable('option', 'scroll', false); // Disable map scroll while dragging
+    baseElement.draggable('option', 'containment', 'window'); // prevent dragging outside of the window
   }
 
   // Run it
--- a/core/code/dialog.js
+++ b/core/code/dialog.js
@@ -242,10 +242,10 @@ window.dialog = function (options) {
     // ui-modal includes overrides for modal dialogs
     dialog.parent().addClass('ui-modal');
   } else {
-    const baseElement = dialog.dialog().parents('.ui-dialog');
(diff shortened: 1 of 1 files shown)


xscreach:
Guys, can also this problem be addressed? 
It might be out of scope, but imho is kinda related... 🤷

When I open popup in landscape on mobile and then flip phone to the portrait, the popup end up half way off the screen..

McBen:
> Guys, can also this problem be addressed? It might be out of scope, but imho is kinda related... 🤷
> 
> When I open popup in landscape on mobile and then flip phone to the portrait, the popup end up half way off the screen..
> 

Dialogs positions are absolute and are set on creation (or when dragging).
There is no trivial solution for repositioning these.

Example:
```
dialog({...., position: "center});
=> $ <div class="ui-dialog" style="position: absolute; top: 200px; left: 800px;">
```

it might by possible to trigger a collision check but RN I've no idea how.

McBen:
AI solution: (looks okay, but untested)

````
function keepDialogsInViewport() {
    $('.ui-dialog').each(function () {
        const $dlg = $(this);

        const dlgWidth = $dlg.outerWidth();
        const dlgHeight = $dlg.outerHeight();

        const winWidth = $(window).width();
        const winHeight = $(window).height();

        let left = parseInt($dlg.css('left'), 10) || 0;
        let top = parseInt($dlg.css('top'), 10) || 0;

        // Prevent dialog from extending beyond right/bottom edge
        left = Math.min(left, winWidth - dlgWidth);
        top = Math.min(top, winHeight - dlgHeight);

        // Prevent dialog from extending beyond left/top edge
        left = Math.max(0, left);
        top = Math.max(0, top);

        $dlg.css({
            left: left + 'px',
            top: top + 'px'
        });
    });
}

$(window).on('orientationchange', function () {
    clearTimeout(resizeTimer);
    resizeTimer = setTimeout(keepDialogsInViewport, 100);
});
````

modos189 (approved):

nexushoratio:
I regularly drag windows (including dialogs) half off-screen rather than rolling them up or resizing them.  Sometimes I may not even let up on the mouse but put them right back.

I've been doing this for nearly 40 years of using windowing systems.

I can appreciate fixing the map scrolling issue.  But no longer being able to move the dialog partially off screen is very annoying.

PR #923, opened 2026-06-15
JQuery & JQuery UI updates

McBen:
nothing special to see here. just lib updates - no visible changes.

/edit
added a jquery-tooltip workaround
(tooltips are added to the DOM but never removed)


McBen:
added workaround for a "special behaviour" of JQueryUI (see https://bugs.jqueryui.com/ticket/10689/ (2014) )

example of the "feature" I try to workaround:
grafik
and more with every button touch

modos189 (approved):

PR #927, opened 2026-06-25
Update how `.ui-dialog`'s `max-width` is set.

### LINKED ISSUE #504: Problem with "Portal list" dialog
SNAG-2021-05-12_18 20 13
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/ab3b4ab561898f2d4f5af64a31daeeeb1f04ce85/core/style.css#L1040

When using on a computer that has a monitor/window with that is bigger than 700px the dialog-box is restricted in width so that you can not see all info (see included image)

nexushoratio:
This removes the `@media` variant and changes the default to be 1-rem narrower than the viewport.

Closes #504.

### DIFF
--- a/core/style.css
+++ b/core/style.css
@@ -1171,13 +1171,7 @@ h3.title {
 }
 
 .ui-dialog {
-  max-width: calc(100vw - 2px);
-}
-
-@media (min-width: 700px) {
-  .ui-dialog {
-    max-width: 600px;
-  }
+  max-width: calc(100vw - 1rem);
 }
 
 .ui-dialog-content-hidden {


modos189:
Why `- 1rem`? Previously, `- 2px` was used to subtract the border width

nexushoratio:
`2px` will allow the dialog to go up to the edge of the viewport.  `1rem` will leave a bit of a gutter.

On devices with narrow screens (e.g., Firefox on Android), the gutters allow seeing and interacting with the background.  In particular, I can still scroll around on the map without needing to rollup the dialog.

modos189 (approved):

Issue #88, opened 2019-02-05
Hi-res icons


johnd0e:
Some icons are extremely low-res and need to be enhanced/replaced.

Most important:
- [ ] Bookmark star from this sprite: `bookmarks-img.png`
- [ ] `icon-bookmark.png`, `icon-bookmark-map.png`
- [ ] `marker-star.png`
- [ ] `current-location.png`

Also:
- [ ] `mission-length.png`, `mission-type-hidden.png`, `mission-type-random.png`, `mission-type-sequential.png`, `mission-type-unknown.png`
- [x] `open-folder-icon_sml.png` (add-kml)
  alt: https://github.com/makinacorpus/Leaflet.FileLayer/blob/master/docs/folder.svg

Externals:
- [x] `minimap-toggle.png`
- [ ] `pan-down.png`, `pan-left.png`, `pan-right.png`, `pan-up.png`
- [ ] `zoom-slider.png`

PR #934, opened 2026-07-05
Proof-of-concept of using Material Icons or Material Symbols.

nexushoratio:
This is not expected to be submitted, but rather a demonstration and playground.

It demonstrates how icons could be replaced using either the older Material Icons (considered deprecated) or the newer Material Symbols.

In this example, I updated three icons in the portal details: core map pin, close, and bookmarks plugin's star.  In all cases, I kept the current icon and added multiple examples of replacements using both sets.

I researched a few of the existing issues on this topic (last discussed in 2019/2020).  In this case, I am using the fonts rather than svg equivalents (as currently used for the map pin).

I did ZERO testing with the Android app.  The "close" icons look a little funky on Firefox on Android, but I am probably the only user of that configuration.

Based upon the previous discussions:

* Icons are available 'out-of-the-box'
* Fairly easy to use (just `<span class='material-icons'>...</span>`), though specifying a filled variant can be cumbersome with Symbols.
* Symbols does use excessive resources [See optimization docs](https://developers.google.com/fonts/docs/material_symbols#optimize_the_icon_font), but they seem cached on page reloads.  As currently specified, Icons weigh in at 130kB and Symbols at 458kB.

Specific details about this PoC:

* Renamed the old `material-icons` class (only kept so that existing map icon could be shown against new ones).
* This uses the live web version, which avoids the need to keep local assets. Related, as in Safari, Firefox makes it trivial for users to block webfonts (I actually do this), but I think that can be considered self-inflicted pain.
* Did not even consider Font Awesome, unsure if it needs JS support or not.
* The change from `<title>...</title>` to an attr `title:` is not necessary, but made for consistency in the demo.
* Heavily tweaked `h3.title` to deal with the excessive number of icons.

Personal thoughts:

* I think the old map pin (Icons) looks better, but as demonstrated with the bookmark stars, Symbols has more variations.
* The function `setupIcons()` can probably go away.
* If Symbols are used, it might be possible to reduce the download size.  A test for just the icons used in this demo is 2.8kB, but the names must be alphabetical.  That means each plugin would have to register what they want, then an appropriate `<style>` header created.  Feels fragile.

Related: #45, #88, #236, #312, #414

### DIFF
--- a/core/code/boot.js
+++ b/core/code/boot.js
@@ -270,6 +270,33 @@ function updateControlBarZIndex() {
   });
 }
 
+function setupWebComponents() {
+
+  class MaterialSymbolIcon extends HTMLElement {
+    // This could be used to implement a fallback if necessary.
+    static #ready = false;
+
+    connectedCallback() {
+      console.log('ready', this.constructor.#ready);
+      this.classList.add('material-symbols-outlined', 'mat-resolved');
+    }
+
+    static {
+      document.fonts.ready.then((fontSet) => {
+        fontSet.forEach((font) => {
+          if (font.family === '"Material Symbols Outlined"') {
+            font.loaded.then(() => {
+              MaterialSymbolIcon.#ready = true;
+            });
+          }
+        });
+      });
+    }
+
+  }
+
+  customElements.define('mat-icon', MaterialSymbolIcon);
+}
 
 /**
  * The main boot function that initializes IITC. It is responsible for setting up the map,
@@ -287,6 +314,7 @@ function boot() {
   var loadPlugins = prepPluginsToLoad();
   loadPlugins('boot');
 
+  setupWebComponents();
   window.setupDialogs();
   checkingIntelURL();
   setupIngressMarkers();
--- a/core/code/portal_detail_display.js
+++ b/core/code/portal_detail_display.js
@@ -147,28 +147,80 @@ window.renderPortalToSideBar = function (portal) {
       $('<h3>', { class: 'title' })
         .text(title)
         .prepend(
-          $('<svg><use xlink:href="#ic_place_24px"/><title>Click to move to portal</title></svg>')
+          $('<svg><use xlink:href="#ic_place_24px"/></svg>')
             .attr({
-              class: 'material-icons icon-button',
+              class: 'material-icons0 icon-button',
               style: 'float: left',
+              title: 'Click to move to portal',
             })
             .click(function () {
               window.zoomToAndShowPortal(guid, [details.latE6 / 1e6, details.lngE6 / 1e6]);
               if (window.isSmartphone()) {
                 window.show('map');
               }
-            })
+            }),
+          $('<span>')
+            .text('place')
+            .attr({
+              class: 'material-icons icon-button',
+              style: 'float: left',
+              title: 'old style material icon',
+            }),
+          $('<span>')
+            .text('location_on')
+            .attr({
+              class: 'material-symbols-outlined icon-button',
+              style: 'float: left; font-variation-settings: "FILL" 1;',
+              title: 'new style material symbol (filled)',
+            }),
+          $('<span>')
+            .text('location_on')
+            .attr({
+              class: 'material-symbols-outlined icon-button',
+              style: 'float: left; font-variation-settings: "FILL" 0;',
+              title: 'new style material symbol (open)',
+            }),
         ),
 
       $('<span>')
-        .attr({
+        .append(
+          $('<span>')
+            .text('close_small')
+            .attr({
+              class: 'material-symbols-outlined',
+              style: 'font-size: revert;',
+              title: 'new style material symbol (small)',
+            }),
+          $('<span>')
+            .text('close')
+            .attr({
+              class: 'material-symbols-outlined',
+              style: 'font-size: revert;',
+              title: 'new style material symbol',
+            }),
+          $('<span>')
+            .text('close')
+            .attr({
+              class: 'material-icons',
+              style: 'font-size: revert;',
+              title: 'old style material icon',
+            }),
+          $('<span>')
+            .text('x')
+            .attr({
+              title: 'small x',
+            }),
+          $('<span>')
+            .text('X')
+            .attr({
+              title: 'Close [w]',
+              accesskey: 'w',
+            })
+            .click(function () {
+              window.renderPortalDetails(null);
+            })
+        ).attr({
           class: 'close',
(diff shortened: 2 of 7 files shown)


nexushoratio:
A couple of screen caps:

normal:
normal

zoomed:
zoomed

modos189:
Let's try adding "material symbols" - a complete set of icons (since we don't know which icons third-party plugin developers might want to use) - but with a fixed configuration (thickness, etc.).

How large the file will be?

It’s also very important to ensure that, for now (and if) the font hasn’t loaded yet, no words or tofu characters are displayed. Perhaps the `documents.font.load` API or CSS could help with this.

We already have a plugin that adds Font Awesome icons. But the Material Symbols set is larger and isn't limited to a small selection of free icons like FA. That's why I think we should go with Material Symbols.
Although it's certainly hard to choose, both have their pros and cons. Material symbols file is larger

nexushoratio:
> Let's try adding "material symbols" - a complete set of icons (since we don't know which icons third-party plugin developers might want to use) - but with a fixed configuration (thickness, etc.).
> 
> How large the file will be?

The woff2 file is 447kB.

I have not done any analysis of what might be reduced due to retiring other assets.

On my machine, the resulting plugin is currently 1.4M.  So roughly a 33% increase in size, which could not be cached separately.

Any reason why the preference would be to embed them rather than referencing them like is already done for other fonts?
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/cfbfd6cc70edc3480c7e1aed24464157270bd7a1/core/total-conversion-build.js#L196

(Obviously, I know icons are different from fonts, as browsers will have local fonts they can use.)

> It’s also very important to ensure that, for now (and if) the font hasn’t loaded yet, no words or tofu characters are displayed. Perhaps the `documents.font.load` API or CSS could help with this.

They definitely display words rather than fonts.  Since I disable most external fonts myself, this is what it looks like for me to view https://fonts.google.com/icons 

image

If I allow them:
image

In the above, that happens to be a font called "Google Symbols Subset Subset" which I think it unique for that website.

For this PoC, if not loaded, it would look like this:
image

loaded:
image

The Google site is using CSS to truncate the text.
 
> We already have a plugin that adds Font Awesome icons. But the Material Symbols set is larger and isn't limited to a small selection of free icons like FA. That's why I think we should go with Material Symbols. Although it's certainly hard to choose, both have their pros and cons. Material symbols file is larger

When looking at #312, I thought there wasn't one.  But I now see there is a community plugin.

As previously argued in the referenced discussions, I also think that some solution should be provided by core rather than as a plugin, whatever the solution might be.

I originally thought about MS because I came across this reference first:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/cfbfd6cc70edc3480c7e1aed24464157270bd7a1/core/code/sidebar.js#L49-L53

I'd be happy to explore other options if folks want.

nexushoratio:
Additional info.

It looks like the equivalent raw svg files are 3.8M.  (Unsurprising since apparently woff files have built in compression.)

Making them into an SVG spritesheet like what `setupIcons()` does, would take 3.6M.

nexushoratio:
I updated this example a little bit:

* Removed the numerous example icons in favor of only the *Material Symbols* ones
  * Also expanded the scope to include bookmark's portal-list integration (since both use *star*)
* Implemented a [Web Component](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) to simplify usage
  * Handles using plain/fill variations of icons
  * Capitalizes the first character of the word if font is not loaded (e.g., would say "Close" rather than "close")
  * Demonstrates how a check for the font being loaded could be hooked in (does nothing more then `console.log()`)

---

modos189:
I previously suggested hiding icons if the font fails to load. But I overlooked that most of the icons we already use aren't optional decoration next to text - they're the entire content of the control (e.g. the close button, the portal location button).

Regarding what we bundle vs. what we load externally: if a regular text font fails to load, the browser just falls back to another font - not a real problem. But if an icon font fails to load, controls like the close button or "locate portal" button effectively break. A text fallback (e.g. showing the ligature name instead of a glyph) looks just as bad in that case.

At the same time, bundling the full icon font (~400KB+) just to guarantee a handful of essential glyphs always render doesn't seem like a good tradeoff either.

So I'd propose splitting this into two stages:

1. **Stage 1:** bundle a small subset of Material Symbols covering only the icons we need to replace the existing SVGs and text-based icons currently in use. This subset stays small and ships inside the bundle, so these essential icons are always guaranteed to render, with no dependency on external loading.
2. **Stage 2 (future):** add a "show optional icons" setting (for example, plugin icons). When enabled, the full icon pack (covering icons contributed by third-party plugins) is automatically loaded from an external source. Since this covers purely decorative/optional icons, a failed load there is not a problem - it can degrade gracefully.

For stage 1, we can generate the subset using the Google Fonts API's `icon_names` parameter, which builds a font containing only the requested glyphs: https://developers.google.com/fonts/docs/material_symbols#optimize_the_icon_font. By selecting only the icons actually used in IITC, with a fixed style and weight, this can bring the font size down to a few kilobytes or even less.

nexushoratio:
Let me do some testing.

I had tried multiple loading of the same font (on the idea that each plugin could have a its own `icon_names`, but that did not appear to work.  Only one worked.

The latest version of the code renders as this if I block that particular font:
Screen Shot 2026-07-11 at 10 23 24

---

FWIW, IITC already uses some unicode characters directly that depend on installed fonts and could display tofu characters, though maybe not uniquely.

For example, resonator location are specified with both ASCII letters and Unicode arrows:
N ↑
NW ↖

Hah!  In other locations in portal_details_display, `×` is actually used instead of `X`!

The sidebar also uses:
location…
code…
${start}…${end}

style.css has:
▲ ⊞ ⊟

Other core code:
©

plugins/distance-to-portal uses:
°

For bookmarks we could use:
☆★

Maybe it makes more sense to instead go for plain Unicode characters first, and see what is left, rather than some icon font set?

modos189:
Unicode characters may display differently on various platforms, and not all icons are available there (such as the search and geolocation icons). That is why SVG icons are currently used there.

> Unicode characters may display differently on various platforms, and not all icons are available there (such as the search and geolocation icons). That is why SVG icons are currently used there.

If we include a small set of icons with IITC and load the full set separately, I think it will work. But we need to check it out.

nexushoratio:
🔍

nexushoratio:
Closing now that an official commit has been made.

Huzzah!

PR #937, opened 2026-07-09
Simplify how the side panel's portal title is constructed.

nexushoratio:
The `h3` is now the whole title line (a `div` would be better, but at least bookmarks uses `h3.title` in a query), with immediate children all inline items (e.g., `svg` and `span`).  And an explicit id of `portaltitle`.

The `h3` is also changed to being a flex container to allow easier control of elements.  Rather than having to guess on what the right margin of the title string should be (e.g., `right-margin: 17px`), by using flex, the browser takes care of it automatically.

This also removes requiring `float: <side>` or absolute positioning.  They now simply show up in the order they are added.

### DIFF
--- a/core/code/portal_detail_display.js
+++ b/core/code/portal_detail_display.js
@@ -144,32 +144,36 @@ window.renderPortalToSideBar = function (portal) {
     .html('') // to ensure it's clear
     .attr('class', window.TEAM_TO_CSS[window.teamStringToId(details.team)])
     .append(
-      $('<h3>', { class: 'title' })
-        .text(title)
-        .prepend(
+      // h3.title is kept for supporting plugins that reference it.
+      $('<h3>')
+        .attr({
+          id: 'portaltitle',
+          class: 'title',
+        })
+        .append(
           $('<svg><use xlink:href="#ic_place_24px"/><title>Click to move to portal</title></svg>')
             .attr({
               class: 'material-icons icon-button',
-              style: 'float: left',
             })
             .click(function () {
               window.zoomToAndShowPortal(guid, [details.latE6 / 1e6, details.lngE6 / 1e6]);
               if (window.isSmartphone()) {
                 window.show('map');
               }
+            }),
+          $('<span>', { class: 'value' })
+            .text(title),
+          $('<span>')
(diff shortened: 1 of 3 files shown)


modos189:
Two small things:
- `sidebar.js:161` now grabs the close button "X" into the image dialog title, since `.close` is a child of `h3.title`. Please read only the title text, e.g. `querySelector('.title .value').innerText`.
- `#portaltitle .close`: I suggest to drop `margin-top: -2px;` so the "X" stays vertically centered in the flex row.

nexushoratio:
> Two small things:
> 
>     * `sidebar.js:161` now grabs the close button "X" into the image dialog title, since `.close` is a child of `h3.title`. Please read only the title text, e.g. `querySelector('.title .value').innerText`.

And this is why we do code reviews!  :->

Would it also make sense to change it from:
```
var title = e.delegateTarget.querySelector('.title .value');
```
to:
```
const title = document.querySelector('#portaltitle .value');
```
?

---

> 
>     * `#portaltitle .close`: I suggest to drop `margin-top: -2px;` so the "X" stays vertically centered in the flex row.

I kept that because that was what was happening before.

Original:
image

This version:
Screen Shot 2026-07-10 at 07 02 44

Your proposal:
Screen Shot 2026-07-10 at 07 05 15

Another option could be change it to a small `x` and raise it even more:
Screen Shot 2026-07-10 at 07 07 30

modos189:
Yes, let's go with `const title = document.querySelector('#portaltitle .value').innerText;`

> Your proposal:

I think it looks better this way because it's centered

DanielOndiordna:
Please add a ? after a querySelector, to prevent errors when not found and a default result:
`const title = document.querySelector('#portaltitle .value')?.innerText || '';`

nexushoratio:
I just realized that I misspelled "vertical" as "veritcal".  Can that be squashed out if this is merged?  (embarrassing!)

nexushoratio:
In another discussion I learned that there is a popular Unicode close character used:  `×`

This is what this would look like if that was used instead of `X`:
Screen Shot 2026-07-11 at 12 12 12

PR #965, opened 2026-09-11
Draw the interface icons from a bundled Material Symbols font

modos189:
Why a font rather than an SVG sprite or `mask-image`: 13 kB in the script against 60 and 84 kB for the same icons, and ligature names match Google's documentation instead of prefixed class names. Either way a generator is needed; the price is the fallback below
- `icons_gen.py` fetches the subset named by `core/icons/icons.txt` (91 icons) as woff2, inlined by `icons.css`; the mobile app gets the whole family as truetype, so its icons do not depend on the list
- Icons are written as ligatures in an `<ms-icon>` element, `<ms-icon label="Remove" fallback="X">close</ms-icon>`, with a `filled` attribute driving the font's `FILL` axis; a `label` becomes `role="img"` with `aria-label`, an unlabelled icon is hidden from assistive technology
- Search box, portal title, chat tab close, both accordions, the sidebar and chat toggles and the Leaflet layers control now draw from the font, replacing three inlined SVGs, an SVG sprite, the unicode glyphs and the CSS triangles
- `setupIconFont` measures the ligature rather than asking `document.fonts`, which reports the face as loaded even when the browser is told to ignore page fonts; `body.icons-unavailable` then shows each icon's `fallback` character or its `label`, hides a decorative one, and gives the layers control back its `leaflet.css` background
- `IITC.toolbox.addButton` and the mobile `addPane` take a ligature name, telling it apart from a Font Awesome class by the `fa-` prefix, so plugins outside this repository keep working
- `mpe.setMultiProjects` gains an `icon` field alongside `fa`, preferred when both are given
- Bookmarks, Keys and MPE draw their own icons from the font; the bookmark star's two sprite positions became the two values of the `FILL` axis
- Mobile panes render the ligature through `IITC_IconDrawable`, the legacy `ic_action_*` names mapped onto icons and 36 bitmaps dropped; an unknown name falls back to `extension`

desktop
mobile

close https://github.com/IITC-CE/ingress-intel-total-conversion/pull/45 https://github.com/IITC-CE/ingress-intel-total-conversion/pull/934 https://github.com/IITC-CE/ingress-intel-total-conversion/issues/236 https://github.com/IITC-CE/ingress-intel-total-conversion/issues/218


nexushoratio:
Show off.  :->

nexushoratio:
Rather devs having to remember:
```html
<i class="icon" aria-label="Remove" data-fallback="X">close</i>
```

Would the Web  Component idea I suggested be appropriate?

Say:
```html
<iitc-icon label="Remove">close</iitc-icon>
```

or

```html
<iitc-icon fallback="Y">close</iitc-icon>
```

Where the implementation could include default mapping of ligature to aria-label and fallback, but could be overridden in places it makes sense.  I suppose it could even automatically have Font Awesome mappings as well giving users of that Community Plugin consistency.

It feels to me that using Web Components is a "more natural" solution than forcing use of other markup that already has well known semantics.

---

I noticed that one `data-fallback` uses 📍.  But isn't that also dependent upon the fonts the user has installed and allowed to be used?  It seems like like the fallbacks should only allow ASCII.

---

Also, thanks for running with this.  Life happened and I have not had a chance to get back to it.

modos189:
> I noticed that one data-fallback uses 📍. But isn't that also dependent upon the fonts the user has installed and allowed to be used? It seems like like the fallbacks should only allow ASCII.

Not just this icon, but also the search and geolocation icons.
On my system (macOS), these emojis display correctly. Only the color and appearance of the icons differ between systems, as we discussed in your PR.

> Would the Web Component idea I suggested be appropriate?

Yeah, that's a good idea

core/code/icons.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @file Defines <ms-icon>, an icon of the bundled Material Symbols font named by its ligature
```
nexushoratio: Is `ms-icons` a good name here?  I tend to associate `ms` with Microsoft.

nexushoratio (commented):
Nice.

core/code/icons.js
modos189: Haha, yeah, I get that 😄. On the other hand, if we (or someelse) ever create web components for other icon sets (for example, Font Awesome), they'll be easy to distinguish by name. Besides, we can't use "<icon>" anyway, since custom element names are required to contain a hyphen

nexushoratio: That was part of why I used `mat-` in my proposal.  (That and I wasn't sure if Symbols or Icons would end up being better).  But also `iitc-` in my comment here.

And at that many characters, it starts to lose the charm of being "easy to use".  Still, if `iitc-` was used, it could support both Material Symbols and Font Awesome in the same implementation.

Essentially that is already happening, only it is named "fallback".

nexushoratio:
Can #88 and #236 be closed now?

PR #966, opened 2026-09-11
Handle Antimeridian

### LINKED ISSUE #341: Issues with crossing Date Line (180th meridian)
https://simple.wikipedia.org/wiki/International_Date_Line
https://en.m.wikipedia.org/wiki/180th_meridian

In all ingress data longitude falls in range [`-180`,`179.99999`].
Crossing Date Line during panning is not a problem itself, because we have code that wraps longitude when map center is beyond this range.

Issues arise when we have in view both sides in the same time: map must be continuous, so one side is always out of range.

Currently IITC queries intel data only for one side. It's possible to query opposite side too, we only need wrap it bounds to be in supported range. Unfortunately it's not enough, as returned data will be out map bounds.

McBen:
- [x] request all tiles in view
- [x] render entities wrapped to current view
- [x] move all entities on worldjump (antimeridian view switch)
- [x] fix GeodesicPolyine

Note: it won't duplicate entities. So if multiple antimeridians are visible the other world views stays empty.

Note2: there is still map "flashing" because if leaflets basemap handling

fix https://github.com/IITC-CE/ingress-intel-total-conversion/issues/341


McBen:
looking good.
some plugins might need to be tweaked to handle it too.

what it does / how it work:
in case of the Antimeridian is rendered leaflet simply extends the coordiantes.
so the map will show (for example) a -200 to -160 or a 160 to 200 Longitude window.

Details:
a) request all tiles in view
previously boundary where "clamped" to the correct +-180 boandaries.
This is removed. And in TileID generation the correct warpping is added.

b) render entities wrapped to current view
Before rendering a tile it is checked if it would be out of current view.
In this case a +-360° offset is added to all portals,links,fields Longitude.

c) move all entities on worldjump
when the user moves the view across the antimeridian the world jumps
from -180 to 180 (or vice versa)
Here we have to move all entities too.

d) fix GeodesicPolyine
GeodesicPolyine made some tweaks to trick the wrapping.
But it fails in one direction.So we simple switch direction while drawing it.
Deeper investigation are required here.


Known Bug:
On a World view with a >360° Entites are only drawn once and so might 
visible only on the left or right side.


e) Plugin: Scoring Region
- added correct wrapping
- replaced Cube-Face drawing (red lines)

McBen:
oh, there is an issue when switching antimeridian without crossing it (moving the whole way around the earth).
I guess it is related to c) while tiles are still in cache.

modos189:
Thanks! Tested this around Fiji and pushed ten commits to your branch.

Broken in practice

- Portals near 180° disappeared - placeholders inherited the world copy of whichever tile named them (d22a65ee)
- Links crossing 180° drawn as a ~354° span, and never cleared afterwards (2a4ee7cb)
- Any setView further than half the globe shifted every entity by 360° (668dce60, test in 5fed8b47)
- Latitude left unclamped in refresh() - rows like 3_22_-2_8_8_100, status stuck on "errors" (0a399f34)
- mapMoveEnd still clamped, so a view entirely beyond 180° skipped its refresh (8c59d036)

Minor

- Tile offset taken from the live view instead of the one it was fetched for (781f7701)
- No lngE6_delta default for fields and links - NaN for anything a plugin injects (f58763e4)
- Dropped the geodesic reverse workaround; it mutated the array getLatLngs() returns (14dca882)
- World copy contract written down in the JSDoc (5916b0b0)

McBen:
will review you commits soon.

what I see is: 
- https://github.com/IITC-CE/ingress-intel-total-conversion/commit/668dce60eb394de8bf95df0fe62bfa8258045bdc
doesn't change anything (replaces Math.Round by "if"'s)

and the link/field is missing now again:
Bildschirmfoto_20260916_122511
Bildschirmfoto_20260916_122724

McBen:
[781f770](https://github.com/IITC-CE/ingress-intel-total-conversion/pull/966/commits/781f77014429a43b587632d4daa20793d97d3e7c)

render should render entities regardless of the requested boundaries..
you cannot request the same  tile twice when just the boundary was changed.

modos189:
>  doesn't change anything (replaces Math.Round by "if"'s)

Right, during a drag it behaves the same. The difference is only for moves without dragging (search, permalink, setView) where the centre jumps by more than 180°: Leaflet doesn't wrap the map there, but the old check still shifted every entity by 360 degrees. On a wide view this pushed already-rendered entities near the target off the map until the next render pass

McBen:
> Right, during a drag it behaves the same. 

so this is the only relevant line here
> if (!window.map.dragging || !window.map.dragging.moving()) return;

I think we have to change that moving thing anyway.
it should check if anitmeridian is visible and reposition all data instead of just check for a crossing.

modos189:
>  it should check if anitmeridian is visible and reposition all data instead of just check for a crossing.

Agreed, and it's done. Every entity is now drawn on the world copy nearest an anchor longitude, calculated from the true longitude. The anchor follows the map centre once they are more than 45° apart. That works for drag wraps, setView and flyTo alike, so dragging.moving() and the tile offset are gone

McBen:
in the end it is simpler as I expected.

I started with the "move tile" approach because I assumed the calculation is far more complicated. 
But it turns out to be a quite simple calculation.

